PR #130, opened 2014-12-15
Add HttpContextAccessor

HaoK:
- Add [I]HttpContextAccessor and stop adding IContextAccessor as part of HostingServices
- Unify RequestContainer to use EnsureRequestServices

Review: Is Hosting the right home for IHttpContextAccessor?  Could also live in HttpAbstractions or RequestContainer...

cc @davidfowl @lodejard @divega


HaoK:
- Updated with ThreadLocal for ASPNETCORE50
- Added IHttpContextAccessor to hosting manifest, 
- Removed SetContextSource Funcs and explicitly added the root logic into HttpContextAccessor with IsRootContext property

HaoK:
Merged f786fb7bd0b60c1551774235a4954830760e6c2e

PR #142, opened 2015-01-14
Simplify HttpContextAccessor (now singleton)

HaoK:
MVC tests all pass with this change so looks good I think

cc @davidfowl @lodejard @divega

### DIFF
--- a/src/Microsoft.AspNet.Hosting/HttpContextAccessor.cs
+++ b/src/Microsoft.AspNet.Hosting/HttpContextAccessor.cs
@@ -1,73 +1,12 @@
 // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-#if ASPNET50
-using System.Runtime.Remoting.Messaging;
-using System.Runtime.Remoting;
-#elif ASPNETCORE50
-using System.Threading;
-#endif
 using Microsoft.AspNet.Http;
 
 namespace Microsoft.AspNet.Hosting
 {
     public class HttpContextAccessor : IHttpContextAccessor
     {
-        private HttpContext _value;
-
-        public bool IsRootContext { get; set; }
-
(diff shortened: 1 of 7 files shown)


HaoK:
@davidfowl Restored the callcontext/asynclocal logic

src/Microsoft.AspNet.Hosting/HttpContextAccessor.cs
```diff
@@ -14,60 +14,39 @@ namespace Microsoft.AspNet.Hosting
...
-            CallContext.LogicalSetData(LogicalDataKey, new ObjectHandle(httpContext));
-            return prior != null ? prior.Unwrap() as HttpContext : null;
-#elif ASPNETCORE50
-            var prior = _httpContextCurrent.Value;
-            _httpContextCurrent.Value = httpContext;
-            return prior;
 #else
-            return null;
+        // REVIEW: or should this throw?
+        public HttpContext Value { get; set; }
```
davidfowl: What is this? Nuke it, there is no else

test/Microsoft.AspNet.TestHost.Tests/TestServerTests.cs
```diff
@@ -38,6 +38,23 @@ public void ThrowsIfNoApplicationEnvironmentIsRegisteredWithTheProvider()
...
 
         [Fact]
+        public async Task CanAccessHttpContext()
+        {
+            var services = new ServiceCollection().BuildServiceProvider();
+            TestServer server = TestServer.Create(app =>
+            {
+                app.Run(context =>
+                {
+                    var accessor = app.ApplicationServices.GetRequiredService<IHttpContextAccessor>();
```
davidfowl: A better test would to be to do this to a singleton service and called from here

davidfowl:
After the changes, :shipit: !

HaoK:
Merged 16a811479eb4f098387b0bcd6702e25003ae6c59 still need to update the test

Issue #145, opened 2015-01-21
Improve hosting APIs for startup self host scenarios


davidfowl:
While updating omnisharp to beta2 I ended up doing this https://github.com/OmniSharp/omnisharp-roslyn/blob/beta2/src/OmniSharp/Startup.cs#L38, because we made it close to impossible to add services if you're the host. I suggest we make it so that the caller of HostingServices.Create can provider types that will be imported by the calling application.

/cc @HaoK @lodejard @Tratcher

HaoK:
Yeah I agree I felt the same way when writing unit tests as well.

PR #146, opened 2015-01-22
Make it easier to add hosting services

HaoK:
Fixes https://github.com/aspnet/Hosting/issues/145

### DIFF
--- a/src/Microsoft.AspNet.Hosting/HostingServices.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingServices.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.DependencyInjection;
-using Microsoft.Framework.DependencyInjection.ServiceLookup;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.Runtime;
 using Microsoft.Framework.Runtime.Infrastructure;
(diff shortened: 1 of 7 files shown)


HaoK:
Hosting.Create() basically is the new fallback service provider method with this change...

/cc @davidfowl @lodejard @Tratcher

HaoK:
Question, do we want to allow the additional services to override the fallback ones?  Currently we add the fallback services after the additional, so I'd just have to flip the order we add the services when Importing if we want to enable overriding behavior...

davidfowl:
@HaoK what about making this `Action<IServiceCollection>` ?

src/Microsoft.AspNet.Hosting/HostingServices.cs
```diff
@@ -25,28 +26,34 @@ private static IServiceCollection Import(IServiceProvider fallbackProvider)
...
 
-        public static IServiceCollection Create(IServiceProvider fallbackServices, IConfiguration configuration = null)
+        public static IServiceCollection Create(IServiceProvider fallbackServices, IConfiguration configuration = null, IServiceCollection additionalHostServices = null)
         {
             configuration = configuration ?? new Configuration();
-            var services = Import(fallbackServices);
+            var services = Import(fallbackServices, additionalHostServices);
             services.AddHosting(configuration);
-            services.AddSingleton<IServiceManifest>(sp => new HostingManifest(fallbackServices));
+            services.AddSingleton<IServiceManifest>(sp => new HostingManifest(fallbackServices, additionalHostServices));
```
HaoK: @davidfowl The code here, is passing the collection to the manifest

The complication with making it an Action, is that we need to be able to figure out what services were added so we add them to the manfiest.  If we take the ServiceCollection itself rather than the action, we know that every service in there should be part of the manivest.  If we simply run the action as part of the service collection we are building, we have no way to determine which services came from hosting (but shouldn't be exported), vs ones they really want added.

HaoK:
Updated with Action

davidfowl:
This makes me wonder if things would be simpler if we just queried for IEnumerable<IServiceManifest> in the Import method instead.

HaoK:
Good call on moving the bookkeeping to Import, makes the code quite compact now.

Also found a few minor issues (missing null guard in ConfigureHostingEnvironment, and fixed an issue with the TestServer unit test using the wrong overload of TestServer.Create

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -47,14 +47,14 @@ public TestServer(IConfiguration config, IServiceProvider serviceProvider, Actio
...
         {
-            return Create(CallContextServiceLocator.Locator.ServiceProvider, app);
+            return Create(CallContextServiceLocator.Locator.ServiceProvider, app, configureHostServices);
         }
 
-        public static TestServer Create(IServiceProvider serviceProvider, Action<IApplicationBuilder> app)
+        public static TestServer Create(IServiceProvider serviceProvider, Action<IApplicationBuilder> app, Action<IServiceCollection> configureHostServices = null)
         {
-            var appServices = HostingServices.Create(serviceProvider).BuildServiceProvider();
+            var appServices = HostingServices.Create(serviceProvider, /*config*/ null, configureHostServices).BuildServiceProvider();
```
davidfowl: Use named arguments when passing null.

HaoK:
Updated PR removing optional parameters, and moving IConfiguration last for Create

HaoK:
Reverted since 3 MVC functionals failed from this, need to investigate

davidfowl:
:shipit: When @HaoK figures out the bug

davidfowl:
@HaoK I want one more overload:

```C#
HostingServices.Create(Action<IServiceCollection> configureHostServices)
```

davidfowl:
We should also look at cleaning up the API a bit. What purpose does the returned IServiceCollection serve if you can't add anything to it.

HaoK:
Well, we use it to create the service collection, so we pass it around like in StartupLoader, and it seems nicer to let Create new up the Collection than require all callers to pass in the service collection they want to import into.  The first iteration of this looked more like that, where it was Import instead of Create and took the collection as a parameter.

HaoK:
Overload added 28cc37de96d8593e81ba5666db546ede764196f5

PR #147, opened 2015-01-28
Register ILogger<T> -> Logger<T> in default services

DamianEdwards:
See https://github.com/aspnet/Logging/pull/96

### DIFF
--- a/src/Microsoft.AspNet.Hosting/HostingServicesCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingServicesCollectionExtensions.cs
@@ -17,6 +17,7 @@ namespace Microsoft.Framework.DependencyInjection
         {
             var describe = new ServiceDescriber(config);
             services.TryAdd(describe.Singleton<ILoggerFactory, LoggerFactory>());
+            services.TryAdd(describe.Singleton(typeof(ILogger<>), typeof(Logger<>)));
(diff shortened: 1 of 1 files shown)


davidfowl:
Omg this extension method is in Hosting?

DamianEdwards:
@davidfowl yeah I had that discussion with @loudej, it's there 'cos the ANIs are conditional on it targeting ASPNET50, blah blah.

davidfowl:
:shipit:

src/Microsoft.AspNet.Hosting/HostingServicesCollectionExtensions.cs
```diff
@@ -17,6 +17,7 @@ public static IServiceCollection AddLogging(this IServiceCollection services, IC
         {
             var describe = new ServiceDescriber(config);
             services.TryAdd(describe.Singleton<ILoggerFactory, LoggerFactory>());
+            services.TryAdd(describe.Singleton(typeof(ILogger<>), typeof(Logger<>)));
```
DamianEdwards: @davidfowl are you sure this should be a singleton? `ILogger` and `Logger` in particular don't appear to be thread safe.

DamianEdwards:
Merged in eba2808109f86b034466051d0abccf2d01cf81fa

Issue #164, opened 2015-02-16
Remove interfaces and classes


muratg:
Not a chain anymore:
- IStartupLoaderProvider
- IStartupManager
- StartupLoaderProvider
- StartupManager

Issue #163, opened 2015-02-13
IServerManager polish


muratg:
- Rename the class to IServerLoader
- Rename the method GetServerFactory to LoadServerFactory

Issue #156, opened 2015-02-12
HostingContext.Servername: misleading name


muratg:
Can be confused with the server name (as in URL)

Praburaj:
Was there a suggestion on name during the discussion? 

A typical value for this looks like `Microsoft.AspNet.Server.WebListener` which is the package containing the server's factory type. Should we rename this to `ServerFactoryAssembly`?

This may also demand a change in the command parameter name from `--server` => `--serverfactory.assembly`

```
"web": "Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.WebListener --server.urls http://localhost:5002"
```

```
"web": "Microsoft.AspNet.Hosting --serverfactory.assembly Microsoft.AspNet.Server.WebListener --server.urls http://localhost:5002"
```

thoughts?

Issue #155, opened 2015-02-12
HostingContext should be thinned


muratg:
IServiceProvider provides IServiceFactory and other types that are currently directly exposed.

glennc:
Either this or https://github.com/aspnet/Hosting/issues/150 should be closed as a dup :)

muratg:
This one is done

PR #167, opened 2015-02-26
Hosting API review feedback # 1

Praburaj:
Addresses:
https://github.com/aspnet/Hosting/issues/164
https://github.com/aspnet/Hosting/issues/163
https://github.com/aspnet/Hosting/issues/155
https://github.com/aspnet/Hosting/issues/156
https://github.com/aspnet/Hosting/issues/154

@loudej for https://github.com/aspnet/Hosting/issues/164


msotclas:
Hi __@Praburaj__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Praburaj Thiagarajan). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
```diff
@@ -9,11 +9,10 @@ namespace Microsoft.AspNet.Hosting.Builder
 {
     public class HttpContextFactory : IHttpContextFactory
     {
-        public HttpContext CreateHttpContext(object serverContext)
+        public HttpContext CreateHttpContext(IFeatureCollection featureCollection)
         {
-            var featureObject = serverContext as IFeatureCollection ?? new FeatureObject(serverContext);
-            var featureCollection = new FeatureCollection(featureObject);
-            var httpContext = new DefaultHttpContext(featureCollection);
+            var featureObject = featureCollection ?? new FeatureObject(null);
```
davidfowl: Just use the passed in feature collection. This should never be null.

davidfowl: Just use the passed in feature collection. This should never be null.

davidfowl: Just use the passed in feature collection. This should never be null.

davidfowl: Just use the passed in feature collection. This should never be null.

davidfowl: Just use the passed in feature collection. This should never be null.

davidfowl: Just use the passed in feature collection. This should never be null.

davidfowl: Github fail

src/Microsoft.AspNet.Hosting/HostingContext.cs
```diff
@@ -11,7 +10,7 @@ namespace Microsoft.AspNet.Hosting
 {
     public class HostingContext
     {
-        public IServiceProvider Services { get; set; }
+        public IApplicationLifetime ApplicationLifeTime { get; set; }
```
Eilon: I think `Lifetime` is one word.

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -105,9 +106,16 @@ private void EnsureApplicationStartup(HostingContext context)
...
+            var diagnosticMessages = new List<string>();
+            context.ApplicationStartup = _startupLoader.LoadStartup(
                 context.ApplicationName,
-                context.EnvironmentName);
+                context.EnvironmentName,
+                diagnosticMessages);
+
+            if (context.ApplicationStartup == null)
+            {
+                throw new Exception(diagnosticMessages.Aggregate("TODO: web application entrypoint not found message", (a, b) => a + "\r\n" + b));
```
Eilon: Use appropriate exception type. This is probably an ArgumentException, with `context` being the invalid argument.

Also, no more "TODO" in the error message - just put whatever the error message should be.

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -94,9 +90,9 @@ private object Invoke(MethodInfo methodInfo, object instance, IApplicationBuilde
             string environmentName,
             IList<string> diagnosticMessages)
         {
-            if (String.IsNullOrEmpty(applicationName))
+            if (string.IsNullOrEmpty(applicationName))
             {
-                return _next.LoadStartup(applicationName, environmentName, diagnosticMessages);
+                throw new ArgumentNullException("applicationName");
```
Eilon: Wrong exception type. This is thrown for null or empty, yet the exception is for null only. Use a regular ArgException with a message that says can't be null or empty (copy whatever message we used in WSR).

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -24,20 +22,21 @@ public class TestServer : IServerFactory, IDisposable
...
         private static readonly ServerInformation ServerInfo = new ServerInformation();
-        private Func<object, Task> _appDelegate;
+        private Func<IFeatureCollection, Task> _appDelegate;
         private IDisposable _appInstance;
         private bool _disposed = false;
 
         public TestServer(IConfiguration config, IServiceProvider serviceProvider, Action<IApplicationBuilder> appStartup)
         {
             var appEnv = serviceProvider.GetRequiredService<IApplicationEnvironment>();
+            var applicationLifeTime = serviceProvider.GetRequiredService<IApplicationLifetime>();
```
Eilon: See earlier comment about LifeTime --> Lifetime.

Eilon:
Overall looks good, just some further cleanup needed.

src/Microsoft.AspNet.Hosting/HostingContext.cs
```diff
@@ -21,7 +20,7 @@ public class HostingContext
         public Action<IApplicationBuilder> ApplicationStartup { get; set; }
         public RequestDelegate ApplicationDelegate { get; set; }
 
-        public string ServerName { get; set; }
+        public string ServerFactoryAssembly { get; set; }
```
Tratcher: That this is the name of an assembly is an implementation detail of the default IServerFactory. How about ServerFactoryLocation?

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -39,16 +40,16 @@ public IDisposable Start(HostingContext context)
             InitalizeServerFactory(context);
             EnsureApplicationDelegate(context);
 
-            var applicationLifetime = (ApplicationLifetime)context.Services.GetRequiredService<IApplicationLifetime>();
+            var applicationLifetime = (ApplicationLifetime)context.ApplicationLifeTime;
```
Tratcher: This cast makes it impossible to replace this service. Maybe the interface needs to be re-worked.

Praburaj: Yeah. But the cast is not introduced with this change. So I would keep that as a separate work item rather than mixing it with API review.

src/Microsoft.AspNet.Hosting/HostingContext.cs
Praburaj: Sure. That makes sense. I can fix that.

src/Microsoft.AspNet.TestHost/TestServer.cs
Praburaj: Will fix this.

src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
Praburaj: okay.

src/Microsoft.AspNet.Hosting/HostingContext.cs
Praburaj: Will fix this.

src/Microsoft.AspNet.Hosting/HostingEngine.cs
Praburaj: Actually I think I can fix it now. `IApplicationLifetime` to include these method names.

src/Microsoft.AspNet.Hosting/HostingEngine.cs
Praburaj: done.

Praburaj:
I addressed all comments except https://github.com/aspnet/Hosting/pull/167#discussion_r25447213. I'm filing a work item and fix that separately. Merging this change in.

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -94,9 +90,9 @@ private object Invoke(MethodInfo methodInfo, object instance, IApplicationBuilde
             string environmentName,
             IList<string> diagnosticMessages)
         {
-            if (String.IsNullOrEmpty(applicationName))
+            if (string.IsNullOrEmpty(applicationName))
             {
-                return _next.LoadStartup(applicationName, environmentName, diagnosticMessages);
+                throw new ArgumentException("applicationName");
```
Eilon: Wrong parameter to this exception. ArgException takes either "message" or "message, paramName". Use this message: https://github.com/aspnet/Mvc/blob/f6e701b2b2d01ae068b144416aeac080c623bc17/src/Microsoft.AspNet.Mvc.Core/Resources.resx#L205-L207

Praburaj: ah..fixed.

PR #171, opened 2015-03-02
Fixes #76 - improve error messages in Hosting

### LINKED ISSUE #76: Cleanup exception message when missing or incorrect configure method
I was trying projectk Custom hosting with console app. I had added startup.cs file without Configure method. Custom host was not even starting. after adding Configure method, custom host started.
I can provide sample if needed.

Eilon:
Removed the TODOs from messages, and generally made them more readable.


davidfowl:
Do we wanna leave it throwing `Exception` instead of `InvalidOperationException`

Eilon:
Thought about that though it's currently only us catching the exception and dumping to console. Think it's worth it?

Issue #151, opened 2015-02-04
IApplicationLifetime ApplicationStopping does not trigger if ConfigureServices method exists


davidfowl:
I spent a good few hours trying to track this down on mono and it seems like it happens on windows as well. 

```C#
using System;
using Microsoft.AspNet.Http;
using Microsoft.AspNet.Builder;
using Microsoft.AspNet.Hosting;
using Microsoft.Framework.DependencyInjection;

public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {

    }

    public void Configure(IApplicationBuilder app, IApplicationLifetime lifetime)
    {
        lifetime.ApplicationStopping.Register(OnStuff);
    }

    public void OnStuff()
    {
        Console.WriteLine("Goodbye!");
    }
}
````

The above code will not print out goodbye.

**Comment by Praburaj** [comment]
Related: https://github.com/aspnet/WebListener/issues/71

**Comment by davidfowl** [comment]
Is it? It happens when use kestrel. Seems like a bug in hosting

**Comment by Tragetaschen** [comment]
On Mono, Kestrel throws an Exception during its `Dispose` (aspnet/KestrelHttpServer#49). This means the cleanup code in `HostingEngine.Start`'s `Disposable` does not invoke [`applicationLifetime.SignalStopped()`](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/HostingEngine.cs#L51)

**Comment by davidfowl** [comment]
I'm not sure that's even related, when you remove ConfigureServices, the text `"Goodbye!"` gets printed

**Comment by davidfowl** [comment]
Here's my workaround:

```
using System;
using Microsoft.AspNet.Builder;
using Microsoft.AspNet.Hosting;
using Microsoft.Framework.DependencyInjection;

namespace WebApplication52
{
    public class Startup
    {
        public void ConfigureServices(IServiceCollection services, IApplicationLifetime lifetime)
        {
            services.AddInstance(lifetime);
        }

        public void Configure(IApplicationBuilder app, IApplicationLifetime lifetime)
        {
            lifetime.ApplicationStopping.Register(OnStuff);
        }

        public void OnStuff()
        {
            Console.WriteLine("Goodbye!");
        }
    }
}

```

Praburaj:
okay. Actually these are all problems related to app shutdown .

davidfowl:
[...]

PR #176, opened 2015-03-04
Adding IApplicationLifetime to the manifest

Praburaj:
Since IApplicationLifetime is not added to the manifest, while calling HostingServices.Create() before invoking ConfigureServices() we end up creating a new instance of IApplicationLifetime. So the Cancellationtoken that hosting triggers on appshutdown is different from what the app is exposed.

Fixes: https://github.com/aspnet/Hosting/issues/151

@HaoK @davidfowl


HaoK:
Can you also add `ILogger<>` as part of updating the manifest, see this https://github.com/aspnet/Hosting/commit/b510370c9b31b6f1f922d90735fb210f52ca0fe4

I didn't mean to back this change out when I was rolling back changes

Praburaj:
@HaoK - done

HaoK:
:shipit:

PR #184, opened 2015-03-06
Apply `configureHostServices` after the `fallbackProvider`

dougbu:
- allows MVC to override `IApplicationEnvironment` in functional tests

### DIFF
--- a/src/Microsoft.AspNet.Hosting/HostingServices.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingServices.cs
@@ -17,16 +17,19 @@ namespace Microsoft.AspNet.Hosting
         private static IServiceCollection Import(IServiceProvider fallbackProvider, Action<IServiceCollection> configureHostServices)
         {
             var services = new ServiceCollection();
-            if (configureHostServices != null)
-            {
-                configureHostServices(services);
-            }
             var manifest = fallbackProvider.GetRequiredService<IServiceManifest>();
             foreach (var service in manifest.Services)
             {
                 services.AddTransient(service, sp => fallbackProvider.GetService(service));
             }
+
             services.AddSingleton<IServiceManifest>(sp => new HostingManifest(services));
+
+            if (configureHostServices != null)
+            {
+                configureHostServices(services);
+            }
+
             return services;
         }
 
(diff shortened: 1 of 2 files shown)


msotclas:
Hi __@dougbu__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Doug Bunting (AAPT)). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

HaoK:
This fix isn't quite correct, as it currently stands, the new services won't be added to the manifest anymore

src/Microsoft.AspNet.Hosting/HostingServices.cs
```diff
@@ -17,16 +17,19 @@ public static class HostingServices
...
-            if (configureHostServices != null)
-            {
-                configureHostServices(services);
-            }
             var manifest = fallbackProvider.GetRequiredService<IServiceManifest>();
             foreach (var service in manifest.Services)
             {
                 services.AddTransient(service, sp => fallbackProvider.GetService(service));
             }
+
```
HaoK: Move the configure call above the call to adding the manifest and I think it will be correct

HaoK: Can you also add a new test that verifies the additional services are found in the manifest? :)

dougbu: odd, it's working for the MVC scenarios and that involves not only replacing but also adding services.  and `AddHosting` doesn't replace the manifest either.  is the `IServiceManifest` needed?

HaoK: MVC scenario might be different because MVC is doing fallback for all services internally (so MVC specifically might not care about the manifest)

dougbu:
Updated /cc @HaoK

HaoK:
:shipit:

dougbu:
Oops, didn't squash commits 89a1ec1cb5460f7f7cce6b8c0d6270625dd792fa through ee55e382606f37c77b2e2c10365602e982c18f58.

PR #190, opened 2015-03-13
Add support for auto creating services

HaoK:
- Copies ContainerMiddleware for now as step 1

cc @davidfowl @lodejard


HaoK:
Replaced by https://github.com/aspnet/Hosting/pull/192

Issue #100, opened 2014-10-14
Add extension method for getting environment name


rustd:
Proposed
```C#
if (env.IsEnvironment("Development"))
```
As it exists today
```C#
if (string.Equals(env.EnvironmentName, "Development", StringComparison.OrdinalIgnoreCase))
```

Praburaj:
https://github.com/aspnet/Hosting/commit/3b0d5fd422e46691cdeddefc5ebd17053c63b932

PR #194, opened 2015-03-16
Add extension method for getting environment name

Praburaj:
Fixes: https://github.com/aspnet/Hosting/issues/100

@rustd @muratg

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public static class HostingEnvironmentExtensions
+    {
+        /// <summary>
+        /// Compares the current hosting environment name against the specified value.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


muratg:
:shipit:

src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        /// <summary>
+        /// Compares the current hosting environment name against the specified value.
+        /// </summary>
+        /// <param name="hostingEnvironment">An instance of <see cref="IHostingEnvironment"/> service.</param>
+        /// <param name="environmentName">Environment name to validate against.</param>
+        /// <returns>True if the specified name is same as the current environment.</returns>
+        public static bool IsEnvironment(
+            [NotNull]this IHostingEnvironment hostingEnvironment,
+            [NotNull]string environmentName)
```
Eilon: Should this really throw on null?

Eilon: (string.Equals() doesn't care about null, so the rest of the code would still work just fine)

Praburaj: Is null a valid environment name? I think there is always a default in hosting layer. That's why I made it throw on null.

Eilon: Yeah I'm thinking more of a mock/lazy implementation of the interface that happens to return null. Would we want that to fail?

Praburaj: hmm. I think I can remove the null check on `environmentName`.

PR #200, opened 2015-03-20
Expose HostingEngine.CreateApplicationServices

HaoK:
@davidfowl @divega Thoughts?  Basically this will be the replacement for the old HostingServices.Create for how to create ApplicationServices without having to go through the whole HostingEngine/context setup.  This eliminates the need for implementing IServerFactory just to create an ApplicationServices.

### DIFF
--- a/src/Microsoft.AspNet.Hosting/HostingContext.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingContext.cs
@@ -17,7 +17,9 @@ namespace Microsoft.AspNet.Hosting
         public IApplicationBuilder Builder { get; set; }
 
         public string ApplicationName { get; set; }
-        public string WebRootPath { get; set; }
+        // REVIEW: BasePath and WebRootPath are basically setting same thing?
+        public string ApplicationBasePath { get; set; }
+        //public string WebRootPath { get; set; }
         public string EnvironmentName { get; set; }
         public StartupMethods StartupMethods { get; set; }
         public RequestDelegate ApplicationDelegate { get; set; }
--- a/src/Microsoft.AspNet.Hosting/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingEngine.cs
@@ -38,9 +38,20 @@ namespace Microsoft.AspNet.Hosting
             _fallbackServices = new WrappingServiceProvider(_fallbackServices, _hostingEnvironment, _appLifetime);
         }
 
+        public static IServiceProvider CreateApplicationServices(HostingContext context)
+        {
+            return CreateApplicationServices(context, fallbackServices: null);
+        }
+
+        public static IServiceProvider CreateApplicationServices(HostingContext context, IServiceProvider fallbackServices)
+        {
+            var engine = new HostingEngine(fallbackServices);
+            engine.EnsureApplicationServices(context);
+            return context.ApplicationServices;
+        }
+
         public IDisposable Start(HostingContext context)
         {
-            EnsureContextDefaults(context);
             EnsureApplicationServices(context);
             EnsureBuilder(context);
             EnsureServerFactory(context);
@@ -87,6 +98,8 @@ namespace Microsoft.AspNet.Hosting
 
         private void EnsureApplicationServices(HostingContext context)
         {
+            EnsureContextDefaults(context);
+
             if (context.ApplicationServices != null)
             {
                 return;
--- a/src/Microsoft.AspNet.Hosting/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingEngine.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
+using System.Runtime.Versioning;
 using System.Threading;
 using Microsoft.AspNet.Hosting.Builder;
 using Microsoft.AspNet.Hosting.Internal;
@@ -21,7 +23,7 @@ namespace Microsoft.AspNet.Hosting
 
         private readonly IServiceProvider _fallbackServices;
         private readonly ApplicationLifetime _appLifetime;
-        private readonly IApplicationEnvironment _applicationEnvironment;
+        private readonly IApplicationEnvironment _appEnv;
         private readonly HostingEnvironment _hostingEnvironment;
 
         private IServerLoader _serverLoader;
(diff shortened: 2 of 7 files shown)


HaoK:
So maybe we need to further tweak a bit, as we still need to short circuit the StartupLoader logic, see the matching SignalR change to use this method

https://github.com/aspnet/SignalR-Server/pull/89

HaoK:
Tweaked it a bit, moved default logic which adds no-op startup methods when calling CreateApplicationServices

HaoK:
Next question is if HostingContext is really the right input parameter... feels a bit too heavy still...

HaoK:
Added support for automatically replacing ApplicationName and BasePath, which is mainly what the tests were doing

src/Microsoft.AspNet.Hosting/HostingContext.cs
```diff
@@ -17,7 +17,9 @@ public class HostingContext
         public IApplicationBuilder Builder { get; set; }
 
         public string ApplicationName { get; set; }
-        public string WebRootPath { get; set; }
+        // REVIEW: BasePath and WebRootPath are basically setting same thing?
```
dougbu: isn't one the project directory and the other wwwroot?

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -21,7 +23,7 @@ public class HostingEngine
 
         private readonly IServiceProvider _fallbackServices;
         private readonly ApplicationLifetime _appLifetime;
-        private readonly IApplicationEnvironment _applicationEnvironment;
+        private readonly IApplicationEnvironment _appEnv;
```
dougbu: why switch from readable name to a confusing one?

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -29,18 +31,39 @@ public class HostingEngine
 
         public HostingEngine() : this(fallbackServices: null) { }
 
-        public HostingEngine(IServiceProvider fallbackServices)
+        public HostingEngine(IServiceProvider fallbackServices) : this(fallbackServices, applicationName: null, applicationPath: null) { }
```
dougbu: nit: long line

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -29,18 +31,39 @@ public class HostingEngine
...
             _fallbackServices = new WrappingServiceProvider(_fallbackServices, _hostingEnvironment, _appLifetime);
         }
 
+        public static IServiceProvider CreateApplicationServices(HostingContext context)
+        {
+            return CreateApplicationServices(context, fallbackServices: null);
+        }
+
+        // REVIEW: so ugly ApplicationName used here as input
+        public static IServiceProvider CreateApplicationServices(HostingContext context, IServiceProvider fallbackServices)
```
dougbu: nit long line

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -29,18 +31,39 @@ public class HostingEngine
...
+            _hostingEnvironment = new HostingEnvironment(_appEnv);
             _fallbackServices = new WrappingServiceProvider(_fallbackServices, _hostingEnvironment, _appLifetime);
         }
 
+        public static IServiceProvider CreateApplicationServices(HostingContext context)
+        {
+            return CreateApplicationServices(context, fallbackServices: null);
+        }
+
+        // REVIEW: so ugly ApplicationName used here as input
```
dougbu: do you mean "This method exists for use in functional tests, not from product code"?

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -29,18 +31,39 @@ public class HostingEngine
...
             _appLifetime = new ApplicationLifetime();
-            _applicationEnvironment = _fallbackServices.GetRequiredService<IApplicationEnvironment>();
-            _hostingEnvironment = new HostingEnvironment(_applicationEnvironment);
+            var appEnv = _fallbackServices.GetRequiredService<IApplicationEnvironment>();
+            _appEnv = new HostingApplicationEnvironment(appEnv, applicationPath, applicationName);
+            _hostingEnvironment = new HostingEnvironment(_appEnv);
             _fallbackServices = new WrappingServiceProvider(_fallbackServices, _hostingEnvironment, _appLifetime);
         }
 
+        public static IServiceProvider CreateApplicationServices(HostingContext context)
```
dougbu: what is the use case for this method?

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -252,6 +273,51 @@ public object GetService(Type serviceType)
             }
         }
 
+        // Represents an application environment that overrides the base path of the original
+        // application environment in order to make it point to the test application folder.
```
dougbu: comment is misleading. you're creating one of these for the mainline scenario as well, just with `appPath` and `appName` both `null`.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -25,22 +25,32 @@ public class TestServer : IServerFactory, IDisposable
...
         }
 
+        // REVIEW: we can configure services via AppStartup or via hostContext.Services
+        public TestServer(IConfiguration config, IServiceProvider serviceProvider, Action<IApplicationBuilder> configureApp, ConfigureServicesDelegate configureServices)
+            : this(serviceProvider, new HostingContext()
+                {
+                    ApplicationName = "Test App",
+                    Configuration = config,
+                    StartupMethods = new StartupMethods(configureApp, configureServices)
+                })
```
dougbu: nit: ick.  how about a static method that returns the `HostingContext`?

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -25,22 +25,32 @@ public class TestServer : IServerFactory, IDisposable
...
+        { }
+
         public Uri BaseAddress { get; set; } = new Uri("http://localhost/");
 
+        public static TestServer Create(HostingContext context)
```
dougbu: why does the test code now need to know `HostingContext`?  seems like moving the faked (but easily-created and understood) `IApplicationEnvironment` from test code into hosting has increased the concept count in code consuming these APIs.  would prefer to see just another `Create` method that takes the application name and path as parameters.

src/Microsoft.AspNet.Hosting/HostingContext.cs
```diff
@@ -17,7 +17,9 @@ public class HostingContext
...
+        public string ApplicationBasePath { get; set; }
+        //public string WebRootPath { get; set; }
```
dougbu: actually, not sure why this class is changing at all.  seems unrelated to avoiding creation of `IApplicationEnvironment` implementations in test code.

dougbu:
:watch:

HaoK:
Replacing with PR with next iteration of hosting https://github.com/aspnet/Hosting/issues/204

PR #202, opened 2015-03-24
Adding some tests for the MapPath extension method

Praburaj:
And looks like I pushed the MapPath extension change too soon. I had to fix an issue in the path for forward and backward slash.

For example if an application tries to do MapPath("/resource") then this code does not work cross plat unless the directory separator characters are fixed appropriately.

Fixed it and also added some test coverage.

@muratg

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
@@ -40,6 +40,8 @@ namespace Microsoft.AspNet.Hosting
                 return hostingEnvironment.WebRootPath;
             }
 
+            // On windows replace / with \.
+            virtualPath = virtualPath.Replace('/', Path.DirectorySeparatorChar);
             return Path.Combine(hostingEnvironment.WebRootPath, virtualPath);
         }
     }
--- a/test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
+++ b/test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
@@ -131,6 +131,32 @@ namespace Microsoft.AspNet.Hosting
             }
         }
 
+        [Theory]
+        [InlineData(null, "")]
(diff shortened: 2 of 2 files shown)


davidfowl:
I meant to ask about slashes

src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
```diff
@@ -40,6 +40,7 @@ public static class HostingEnvironmentExtensions
                 return hostingEnvironment.WebRootPath;
             }
 
+            virtualPath = virtualPath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
```
davidfowl: Do we want to support MapPath("x\y\z\") on *nix?

Praburaj: do you see any issues?

davidfowl: Just weird is all. If we call out the behavior explicitly that would be better. Rather than things working as implemented

Praburaj: `/` has to work always. How about this? 

```
virtualPath = Path.DirectorySeparatorChar == '\\' ?         // On windows replace / with \
                virtualPath.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar) :
                virtualPath;
```

davidfowl: Or, always translate `'/'` to the underlying `Path.DirectorySeparatorChar`

Praburaj: yeah. Updated the PR.

davidfowl:
:shipit:

Issue #204, opened 2015-03-24
Hosting rework part 2


HaoK:
- Add static `HostingEngineFactory.Create(IServiceProvider fallback, Action<IServiceCollection>)` (no overload without ISP)
- Add IHostingEngine Build/CreateApplicationServices(HostingContext context)
- IHostingEnvironment WebRootPath settable - Add note to not ever make it not settable
- Mvc Precompilation can configure RazorOptions instead of mucking around with IApplicationEnv

- Bring back IHostingEngine, IStartupLoader
- Remove HostingContext.Services, merge IHostingEngine and HostingContext as a worker
- IHostingEngineFactory.Create(IConfiguration)
 -- Start
 -- BuildApplicationServices
 -- Dispose
 -- WithServer x2
 -- WithStartup(string name) WithStartup(StartupDelegates)

cc @davidfowl @glennc @lodejard

HaoK:
Fixed 77e2dc263f11655312d4c73bb8e22d7b6254d485

PR #205, opened 2015-03-25
Hosting rework #2

HaoK:
Initial rough shape of discussed changes (https://github.com/aspnet/Hosting/issues/204)
Moved HostingContext to private class inside of HostingEngine for now (will nuke enitrely soon)
Next iteration will add back the HostingContainer and IStartupLoader

Main point of this initial clickstop is to make sure we like the public usage now

Program.cs
```
            var engine = HostingEngineFactory.Create(_serviceProvider, config)
                .UseServer(config.Get("server"))
                .UseStartup(config.Get("app"));
 
            var serverShutdown = engine.Start();
```
 and TestServer:
```
           var engine = HostingEngineFactory.Create(serviceProvider, config, configureServices)
                .UseServer(this)
                .UseStartup(configureApp, configureServices);
            _appInstance = engine.Start();
```

cc @lodejard @davidfowl


HaoK:
Question: do we want to expose UseEnvironment() to specifically target Development/Testing/Etc, the old Context had this property

HaoK:
- Added HostingEngineContainer, only IHostingEngine/IStartupLoader services added to hosting engine
- Added UseConfiguration/UseFallbackServices to IHostingEngine to preserve existing functionality/flow (can revisit to see if we can do this a different way)

New usage:
```
            var engine = HostingEngineFactory.Create(_serviceProvider)
                .UseConfiguration(config)
                .UseServer(config.Get("server"))
                .UseStartup(config.Get("app"));
```

HaoK:
Added support for specifying startup by type via: `UseStartup<T>() where T:class`

HaoK:
Added overload `UseStartup(Type startupType)` used for `TestServer.Create<T>`

HaoK:
Made TestServer extend HostingEngine so test servers can be configured the same as hosting engine without needing an extra layer of sugar/billion overloads.  Currently TestServer.Create still calls Start() automatically to preserve existing behavior

HaoK:
Replacing shortly with lou's feedback, new PR

PR #207, opened 2015-03-25
Create a logging scope having request id

kichalla:
Here we create a log scope per request so that any logs that get written fall into this scope giving a correlation. `SerilogLogger` is an example which logs the scope information for every log message that it writes.

I have verified this change in Weblistener and IIS but not in Kestrel though (will have to work with some other folks to test this).

Some integration points:
**Weblistener** 
https://github.com/aspnet/WebListener/blob/dev/src/Microsoft.AspNet.Server.WebListener/MessagePump.cs#L164

**Helios**
https://github.com/aspnet/Helios/blob/1f76783178e54d5e9c020a27a2cfbd6f2dda1d53/src/Microsoft.AspNet.Loader.IIS/RuntimeHttpApplication.cs#L113

**Kestrel**
https://github.com/aspnet/KestrelHttpServer/blob/cb3def56683cc9e0294b5e78fb70f686be4074ee/src/Kestrel/ServerFactory.cs#L50

Yes, doing this in hosting layer will cause the log messages written at specific hosts to not have this scope(this is still an open discussion and will revisit later).

**Please note**: This PR does not have any unit tests yet. I wanted to get this out to get some initial comments.

@davidfowl @Praburaj @rynowak @loudej

### DIFF
--- a/src/Microsoft.AspNet.Hosting/DefaultRequestIdentifierFeature.cs
+++ b/src/Microsoft.AspNet.Hosting/DefaultRequestIdentifierFeature.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Http;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public class DefaultRequestIdentifierFeature : IRequestIdentifierFeature
+    {
+        public DefaultRequestIdentifierFeature()
+        {
+            TraceIdentifier = Guid.NewGuid();
+        }
+
+        public Guid TraceIdentifier { get; }
+    }
+}
--- a/src/Microsoft.AspNet.Hosting/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingEngine.cs
@@ -6,11 +6,14 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.FeatureModel;
 using Microsoft.AspNet.Hosting.Builder;
 using Microsoft.AspNet.Hosting.Server;
 using Microsoft.AspNet.Hosting.Startup;
+using Microsoft.AspNet.Http;
 using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Logging;
 
 namespace Microsoft.AspNet.Hosting
 {
@@ -24,7 +27,7 @@ namespace Microsoft.AspNet.Hosting
 
         // Start/ApplicationServices block use methods
         private bool _useDisabled;
-        
+
         private IApplicationBuilderFactory _builderFactory;
         private IApplicationBuilder _builder;
         private IServiceProvider _applicationServices;
@@ -56,14 +59,20 @@ namespace Microsoft.AspNet.Hosting
 
             var applicationDelegate = BuildApplicationDelegate();
 
-            var _contextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
-            var _contextAccessor = _applicationServices.GetRequiredService<IHttpContextAccessor>();
+            var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
+            var contextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
+            var contextAccessor = _applicationServices.GetRequiredService<IHttpContextAccessor>();
             var server = _serverFactory.Start(_serverInstance,
-                features =>
+                async features =>
                 {
-                    var httpContext = _contextFactory.CreateHttpContext(features);
-                    _contextAccessor.HttpContext = httpContext;
-                    return applicationDelegate(httpContext);
+                    var httpContext = contextFactory.CreateHttpContext(features);
+                    var requestIdentifier = GetRequestIdentifier(httpContext);
+
+                    using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
+                    {
+                        contextAccessor.HttpContext = httpContext;
+                        await applicationDelegate(httpContext);
+                    }
                 });
 
             return new Disposable(() =>
@@ -163,6 +172,18 @@ namespace Microsoft.AspNet.Hosting
             }
         }
 
+        private Guid GetRequestIdentifier(HttpContext httpContext)
+        {
+            var requestIdentifierFeature = httpContext.GetFeature<IRequestIdentifierFeature>();
+            if (requestIdentifierFeature == null)
+            {
+                requestIdentifierFeature = new DefaultRequestIdentifierFeature();
+                httpContext.SetFeature<IRequestIdentifierFeature>(requestIdentifierFeature);
+            }
+
+            return requestIdentifierFeature.TraceIdentifier;
+        }
+
         // Consider cutting
         public IHostingEngine UseEnvironment(string environment)
         {
@@ -207,7 +228,7 @@ namespace Microsoft.AspNet.Hosting
         public IHostingEngine UseStartup(Action<IApplicationBuilder> configureApp, Action<IServiceCollection> configureServices)
         {
             CheckUseAllowed();
-            _startup = new StartupMethods(configureApp, 
+            _startup = new StartupMethods(configureApp,
                 services => {
                     if (configureServices != null)
                     {
--- a/test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
+++ b/test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
@@ -7,12 +7,16 @@ using System.IO;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.FeatureModel;
+using Microsoft.AspNet.Hosting.Builder;
 using Microsoft.AspNet.Hosting.Server;
 using Microsoft.AspNet.Hosting.Startup;
+using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Testing.xunit;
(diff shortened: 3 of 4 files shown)


msotclas:
Hi __@kichalla__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Kiran Challa). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

kichalla:
ping

src/Microsoft.AspNet.Hosting/Logging/RequestInformationValues.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <inheritdoc />
+        public string Format()
+        {
+            return LogFormatter.FormatLogValues(this);
+        }
+
+        /// <inheritdoc />
+        public IEnumerable<KeyValuePair<string, object>> GetValues()
+        {
+            return new[]
```
IDisposable: yield return new KeyValuePair<string, object>(nameof(RequestId), RequestId); // instead of a new[]?

davidfowl: Why?

IDisposable: First, it's tons clearer that we don't want an array, we just want one ```KeyValuePair<string, object>```

Second, unless someone calls ```.Next()``` and ```.Current```, the actual ```KeyValuePair<string, object>``` is never allocated.

Third, this is an allocation a ```KeyValuePair<string, object>```, then an allocation of an array and a ```.Add()``` on that array. This array is then returned, which on call to ```IEnumerable<KeyValuePair<string, object>>GetEnumerator()``` returns an ```IEnumerator<KeyValuePair<string, object>>```. That enumerator's code is quite small, but still exists as an allocation.

Meanwhile, the ```yield return``` version news up a mini-state machine (about a wash for the new of the array) which directly handles the ```IEnumerator<KeyValuePair<string, object>>``` logic (thus no second allocation).

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -47,14 +50,21 @@ public IDisposable Start(HostingContext context)
...
             var contextFactory = context.ApplicationServices.GetRequiredService<IHttpContextFactory>();
             var contextAccessor = context.ApplicationServices.GetRequiredService<IHttpContextAccessor>();
             var server = context.ServerFactory.Start(context.Server,
-                features =>
+                async features =>
                 {
-                    var httpContext = contextFactory.CreateHttpContext(features);
-                    contextAccessor.HttpContext = httpContext;
-                    return context.ApplicationDelegate(httpContext);
+                    var requestIdentifier = GetRequestIdentifier(features);
```
davidfowl: Should we avoid the scope altogether if there's no request identifier?

kichalla: Sure, I am not too strong on this as currently its `Guid.Empty` and not useful...a related question: Should we create a default request identifier feature in case one is not provided by the host?(ex: Kestrel)...I deliberately did not create a new Guid over here as this would be specific to logging scope and a user might try to get the identifier feature in their own code for whatever reason and would not see this id...

kichalla:
Updated PR as discussed in email and added tests. 

/cc: @lodejard @loudej @davidfowl @Tratcher @rynowak @Eilon

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -56,14 +59,20 @@ public virtual IDisposable Start()
 
             var applicationDelegate = BuildApplicationDelegate();
 
+            var _logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
```
Eilon: OK weird. Why do all these local vars have underscores?

test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
```diff
@@ -150,6 +154,87 @@ public void MapPath_Windows_Facts(string virtualPath, string expectedSuffix)
             RunMapPath(virtualPath, expectedSuffix);
         }
 
+        [Fact]
+        public void HostingEngine_CreatesDefaultRequestIdentifierFeature_IfNotPresent()
+        {
+            // Arrange
+            HttpContext httpContext = null;
+            var requestDelegate = new RequestDelegate((httpCntxt) =>
```
Eilon: `httpCntxt` --> `innerHttpContext`? The abbreviation looks more like a typo. Also, can remove the parens around it (single-param lambdas don't need parens).

test/Microsoft.AspNet.Hosting.Tests/project.json
```diff
@@ -5,7 +5,8 @@
         "Microsoft.AspNet.Testing": "1.0.0-*",
         "Microsoft.Framework.OptionsModel": "1.0.0-*",
         "Microsoft.Framework.Runtime.Interfaces": "1.0.0-*",
-        "xunit.runner.aspnet": "2.0.0-aspnet-*"
+        "xunit.runner.aspnet": "2.0.0-aspnet-*",
+        "Moq": "4.2.1312.1622"
```
Eilon: Sort 'em.

src/Microsoft.AspNet.Hosting/DefaultRequestIdentifierFeature.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Microsoft.AspNet.Hosting
+{
+    public class DefaultRequestIdentifierFeature : IRequestIdentifierFeature
+    {
+        public DefaultRequestIdentifierFeature()
+        {
+            TraceIdentifier = Guid.NewGuid();
+        }
+
+        public Guid TraceIdentifier { get; }
```
Eilon: For your FYI information in the future I think this will change from a `Guid` to a `string` but don't worry about that here.

Eilon:
Overall looks :shipit: to me, just had a few stylistic comments.

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -56,14 +59,20 @@ public virtual IDisposable Start()
...
             var _contextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
             var _contextAccessor = _applicationServices.GetRequiredService<IHttpContextAccessor>();
             var server = _serverFactory.Start(_serverInstance,
                 features =>
                 {
-                    var httpContext = _contextFactory.CreateHttpContext(features);
-                    _contextAccessor.HttpContext = httpContext;
-                    return applicationDelegate(httpContext);
+                    var requestIdentifier = GetRequestIdentifier(features);
```
Tratcher: This needs to be re-arranged. At this point there's no grantee the feature collection is mutable so this can fail if you try to add to it (like on Kestrel?).  The context factory wraps this in a FeatureCollection to ensure it's mutable (https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs#L13). You should move `_contextFactory.CreateHttpContext(features);` up and then pass the HttpContext into GetRequestIdentifier.

src/Microsoft.AspNet.Hosting/HostingEngine.cs
kichalla: yeah, changed it now...thanks...

src/Microsoft.AspNet.Hosting/DefaultRequestIdentifierFeature.cs
kichalla: Sure

src/Microsoft.AspNet.Hosting/HostingEngine.cs
kichalla: Good point!...actually it works fine on Kestrel, but I see your point...

kichalla:
Updated. /cc: @Eilon @Tratcher

Eilon:
The changes for my comments look good. Need @Tratcher to review the feature collection stuff.

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -56,14 +59,20 @@ public virtual IDisposable Start()
...
-                    var httpContext = _contextFactory.CreateHttpContext(features);
-                    _contextAccessor.HttpContext = httpContext;
-                    return applicationDelegate(httpContext);
+                    var httpContext = contextFactory.CreateHttpContext(features);
+                    var requestIdentifier = GetRequestIdentifier(httpContext);
+
+                    using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
+                    {
+                        contextAccessor.HttpContext = httpContext;
+                        return applicationDelegate(httpContext);
```
Tratcher: This needs to be `await applicationDelegate(httpContext);` for the using statement to work correctly.

Tratcher:
:shipit: after fixing the missing await.

src/Microsoft.AspNet.Hosting/HostingEngine.cs
kichalla: :+1: I suppose you are referring to the scope(async local) availability...I initially had async-await, but wasn't sure if it was necessary...added it now...

kichalla:
Updated /cc: @Tratcher

src/Microsoft.AspNet.Hosting/HostingEngine.cs
Tratcher: No, I meant that `return applicationDelegate(httpContext);` exits the using statement as soon as any Task is returned from the applicationDelegate. You need to use await to make sure the returned task actually finishes before you close the scope.
The changes look good. :shipit:

PR #208, opened 2015-03-26
Add IsDevelopment and IsProduction extension methods

glennc:
Production and Development are the two names that we will have baked into the stack. Development is what VS will set when running applications, and Production will be the default that the DNX sets.

We should provide these methods so I don't need to guess what I should put in there, or look up the docs, when I am starting a new application without the scaffolding.

I think it looks slightly nicer than IsEnvironment("Development") as well.

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
@@ -9,6 +9,29 @@ namespace Microsoft.AspNet.Hosting
 {
     public static class HostingEnvironmentExtensions
     {
+        private const string DevelopmentEnvironmentName = "Development";
+        private const string ProductionEnvironmentName = "Production";
+
+        /// <summary>
+        /// Checks if the current hosting environment name is development.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


glennc:
@anurse @davidfowl @lodejard

src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
```diff
@@ -9,6 +9,29 @@ namespace Microsoft.AspNet.Hosting
 {
     public static class HostingEnvironmentExtensions
     {
+        private const string DevelopmentEnvironmentName = "Development";
```
anurse: Should we put `public static readonly` versions of these somewhere? It would probably be good for devs to be able to access the "official" Dev and Production names.

anurse:
:shipit:

src/Microsoft.AspNet.Hosting.Interfaces/HostingEnvironmentExtensions.cs
davidfowl: Yar

davidfowl:
I merged it in for you @glennc

PR #210, opened 2015-03-26
Hosting rework #3

HaoK:
Incorporated @lodejard feedback:

Introduce static WebApplication which is the static helpers used to create hosting engines and factories
- IHostingEngineFactory / IStartupLoader are the two hosting engine services which can be replaced
- IHostingEngine/HostingEngine are the actual workers which (is the merged result of Engine/Context)
- Nuked all of the wrapping service providers, made IHostingEnviroment get/set, and we now directly manipulate the instance.

cc @davidfowl @lodejard


src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -5,253 +5,340 @@
...
+        private IServerInformation _serverInstance;
+
+        public HostingEngine(IServiceCollection appServices, IStartupLoader startupLoader, IConfiguration config, IHostingEnvironment hostingEnv, string appName)
         {
-            _fallbackServices = fallbackServices ?? CallContextServiceLocator.Locator.ServiceProvider;
-            _appLifetime = new ApplicationLifetime();
-            _applicationEnvironment = _fallbackServices.GetRequiredService<IApplicationEnvironment>();
-            _hostingEnvironment = new HostingEnvironment(_applicationEnvironment);
-            _fallbackServices = new WrappingServiceProvider(_fallbackServices, _hostingEnvironment, _appLifetime);
+            _config = config ?? new Configuration();
```
davidfowl: Dont allow nulls

src/Microsoft.AspNet.Hosting/IHostingEngine.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public interface IHostingEngine
+    {
+        IDisposable Start();
+        void Dispose();
+
+        // Accessing this will block Use methods
+        IServiceProvider ApplicationServices { get; }
+
+        // Use methods blow up after any of the above methods are called
```
davidfowl: Why don't we separate these?

src/Microsoft.AspNet.Hosting/IHostingEngine.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.AspNet.Hosting.Startup;
+using Microsoft.Framework.ConfigurationModel;
+using Microsoft.Framework.DependencyInjection;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public interface IHostingEngine
```
davidfowl: And why isn't this object IDisposable

src/Microsoft.AspNet.Hosting.Interfaces/IHostingEnvironment.cs
```diff
@@ -7,10 +7,13 @@ namespace Microsoft.AspNet.Hosting
 {
     public interface IHostingEnvironment
     {
+        // This must be settable!
         string EnvironmentName { get; set;  }
```
davidfowl: Can you change this in the application so that you start off as dev then end up as prod? Is that what we want?

HaoK:
Added support for `TestServer.Create<TStartup>` Will address PR feedback next

HaoK:
Added support for swapping out Appname and BasePath in TestServer

src/Microsoft.AspNet.Hosting/IHostingEngine.cs
HaoK: We tried this, and decided against splitting builder/engine

HaoK:
Added full permutations of overloads for WebApplication statics, cleans up music store and some test usage nicely

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -5,251 +5,232 @@
...
+        private readonly IStartupLoader _startupLoader;
+        private readonly ApplicationLifetime _applicationLifetime;
+        private readonly IHostingEnvironment _hostingEnvironment;
 
-        private readonly IServiceProvider _fallbackServices;
-        private readonly ApplicationLifetime _appLifetime;
-        private readonly IApplicationEnvironment _applicationEnvironment;
-        private readonly HostingEnvironment _hostingEnvironment;
+        // Start/ApplicationServices block use methods
+        private bool _useDisabled;
```
davidfowl: Do we even need this?

HaoK: I can remove the guards, but then its up to the user to beware and know that calling Use after Start() or accessing ApplicationServices is ignored / no-op

HaoK:
Renamed WebApplication -> AspNetHosting and dropped Hosting from CreateFactory/Engine
Bury'd and uglified IHostingServicesBuilder to a super scary non interfaced service in Internal namespace: RootServiceCollectionInitalizer !!

davidfowl:
Omg, AspNetHosting? That's bizzare. Can we not put product names in the API please?

HaoK:
Hosting conflicts with the namespace, so we need to making it something like Hosting but slightly diff... Host?

davidfowl:
Microsoft.AspNet.Hosting.WebHost?

davidfowl:
WebHost.Create?

HaoK:
WebHost.CreateEngine/Factory  sounds good to me, lets go with that

davidfowl:
Also, the file is still called WebApplication.cs

src/Microsoft.AspNet.Hosting/Internal/AutoRequestServicesStartupFilter.cs
```diff
@@ -14,7 +14,10 @@ public Action<IApplicationBuilder> Configure(IApplicationBuilder app, Action<IAp
             return builder =>
             {
                 app.UseMiddleware<RequestServicesContainerMiddleware>();
-                next(builder);
+                if (next != null)
```
davidfowl: next can be null?

HaoK:
Renamed WebHost

src/Microsoft.AspNet.Hosting/Internal/AutoRequestServicesStartupFilter.cs
HaoK: Probably was an edge case with start up filters/mangled app pipelines, but I hit this yes...

davidfowl: It should never be null, this would be awful to code against.

HaoK: Fair enuf, removed

HaoK:
HostingEngine will now throw if Start() is called when a server wasn't configured via UseServer()

test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
```diff
@@ -21,55 +23,51 @@ public class HostingEngineTests : IServerFactory
         [Fact]
         public void HostingEngineCanBeStarted()
         {
-            var context = new HostingContext
-            {
-                ServerFactory = this,
-                ApplicationName = "Microsoft.AspNet.Hosting.Tests"
-            };
+            var engine = WebHost.CreateEngine()
```
davidfowl: this isn't the engine, call it server

HaoK:
Merged 77e2dc263f11655312d4c73bb8e22d7b6254d485

PR #219, opened 2015-03-31
Add Program tests and null guard for UseStartup

HaoK:
cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Hosting/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingEngine.cs
@@ -124,7 +124,7 @@ namespace Microsoft.AspNet.Hosting
                 // Blow up if we don't have a server set at this point
                 if (_serverFactoryLocation == null)
                 {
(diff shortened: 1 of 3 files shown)


davidfowl:
:shipit:

src/Microsoft.AspNet.Hosting/HostingEngine.cs
```diff
@@ -124,7 +124,7 @@ private void EnsureServer()
                 // Blow up if we don't have a server set at this point
                 if (_serverFactoryLocation == null)
                 {
-                    throw new InvalidOperationException("UseStartup() is required for Start()");
+                    throw new InvalidOperationException("UseServer() is required for Start()");
```
Eilon: Use `nameof()`?

Eilon: (Can do it with fancy C# 6 string interpolation too.)

HaoK:
Rolled these changes in with https://github.com/aspnet/Hosting/pull/237/files, closing this PR

Issue #172, opened 2015-03-02
Move IServerInformation and IServerFactory into a package that server authors can use


davidfowl:
Today IServerInformation exists in Microsoft.AspNet.Http and IServerFactory exists in Hosting. We need to have a simple package that server authors like Nowin can use to implement the ASP.NET 5 server abstraction.

davidfowl:
## Proposal

**Microsoft.AspNet.Hosting.Interfaces** - Interfaces made available to applications that use the ASP.NET hosting model.

- IHttpContextAccessor.cs
- IHostingEnvironment.cs

**Microsoft.AspNet.Hosting.Server.Interfaces** - The WebServer contract.

- IServerFactory.cs

Depends on
- Microsoft.Framework.ConfigurationModel.Interfaces
- Microsoft.AspNet.FeatureModel
- Microsoft.AspNet.Http

Some interfaces stay inside of the hosting package itself since they are pretty much implementation details.

- IApplicationLifetime.cs
- IConfigureHostingEnvironment.cs
- IHostingEngine.cs
- IStartupLoader.cs
- IServerLoader.cs
- IApplicationBuilderFactory.cs
- IHttpContextFactory.cs

Thoughts? /cc @lodejard @glennc @Eilon @muratg

muratg:
Sounds fine to me.  How about IServerInformation?

muratg:
@davidfowl assuming you wanted this in RC, assigning this to Prabu.

davidfowl:
Renamed Hosting.Server interfaces to M.A.Server.Interfaces

HaoK:
Since we already have a .Server subnamespace, maybe it would be clearer if we had a Engine namespace for the engine and the four interfaces that are only used by the HostingEngine?  i.e. 

M.A.Hosting.Engine:
- HostingEngine
- IHostingEngine
- IApplicationBuilderFactory
- IHttpContextFactory
- IStartupFilter
- IServerLoader

davidfowl:
I would move things out of the Engine namespace that can be overridden by the application

HaoK:
So maybe collapse everything into root Microsoft.AspNet.Hosting except for

IHostingEngine and IStartupLoader which live in Microsoft.AspNet.Hosting.Engine, and the ones you've already called out?

davidfowl:
Yep

HaoK:
[...]

PR #223, opened 2015-04-01
Move interfaces around

davidfowl:
- Moved IServerFactory to Microsoft.AspNet.Server.Interfaces
- Moved IHttpContextAccessor to Microsoft.AspNet.Hosting.Interfaces
- Fixed dependencies so that they are minimal

#172

/cc @lodejard @HaoK @Praburaj @rynowak 

Should we rename the Server.Interfaces to Hosting.Server.Interfaces? (seems long to me so I reduced it). Also, the namespace is currently `Microsoft.AspNet.Hosting.Server`


Eilon:
Nice :shipit:

PR #227, opened 2015-04-02
Have StartupManagerTests use real fallback services

HaoK:
cc @davidfowl


davidfowl:
ughh, do we want this?

HaoK:
Fine

PR #232, opened 2015-04-08
Add overload for creating test server builder

HaoK:
Needed to fix template tests, need an overload which lets the Startuploader do its thing

cc @davidfowl @Tratcher


Tratcher:
That is a lot of overloads...
:shipit:

HaoK:
Merged f337f8cd04459e541431df04d859b705f3d2f6f7

Issue #236, opened 2015-04-10
API Review: Hosting Engine


glennc:
```csharp
public void Main()
{
    var builder = new WebHostBuilder(_serviceProvider, _config);

    builder.UseServices((sc) =>
    {
        sc.AddSingleton<IMyCustomService>();
    });
    builder.UseEnvironment("Staging");
    builder.UseServer("Microsoft.AspNet.Server.WebListener");

    var host = builder.Build();

    var waitHandle = new ManualResetEventSlim();

    var shutdown = host.ApplicationServices.GetRequiredService<IApplicationShutdown>();
    shutdown.ShutdownRequested.Register(() =>
    {
        host.Dispose();
        waitHandle.Set();
    });

    host.Start();

    waitHandle.Wait();
}
```

* IHostingEngine moves to internal namespace
* New WebHostBuilder to build an engine, Use* methods from IHostingEngine move to it
* IServiceProvider is required for the builder. If you don't provide a configuration we use an InMemory one.
* Add UseUrl, UseUrls, UsePort methods to the builder. These set known keys on Configuration.
* UseServices method added to builder. Accepts the configureServices action.
* builder.Build needs to do the same thing that WebHost was doing in Create previously.

HaoK:
@davidfowl Remind me why we didn't keep the iteration where we had Builder/Engine split like this before?  I know we had this split but decided to drop it...

davidfowl:
That came up today and now I can't remember (that why we called you! :smile:). I might still exist somewhere in a PR.

HaoK:
unfortunately I think this was in lync, does lync save history?

davidfowl:
Only sometimes

HaoK:
I think the difference in usage pattern here is configuring the service before calling start.  I don't think we considered that scenario when we killed the Builder/Engine split.  If we didn't expose ApplicationServices there isn't too much you could do with the engine after building it

HaoK:
But I like this change, so I'm not trying to push back, just want to make sure there wasn't a real reason why we didn't do it before...

Can we add `UseStartup<TStartup>` for the builder?  I'm currently having issues getting this to work properly only in the TestServer, since we want TStartup to be activated from the builder's SP to work correctly

davidfowl:
I think there is a UseStartup. The Builder has everything that the IHostingEngine had before that includes UseStartup

HaoK:
[...]

PR #237, opened 2015-04-13
API review changes

HaoK:
Rename IHostingFactory -> IHostingBuilder and move Use methods from IHostingEngine
Rename HostingFactory -> WebHostBuilder
Move HostingEngine to internal

Fixes https://github.com/aspnet/Hosting/issues/236

cc @davidfowl @lodejard @Tratcher


src/Microsoft.AspNet.Hosting/Internal/RootHostingServiceCollectionInitializer.cs
```diff
@@ -43,7 +41,7 @@ public IServiceCollection Build()
             services.AddInstance(this);
             services.AddInstance(_loggerFactory);
 
-            services.AddTransient<IHostingFactory, HostingFactory>();
+            services.AddTransient<IHostingBuilder, WebHostBuilder>();
```
davidfowl: This shouldn't be a service. It's something you new up with no state.

davidfowl:
Some extra notes that weren't in the bug
- WebHost shouldn't exist anymore
- WebHostBuilder is *not* a service and there doesn't need to be an interface
- IConfiguration is optional to WebHostBuilder and should not be an argument to Build.
- WebHostBuilder has 2 ctors -> WebHostBuilder(IServiceProvider), WebHostBuilder(IServiceProvider, IConfiguration)
- Program.cs shouldn't have any code that reads configuration. The WebHostBuilder uses defaults from configuration when the constructor that takes IConfiguration is used.

HaoK:
@davidfowl Can we expose UseConfiguration instead of ctor overload?

HaoK:
Updated PR, nuked WebHost

HaoK:
Split IStartupLoader methods into FindStartupType/LoadMethods which enables `UseStartup<T>` also moved environment off interface methods, service can get it from IHostingEnvironment

davidfowl:
> @davidfowl Can we expose UseConfiguration instead of ctor overload?

We decided against that since config is something that needs to be ready super early. There were other reasons as well /cc @glennc @lodejard

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -0,0 +1,164 @@
...
+        internal Type StartupType { get; set; }
+
+        // Only one of these should be set
+        internal IServerFactory ServerFactory { get; set; }
+        internal string ServerFactoryLocation { get; set; }
+        private IServerInformation _serverInstance;
+
+        public HostingEngine(IServiceCollection appServices, IStartupLoader startupLoader, IConfiguration config)
+        {
+            _config = config ?? new Configuration();
```
davidfowl: This should never be null

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -23,31 +24,37 @@ public Program(IServiceProvider serviceProvider)
...
-
-            var engine = WebHost.CreateEngine(_serviceProvider, config);
+            var builder = new WebHostBuilder(_serviceProvider);
             var server = config.Get("server");
             if (server != null)
             {
-                engine.UseServer(server);
+                builder.UseServer(server);
             }
             var startup = config.Get("app");
```
davidfowl: This code should be gone. Passing the configuration to the WebHostBuilder makes this logic obsolete as the builder should use the config system passed in as a sign to set these default.

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+        private string _serverFactoryLocation;
+        private IServerFactory _serverFactory;
+
+        public WebHostBuilder() : this(null) { }
+
+        public WebHostBuilder(IServiceProvider fallbackServices)
+        {
+            _hostingEnvironment = new HostingEnvironment();
+            _loggerFactory = new LoggerFactory();
+            _fallbackServices = fallbackServices ?? CallContextServiceLocator.Locator.ServiceProvider;
```
davidfowl: Drop support for CallContextServiceLocator.Locator.ServiceProvider, we're removing this across the board. It can stay in test host for now but all mainline APIs should remove it.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -48,51 +48,51 @@ public static TestServer Create(Action<IApplicationBuilder> configureApp, Action
...
         {
-            return CreateBuilder(fallbackServices, config, configureApp, configureServices).Build();
+            return new TestServer(CreateBuilder(fallbackServices, config, configureApp, configureServices));
         }
 
-        public static TestServer Create<TStartup>() where TStartup : class
-        {
-            return Create<TStartup>(fallbackServices: null, config: null, configureApp: null, configureServices: null);
-        }
+        //public static TestServer Create<TStartup>() where TStartup : class
```
davidfowl: Dead code?

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+        private Type _startupType;
+        private string _startupAssemblyName;
+
+        // Only one of these should be set
+        private string _serverFactoryLocation;
+        private IServerFactory _serverFactory;
+
+        public WebHostBuilder() : this(null) { }
+
+        public WebHostBuilder(IServiceProvider fallbackServices)
```
davidfowl: Rename fallbackServices to services

src/Microsoft.AspNet.Hosting/Program.cs
HaoK: So we want to bake "server" and "app" in to the builder as default config keys for UseServer and UseStartup?  Should we pick new new keys?

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
HaoK: So [NotNull] for IServiceProvider then?

src/Microsoft.AspNet.Hosting/Program.cs
davidfowl: might be the time to pick new keys /cc @lodejard. server and app are terrible. Lets keep new app and server as a fallback. Maybe something like:

``
var server = config.Get("Hosting:Server") ?? config.Get("server");
var app = config.Get("Hosting:Application") ?? config.Get("app");
``

davidfowl: Also do the same for Hosting:Environment?

HaoK: Okay that looks much better, I can do the same thing with the environment name: "ASP_ENV" or "Hosting:Environment"

davidfowl: Although I realized today that on *nix env variables can't contain `';'`.

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
davidfowl: Yep

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
HaoK: So today after nuking the default ctor(), we are calling GetRequiredService on the ISP, so things go kerboom if you try to pass in any arbitrary ISP...  What behavior do we want here?  The tests need to either pass in the old CallContextServiceLocator, or we can fix the import to allow this?

HaoK: Actually I guess the hosting engine tests have to pass the CallContextSeviceLocator since it relies on the services there...

src/Microsoft.AspNet.TestHost/TestServer.cs
HaoK: yep nuked and cleaned up TestServer

HaoK:
Addressed PR feedback

test/Microsoft.AspNet.TestHost.Tests/TestServerTests.cs
```diff
@@ -56,53 +56,53 @@ public void ThrowsIfNoApplicationEnvironmentIsRegisteredWithTheProvider()
...
-                },
-                configureServices: null);
-
-            builder.ApplicationBasePath = appPath;
-            var server = builder.Build();
-
-            string result = await server.CreateClient().GetStringAsync("/path");
-            Assert.Equal("AppPath:" + appPath, result);
-        }
+        //[Fact]
```
davidfowl: What's the deal with these tests?

HaoK: I used to have support on TestServer for overriding the ApplicationBasePath/ApplicationName, not sure we still need this with the config changes to allow setting the base path directly?  I ripped out the support for now and MVC functionals pass without this, so I think I can just nuke this once and for all

HaoK: Nuked the AppPathName tests, reenabled the Create<T> tests for TestServer

src/Microsoft.AspNet.Hosting/Program.cs
HaoK: So we want config to win right?  Where config is now going to override anything set problematically before hand?  Or do we want config to be the defaults applied in the ctor?  and UseStartup will win?

HaoK: Today it looks like config Environment wins over UseEnvironment (but I'm not sure that's what we want either...)

HaoK:
Pushed config server/application key logic into WebHostBuilder

HaoK:
ping @davidfowl 

@glennc nag Fowler if you want to play with these changes soon

HaoK:
Rebased and fixed tests to accommodate new RequestIdentifier changes

HaoK:
234bbf82f2329e1ce2092962264ba5d4567d115b

PR #244, opened 2015-04-20
Removing timeout for dnu publish

Praburaj:
Sometimes on slower machines that it takes more than this time especially when -no-source is turned on.

@Tratcher


Tratcher:
:shipit: I hope this only affects music store. 60 seconds would be prohibitively long in any of the tests I want to run.

Praburaj:
that's right. I observed this issue on the cross plat CI for MusicStore.

PR #245, opened 2015-04-20
Add more logging while starting dnx.exe or dnu.cmd

Praburaj:
@Tratcher 

Previously there was a failure on CI on turning this on. Investigating the failure by changing the flag as there is no better way to debug. If this still fails will rollback.


Praburaj:
Talked with @Tratcher offline. Pushing the change.

PR #248, opened 2015-04-21
Assigning application pool runtime version to 4.0.30319 explicitly

Praburaj:
I'm noticing that leaving it to an empty value defaults to 4.0 runtime on Win8.1 and above. But on downlevel OSes this defaults to v2.0 app pool resulting in errors.

Always assigning the app pool version explicitly.

@Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Server.Testing/Deployers/IISDeployer.cs
+++ b/src/Microsoft.AspNet.Server.Testing/Deployers/IISDeployer.cs
@@ -117,8 +117,9 @@ namespace Microsoft.AspNet.Server.Testing
 
         private class IISApplication
         {
-            private const string WEBSITE_NAME = "ASPNETTESTRUNS";
-            private const string NATIVE_MODULE_MANAGED_RUNTIME_VERSION = "vCoreFX";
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Testing/Deployers/IISDeployer.cs
```diff
@@ -117,8 +117,9 @@ public override void Dispose()
 
         private class IISApplication
         {
-            private const string WEBSITE_NAME = "ASPNETTESTRUNS";
-            private const string NATIVE_MODULE_MANAGED_RUNTIME_VERSION = "vCoreFX";
+            private const string WebSite_Name = "ASPNETTESTRUNS";
```
Tratcher: If you're going to fix the name style, also remove the underscores.

Tratcher:
:shipit:

src/Microsoft.AspNet.Server.Testing/Deployers/IISDeployer.cs
Praburaj: good point. Fixed.

PR #253, opened 2015-05-01
Application started action

victorhurdugaci:
In response to this https://github.com/aspnet/Hosting/pull/129, here is the solution proposed by @lodejard 

Please review: @lodejard @davidfowl 
cc @Tragetaschen

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
+++ b/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Threading;
 
 namespace Microsoft.AspNet.Hosting
@@ -24,5 +25,7 @@ namespace Microsoft.AspNet.Hosting
         /// </summary>
         /// <returns></returns>
         CancellationToken ApplicationStopped { get; }
+
+        Action ApplicationStarted { get; set; }
     }
 }
--- a/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
+++ b/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
@@ -26,6 +26,15 @@ namespace Microsoft.AspNet.Hosting
         /// <returns></returns>
         CancellationToken ApplicationStopped { get; }
 
-        Action ApplicationStarted { get; set; }
+        /// <summary>
+        /// Sends the notification that the application has started
+        /// </summary>
+        void ApplicationStarted();
+
+        /// <summary>
+        /// Registers a new application started action
+        /// </summary>
+        /// <param name="action">The action to register</param>
+        void OnApplicationStarted(Action action);
     }
 }
--- a/src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
+++ b/src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
@@ -35,6 +35,9 @@ namespace Microsoft.AspNet.Hosting
             get { return _stoppedSource.Token; }
         }
 
+        public Action ApplicationStarted { get; set; }
+
+
         /// <summary>
         /// Signals the ApplicationStopping event and blocks until it completes.
         /// </summary>
--- a/src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
+++ b/src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Threading;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Hosting
 {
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -54,13 +54,20 @@ public void Main(string[] args)
                 shutdownHandle.Set();
             });
 
-            var ignored = Task.Run(() =>
+            var appLifeline = host.ApplicationServices.GetRequiredService<IApplicationLifetime>();
```
Eilon: Lifeline? Lifetime!

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -54,13 +54,20 @@ public void Main(string[] args)
...
+            if (appLifeline.ApplicationStarted == null)
             {
-                Console.WriteLine("Started");
-                Console.ReadLine();
-                appShutdownService.RequestShutdown();
-            });
+                appLifeline.ApplicationStarted = ()=> {
```
Eilon: Formatting: Space after the `()`

src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
```diff
@@ -35,6 +35,9 @@ public CancellationToken ApplicationStopped
             get { return _stoppedSource.Token; }
         }
 
+        public Action ApplicationStarted { get; set; }
```
Eilon: Doc comment

src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
```diff
@@ -24,5 +25,7 @@ public interface IApplicationLifetime
         /// </summary>
         /// <returns></returns>
         CancellationToken ApplicationStopped { get; }
+
+        Action ApplicationStarted { get; set; }
```
Eilon: Doc comment

src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
```diff
@@ -35,6 +35,9 @@ public CancellationToken ApplicationStopped
...
+
+
```
muratg: Formatting: Extra empty line

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -54,13 +54,20 @@ public void Main(string[] args)
                 shutdownHandle.Set();
             });
 
-            var ignored = Task.Run(() =>
+            var appLifeline = host.ApplicationServices.GetRequiredService<IApplicationLifetime>();
+            if (appLifeline.ApplicationStarted == null)
```
Eilon: Before, we always would print out "Started", then wait, then shut down. Is it deliberate that if there's a custom AppStarted handler that we don't to anything at all?

davidfowl:
I'm sorry this API isn't good. I can't think of any example where I've seen an assignable delegate type in a property. A register method call would be better

davidfowl:
this also doesn't work if you want to disable the default behavior of hosting.

HaoK:
Maybe `appLifetime.OnApplicationStarted(() => { })?`

src/Microsoft.AspNet.Hosting/Program.cs
victorhurdugaci: Yes, by design because that's how you overwrite the default behavior. You should also be able to overwrite the message.

JamesNK:
Some of the OWIN security libraries have Actions and Funcs as properties, e.g. OpenIdConnectAuthenticationNotifications.AuthenticationFailed

davidfowl:
@JamesNK that doesn't count.

victorhurdugaci:
@davidfowl what do you think about the updated PR?

davidfowl:
A little better but the setting of the default is in the wrong place. Helios also uses this API and there should be no call to Console.Read or Console.WriteLine.

src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
```diff
@@ -24,5 +25,16 @@ public interface IApplicationLifetime
...
+        /// <summary>
+        /// Sends the notification that the application has started
+        /// </summary>
+        void ApplicationStarted();
+
+        /// <summary>
+        /// Registers a new application started action
+        /// </summary>
+        /// <param name="action">The action to register</param>
+        void OnApplicationStarted(Action action);
```
davidfowl: RegisterApplicationStarted

JamesNK:
Maybe a dumb question but why not use a C# event to add/remove startup events?

Tragetaschen:
I cannot really wrap my head around the naming here. Everything points to a semantic that is usually implemented using an event and it sounds like you would do stuff in here that is necessary right after everything started to run. But the truth is that the registered function defines the lifetime of the program.

That being said, I think there is value in an event in `IApplicationLifetime` as described.

src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
```diff
@@ -24,5 +25,16 @@ public interface IApplicationLifetime
         /// </summary>
         /// <returns></returns>
         CancellationToken ApplicationStopped { get; }
+
+        /// <summary>
+        /// Sends the notification that the application has started
```
Eilon: I know the code is still being iterated on, but remember that doc comments need to be full sentences, and thus end with periods

PR #255, opened 2015-05-05
Third way of making configurable how hosting is kept alive

Tragetaschen:
I have tried to incorporate the comments on the other two approaches (#129 and #253) to tackle the problem from yet another angle.

This now entirely relies on `IApplicationShutdown` as the method of communication and instead of going back and forth between Program and the keep-alive implementation, the latter is now responsible for doing whatever is required to call `RequestShutdown()`. For things like `Console.CancelKeyPressed`, this is the simplest event handler imaginable.

Additionally, it's now possible to register any number of `IHostingKeepAlive` implementations and any one can call `RequestShutdown()` independently.

The default implementation can be `IServiceCollection.Replace`d, but keep in mind the open issue and discussion from aspnet/DependencyInjection#155.

As a bonus, the implementation in Program.cs is now as simple as it can get.

@lodejard @davidfowl @victorhurdugaci @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Program.cs
+++ b/src/Microsoft.AspNet.Hosting/Program.cs
@@ -3,12 +3,9 @@
 
 using System;
 using System.IO;
-using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.AspNet.Hosting.Internal;
 using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.DependencyInjection;
-using Microsoft.Framework.Logging;
 using Microsoft.Framework.Runtime;
 
 namespace Microsoft.AspNet.Hosting
@@ -35,33 +32,12 @@ namespace Microsoft.AspNet.Hosting
             config.AddCommandLine(args);
 
             var host = new WebHostBuilder(_serviceProvider, config).Build();
-            var serverShutdown = host.Start();
-            var loggerFactory = host.ApplicationServices.GetRequiredService<ILoggerFactory>();
-            var appShutdownService = host.ApplicationServices.GetRequiredService<IApplicationShutdown>();
-            var shutdownHandle = new ManualResetEvent(false);
-
-            appShutdownService.ShutdownRequested.Register(() =>
+            using (host.Start())
             {
-                try
-                {
-                    serverShutdown.Dispose();
-                }
-                catch (Exception ex)
-                {
-                    var logger = loggerFactory.CreateLogger<Program>();
-                    logger.LogError("Dispose threw an exception.", ex);
-                }
-                shutdownHandle.Set();
-            });
-
-            var ignored = Task.Run(() =>
-            {
-                Console.WriteLine("Started");
-                Console.ReadLine();
-                appShutdownService.RequestShutdown();
-            });
-
-            shutdownHandle.WaitOne();
+                var appShutdownService = host.ApplicationServices.GetRequiredService<IApplicationShutdown>();
+                Console.CancelKeyPress += delegate { appShutdownService.RequestShutdown(); };
+                appShutdownService.ShutdownRequested.WaitHandle.WaitOne();
+            }
         }
     }
 }


src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -90,6 +90,7 @@ private void EnsureApplicationServices()
         {
             EnsureStartup();
 
+            _applicationServiceCollection.AddSingleton<IHostingKeepAlive, ConsoleHostingKeepAlive>();
```
davidfowl: This implementation should be registered in Program.cs. Helios (the IIS layer) also uses this library.

Tragetaschen:
I have updated my [sample code](https://gist.github.com/Tragetaschen/18e59107934333b8a3ea)

victorhurdugaci:
I don't really like the name of the interface `IHostingKeepAlive`. It implies that it is used to keep the hosting alive. It has nothing to do with that.

If the implementation doesn't call one of the termination methods, then the code will keep running forever. It would be a keep alive contract if it would block or return a cancellation token.

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -34,34 +34,19 @@ public void Main(string[] args)
...
+                var hostingKeepAlives = host.ApplicationServices.GetRequiredServices<IHostingKeepAlive>();
 
-            var ignored = Task.Run(() =>
-            {
-                Console.WriteLine("Started");
-                Console.ReadLine();
-                appShutdownService.RequestShutdown();
-            });
+                foreach (var hostingKeepAlive in hostingKeepAlives)
+                    hostingKeepAlive.Setup(appShutdownService);
```
victorhurdugaci: 1. Coding style: always use `{ }`
2. This approach doesn't allow me to overwrite the default behavior and that's a critical scenario for the Nano server

HaoK: Well we actually do have this pattern already today with multiservices, its not great but you can override the default behavior by removing the IHostingKeepAliveService and adding your own.

Tragetaschen:
`IShutdownMethod`?

src/Microsoft.AspNet.Hosting.Abstractions/IHostingKeepAlive.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Provide a way to keep the application running
+    /// </summary>
+    public interface IHostingKeepAlive
+    {
+        /// <summary>
+        /// Setup a way to signal application shutdown
+        /// </summary>
+        /// <param name="applicationShutdown">The application shutdown</param>
+        void Setup(IApplicationShutdown applicationShutdown);
```
HaoK: Perhaps Register would be a better name than Setup?

Tragetaschen:
`IShutdownMethod`

src/Microsoft.AspNet.Hosting/Program.cs
Tragetaschen: That's what I meant by `IServiceCollection.Replace`. There is an extension method for that in DependencyInjection.

HaoK:
`IApplicationShutdownHandler` /  `Handle` (instead of Setup)?

victorhurdugaci:
Isn't it weird that the "shutdown" handler writes "started"? :smile:

Tragetaschen:
Hehe. See my samples for some better texts. It should be something along the lines of `To terminate, press enter` or similar.

HaoK:
IShutdownManager might not be bad either, since its his job to request shutdown...

Tragetaschen:
Well, then all the way: `IShutdownRequester`. "Manager" is such an overloaded term…

HaoK:
Ok, then my vote is back to `IShutdownHandler` / `Handle(IApplicationShutdown)`

Tragetaschen:
Huh, cla-required?

Eilon:
@Tragetaschen We just changed the repos from MS Open Tech to .NET Foundation so there's a different CLA. The new CLA is simpler than the old one and so should only take a few moments to re-sign. Apologies for the extra bit of work!

Tragetaschen:
I have submitted the .NET Foundation's CLA, but @dnfclas doesn't recognize it yet. Maybe I did something wrong? @danroth27

Tragetaschen:
Back to topic: Yes, indeed, `IShutdownHandler` concisely covers the the intention of the interface.

However, I'm having a harder time with the method name. I tried to write the API documentation when the name was `Handle`, but couldn't come up with a sentence that made it the logical conclusion. `Setup` on the other hand strongly points to the do-something-and-return mentality of the method.

HaoK:
Perhaps something like: "This method is responsible for handling shutdown and eventually calling RequestShutdown."

davidfowl:
After thinking through this a bit I'd (and after a discussion with @victorhurdugaci), I'm wondering why we need any of this at all? We can just always wait on the reset event and nothing else. If you want to setup something then just do it in your Configure method. Hook whatever it is you need and then call RequestShutdown at the appropriate time.

Tragetaschen:
And an `app.UseDefaultShutdown()` to get the the `Console.ReadLine()` behavior?

davidfowl:
Nah just cut it altogether.

davidfowl:
We were going to do that anyways in favor of ctrl+c

Tragetaschen:
You won't see me shed tears :smile:

Tragetaschen:
And again an [update to the sample](https://gist.github.com/Tragetaschen/18e59107934333b8a3ea)

victorhurdugaci:
@davidfowl I spoke to @lodejard and he thinks that this will not work because one might want to override the default behavior for Ctrl+C

Eilon:
@Tragetaschen regarding the CLA did you get an email from DocuSign, and then did you do the digital signature that it requests? If you've done that and it's not working let me know and I'll contact the CLA people.

Tragetaschen:
You can always register an event handler on `Console.CancelKeyPress` to override the default behavior for Ctrl+C. I'm using a variant that maps Ctrl+C to graceful application shutdown:

```
public static void OnCancelKeyPress(this IApplicationShutdown applicationShutdown)
{
    Console.WriteLine("To terminate, press CTRL+C");
    Console.CancelKeyPress += delegate
    {
        applicationShutdown.RequestShutdown();
        Console.WriteLine("Received cancel key press");
    };
}
```

Of course, the event handler can do anything and Hosting doesn't put any requirement on that.

Maybe, I'm missing some information what this has to do with this PR :smile:

victorhurdugaci:
We want a default behavior that allows the server to shutdown gracefully, but that behavior has to be override-able.

davidfowl:
I actually think is the best approach. What happens when you do ctrl + c right now?

Tragetaschen:
This ends up calling the equivalent of `Environment.Exit` on both Mono and Windows and runs the Finalizers, not the `Dispose` logic. Btw, it's that finalizer code path that currently blocks Kestrel when doing CTRL+C.

davidfowl:
@Tragetaschen so as it is right now, things will hang for kestrel without the graceful unwind? We could just add the control c handling then.

Tragetaschen:
The main problem with `Console.ReadLine` is, that it requires a working TTY on Unix. That's the reason why this specific code needs to be removed from Hosting, so a Docker image or a systemd service can remove this superfluous requirement.

On the other hand, I don't think it's necessary to overriding/replace a default CTRL+C handler when Hosting registers it. You would only need this, if you don't want the process to shutdown and that's quite counter-intuitive when actually running on a terminal.
All other requirements can be easily wired against the members of `IApplicationLifetime`.

(dang, you beat me with your comment)

davidfowl:
@Tragetaschen add that to your change

Tragetaschen:
The `Console.CancelKeyPress` hasn't landed in .NET Core (System.Console 4.0.0-beta-22906) yet. I've if'ed that part and left a comment

davidfowl:
@ellismg @ericstj Why don't we have that change yet in our builds? Has it made it in?

ericstj:
Looks like @ellismg didn't update the reference assembly.

ericstj:
I made a fix to the reference assembly, you should see it in the next build.

Tragetaschen:
4.0.0-beta-22910 has landed and includes `CancelKeyPress`

:clap: :smile:

victorhurdugaci:
How do you overwrite the Ctrl+C behavior?

Tragetaschen:
Why would you want to do that?

As discussed above, when not running in a console environment, the registered handler is never called and you can still implement any method you want to call `IApplicationShutdown.RequestShutdown()`.

If you are running in a console environment, that's an interactive environment and properly reacting to CTRL+C like here is a genuinely good idea.

Still, I might be missing something, but right now I think a good answer to your question is: You don't need to.

PR #261, opened 2015-05-08
Add an ApplicationStarted event

Tragetaschen:
The discussion in #253 and my own experience show, that it would be great to have a notification once everything is up and running.

The canonical use case is the `Console.WriteLine("Started")` for console users. My own requirement is a notification for systemd that the process has completely finished starting up and that it's safe to start the web browser on my embedded device using [libsystemd](http://www.freedesktop.org/software/systemd/man/sd_notify.html):

```C#
public static class SystemdStartupNotifier
{
    public static void NotifySystemd(this IApplicationLifetime applicationLifetime)
    {
        applicationLifetime.ApplicationStarted.Register(handler);
    }

    private static void handler()
    {
        sd_notify(0, "READY=1");
    }

    [DllImport("libsystemd.so.0")]
    private static extern int sd_notify(int unset_environment, string state);
}
```

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
+++ b/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
@@ -11,6 +11,12 @@ namespace Microsoft.AspNet.Hosting
     public interface IApplicationLifetime
     {
         /// <summary>
+        /// Triggered when the application host has fully started and is about to wait
+        /// for a graceful shutdown.
+        /// </summary>
+        CancellationToken ApplicationStarted { get; }
+
+        /// <summary>
         /// Triggered when the application host is performing a graceful shutdown.
         /// Request may still be in flight. Shutdown will block until this event completes.
         /// </summary>
--- a/src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
+++ b/src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
@@ -11,8 +11,18 @@ namespace Microsoft.AspNet.Hosting
     /// </summary>
     public class ApplicationLifetime : IApplicationLifetime
     {
+        private readonly CancellationTokenSource _startedSource = new CancellationTokenSource();
         private readonly CancellationTokenSource _stoppingSource = new CancellationTokenSource();
         private readonly CancellationTokenSource _stoppedSource = new CancellationTokenSource();
+
+        /// <summary>
+        /// Triggered when the application host has fully started and is about to wait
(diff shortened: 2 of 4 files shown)


davidfowl:
Nice

davidfowl:
Write a test

Tragetaschen:
Naturally

test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
```diff
@@ -91,6 +91,20 @@ public void HostingEngineCanBeStarted()
...
+        public void HostingEngineNotifiesApplicationStarted()
+        {
+            var host = CreateBuilder()
+                .UseServer(this)
+                .Build();
+            var applicationLifetime = host.ApplicationServices.GetRequiredService<IApplicationLifetime>();
+
+            var applicationStartedCalled = false;
+            applicationLifetime.ApplicationStarted += delegate { applicationStartedCalled = true; };
+            using (host.Start())
```
HaoK: nit: required to always have { } (I wish it was otherwise myself...)

src/Microsoft.AspNet.Hosting/ApplicationLifetime.cs
```diff
@@ -15,6 +15,12 @@ public class ApplicationLifetime : IApplicationLifetime
         private readonly CancellationTokenSource _stoppedSource = new CancellationTokenSource();
```
davidfowl: Maybe we should use Events all over /cc @lodejard @Tratcher

src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
```diff
@@ -11,6 +12,12 @@ namespace Microsoft.AspNet.Hosting
     public interface IApplicationLifetime
     {
         /// <summary>
+        /// Triggered when the application host has fully started and is about to wait
+        /// for a graceful shutdown.
+        /// </summary>
+        event EventHandler ApplicationStarted;
```
davidfowl: The benefit of cancellation tokens is that you can both poll it an handle an event.

davidfowl:
For the sake of consistency, it should be a cancellation token. We can have another discussion if we decide to change each of the things to a bool + event, or stick with CT

Tragetaschen:
I like the CT, even more since I now know the good reasons

davidfowl:
LGTM

Tratcher:
Should be fine. We haven't even implemented this interface in Helios yet.

PR #264, opened 2015-05-10
Provide access to IApplicationShutdown's RequestShutdown through IApplicationLifetime

Tragetaschen:
Currently you have to work with two interfaces if you want to control application lifetime entirely. This change extends `IApplicationLifetime` with the ability to initiate application shutdown by forwarding it to `IApplicationShutdown`, so consumers of Hosting need only one interface for application lifetime purposes.

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
+++ b/src/Microsoft.AspNet.Hosting.Abstractions/IApplicationLifetime.cs
@@ -17,6 +17,11 @@ namespace Microsoft.AspNet.Hosting
         CancellationToken ApplicationStarted { get; }
 
         /// <summary>
+        /// Terminates the application
+        /// </summary>
+        void StopApplication();
+
+        /// <summary>
         /// Triggered when the application host is performing a graceful shutdown.
         /// Request may still be in flight. Shutdown will block until this event completes.
         /// </summary>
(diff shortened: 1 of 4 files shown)


Tragetaschen:
The `IApplicationShutdown.ShutdownRequested` and `IApplicationLifetime.ApplicationStopping` CTs denote the same point in time—there are only a couple of instructions between them. There is no need to have the former in `IApplicationLifetime`.

davidfowl:
I'm not sure about this change. I get that it's convenient to have a single interface but triggering was left off this interface so that there's a single way to do it for both console and web apps.

Tragetaschen:
I understand.

My driving force behind this is integration with Linux' systemd. For that, I currently have to call two extension methods on the interfaces to have everything set up correctly:
```
applicationLifetime.NotifySystemd();
applicationShutdown.OnUnixSignals();
```
The first uses the started-event to ping back to systemd, the latter installs additional signal handlers for shutdown. With this change, I could merge them into one `applicationLifetime.UseSystemd()`.

davidfowl:
@Tragetaschen do you have that implementation anywhere handy? That would be a neat component to look at. I'd imagine lots of people will want to do the same.

Tragetaschen:
It's handy now in this [gist](https://gist.github.com/Tragetaschen/435c5bb2062c29753070)

davidfowl:
@Tragetaschen closing this out

davidfowl:
@Tragetaschen See https://github.com/aspnet/dnx/issues/2683

Tragetaschen:
Already had: Yay!

PR #270, opened 2015-05-13
Use a json file instead of an ini file to configure hosting environment

Matthew-Bonner:


src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.AspNet.Hosting
 {
     public class Program
     {
-        private const string HostingIniFile = "Microsoft.AspNet.Hosting.ini";
+        private const string HostingJsonFile = "Microsoft.AspNet.Hosting.json";
```
HaoK: Perhaps it would be better to simply support both?  Prefer json first, if not present, look for ini?

Matthew-Bonner:
I was under the impression we were trying to keep things simple so supporting one language opposed to multiple was the way forward?

src/Microsoft.AspNet.Hosting/Program.cs
Matthew-Bonner: What is the point? The whole idea of switching to JSON is to make this consistent with where everything else is headed, so we might as well ditch the ini file now so that we can give people time to update their projects and then we can work on more pressing issues.

davidfowl: Yea we're just going to use json forget the ini file.

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -24,9 +24,9 @@ public Program(IServiceProvider serviceProvider)
         public void Main(string[] args)
         {
             var config = new Configuration();
-            if (File.Exists(HostingIniFile))
+            if (File.Exists(HostingJsonFile))
             {
-                config.AddIniFile(HostingIniFile);
+                config.AddJsonFile(HostingJsonFile);
```
Tratcher: Do you need to pull in the config json dependency for this? I don't think we want to do that in Hosting as it brings in Newtonsoft.json.

Michael-DST:
I would like to vote for both .json & .ini.  You are able to configure ASP.NET in any format you want as long as there is a provider, correct?  Why not make it the same here?

src/Microsoft.AspNet.Hosting/Program.cs
davidfowl: Already the case

muratg:
@Matthew-Bonner Could you rebase the PR?  

@Tratcher Could you take this in?

Tratcher:
@Matthew-Bonner don't worry about it, I'll take care of it.

Tratcher:
Rebased, updated, and merged.

Matthew-Bonner:
Thanks guys :+1:

Issue #252, opened 2015-04-30
The parameters of the configure services delegate are not injected


Tragetaschen:
All parameters [but `IServiceCollection`](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Startup/ConfigureServicesDelegate.cs#L33) are `null`

davidfowl:
That's by design. We don't want to resolve anything from the service collection until it is built.

Tragetaschen:
But constructor arguments for `Startup` are injected already (which is my workaround now).

If this is the case, then there should be a better error message.

davidfowl:
+1 on the error message. This just hit me in omnisharp.

HaoK:
Yeah seems like we can just always throw if the ConfigureServices isn't returning void or IServiceProvider, and its taking anything other than a single IServiceCollection parameter or no parameters...

Tragetaschen:
I'm not so convinced that only taking `IServiceCollection` as a parameter is the right thing, though. When I'm implementing `Configure` I can get dependencies like `IHostingEnvironment` or `ILoggerFactory` (see the default templates in VS), but that's not possible for `ConfigureServices`. This seems odd.

davidfowl:
@Tragetaschen ConfigureServices is setting up your dependencies. It's the last ditch effort to provide any more dependencies before the container is built. Configure runs after the container has been built and has access to everything.

Tragetaschen:
I'm well aware, but try to look at how a startup class works: I can get dependencies in the constructor (that's the first opportunity) and I can get dependencies in `Configure` (that's the last opportunity), but I cannot get any in `Configure*Services`. That's why I think it's odd.

HaoK:
We actually removed this functionally as a simplification/cleanup to avoid passing the host service provider to the startup loader in the last hosting iteration I think. 

A workaround would be to just grab the dependencies in the startup ctor and store them for later use in configure services. A further subtlety is that the startup dependencies are actually from a different (hosting) container than the configure dependencies (application) container.

HaoK:
5b851c49a337e49001ed2351253ab0f52b696040

PR #271, opened 2015-05-14
Throw if ConfigureServices takes more than IServiceCollection

HaoK:
Fixes https://github.com/aspnet/Hosting/issues/252

cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Startup/ConfigureDelegate.cs
+++ b/src/Microsoft.AspNet.Hosting/Startup/ConfigureDelegate.cs
@@ -20,10 +20,7 @@ namespace Microsoft.AspNet.Hosting.Startup
 
         public MethodInfo MethodInfo { get; }
 
-        public Action<IApplicationBuilder> Build(object instance)
-        {
-            return builder => Invoke(instance, builder);
-        }
+        public Action<IApplicationBuilder> Build(object instance) => builder => Invoke(instance, builder);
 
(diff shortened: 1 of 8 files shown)


Tragetaschen:
Building the `parameters` array in the `Invoke` method now can be simplified.

HaoK:
Good idea @Tragetaschen!  updated PR

src/Microsoft.AspNet.Hosting/Startup/ConfigureDelegate.cs
```diff
@@ -20,10 +20,7 @@ public ConfigureBuilder(MethodInfo configure)
 
         public MethodInfo MethodInfo { get; }
 
-        public Action<IApplicationBuilder> Build(object instance)
-        {
-            return builder => Invoke(instance, builder);
-        }
+        public Action<IApplicationBuilder> Build(object instance) => builder => Invoke(instance, builder);
```
Tratcher: wow, where did you find this syntax?

HaoK: C# 6.0 ftw, yeah its pretty snazzy looking since it gets rid of my personal pet peeve, the bloated naked { } lines for one liners :)

Tratcher:
:shipit:

HaoK:
5b851c49a337e49001ed2351253ab0f52b696040

PR #274, opened 2015-05-17
Remove ConfigureServicesDelegate

davidfowl:
- Removed the delegate type as it's not used anywhere in user code
- Also removed platform helper


Tragetaschen:
LGTM

PR #275, opened 2015-05-18
Shift 'Development' & 'Production' constants into 'EnvironmentName' class

hishamco:


src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Hosting
+{
+    public class EnvironmentName
```
HaoK: Make this a static class?

hishamco: I already did that before, but while ```Development``` and ```Production``` are constants they are accessible without static class

HaoK: My point was just that if it never make sense to call `new EnvironmentName()`?  Why not make it a static class?

hishamco: :+1:

hishamco:
I updated the PR to make the ```EnvironemtName``` as static class.
thanks @HaoK,

src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Hosting
+{
+    public static class EnvironmentName
```
davidfowl: Should it be EnvironmentNames

src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        public const string Development = "Development";
+        public const string Production = "Production";
```
davidfowl: Should this be public static readonly /cc @Eilon

hishamco: @davidfowl I made them static readonly before I made a PR, nothing but there's no possibility to change the values in the construct, so I think the constant will make sense

src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
Eilon: I think singular is maybe more correct, no? Think of this as a "Mode." Would it be `AppMode { ModeA, ModeB, ModeC }` or `AppModes { ModeA, ModeB, ModeC }`? When you set it, is it `CurrentMode = AppMode.ModeA;` or `CurrentMode = AppModes.ModeA;`

src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
Eilon: Great question. Part of me says to make it `static readonly` to avoid issues if we change these constants while we're in beta... though I admit that's a pretty weak reason. One unique benefit of having a `const` is that you can use it in attributes. Though we don't have any such attributes ourselves, it's conceivable that someone else might have such an attribute, e.g. `[StartupSelector(EnvironmentName.Production)]` or something.

But, having said all that, using `static readonly` is certainly "safer." Our general rule really is that `const` should be used for things that actually *can't* change, such as mathematical truths, or other intrinsic concepts (e.g. `const int MillisecondsPerSecond = 1000`).

src/Microsoft.AspNet.Hosting/HostingEnvironment.cs
```diff
@@ -8,12 +8,10 @@ namespace Microsoft.AspNet.Hosting
...
-        public string EnvironmentName { get; set; } = DefaultEnvironmentName;
+        public string EnvironmentName { get; set; } = Microsoft.AspNet.Hosting.EnvironmentName.Production;
 
         public string WebRootPath { get; set; }
 
         public IFileProvider WebRootFileProvider { get; set; }
     }
-}
\ No newline at end of file
+}
```
Eilon: Remove trailing spaces! :smile:

src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
hishamco: :+1: @Eilon

src/Microsoft.AspNet.Hosting/HostingEnvironment.cs
hishamco: Thanks @Eilon

hishamco:
@Eilon I updated the PR according to the changed above

src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
Tragetaschen: IMHO when looking at the code consuming the members, it looks like the declaration could also be an `enum`. So the guideline regarding (non-)`[Flags]` could be applied which makes singular the right choice.

hishamco: @Tragetaschen I already implemented as ```enum``` before but when I revise the entire repo I notice that environment name may comes from configuration as string, so that's why I used ```class enum```

Eilon: Yeah I don't think an enum is the best option here for the reason @hishamco mentions. These really are fundamentally strings.

Tragetaschen: I know that. I was giving a (hopefully) good reason for the singular/plural decision...

Eilon: Ah, understood!

hishamco:
@Tratcher what is the status of this PR

Eilon:
@hishamco we're doing some stabilization right now, but also we have a 3-day weekend here so we will merge it next week. Thanks!

hishamco:
@Eilon can you take an action about this PR

Eilon:
Whoops we forgot about this, sorry!

@hishamco can you rebase on the latest bits and we can merge?

@Tratcher can you keep an eye on this?

hishamco:
@Eilon I rebase on the latest bits
/cc @Tratcher

Eilon:
@hishamco thanks.

@Tratcher this could go into beta7 or beta8, whichever you get around to.

hishamco:
Thanks @davidfowl

Issue #277, opened 2015-05-20
Hosting is not able to load Microsoft.AspNet.Hosting.ini file


rustd:
I have a sample which has the following in web command and hosting file. This does not work today.
```C#
"web": "Microsoft.AspNet.Hosting",
```
and the following in Microsoft.AspNet.Hosting.ini
```C#
[Microsoft.AspNet.Hosting]
server=Microsoft.AspNet.Server.WebListener
server-urls=http://localhost:5000

```

```C#
System.InvalidOperationException: Unable to resolve path 'Microsoft.AspNet.Hosting.ini'; construct
his IConfigurationSourceRoot with a non-null BasePath.
   at Microsoft.Framework.ConfigurationModel.Helper.ConfigurationHelper.ResolveConfigurationFilePat
(IConfigurationSourceRoot configuration, String path)
   at Microsoft.Framework.ConfigurationModel.ConfigurationExtensions.AddIniFile(IConfigurationSourc
Root configuration, String path, Boolean optional)
   at Microsoft.AspNet.Hosting.Program.Main(String[] args)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Framework.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] arg
, IServiceProvider serviceProvider)
   at Microsoft.Framework.ApplicationHost.Program.ExecuteMain(DefaultHost host, String applicationN
me, String[] args)
   at Microsoft.Framework.ApplicationHost.Program.Main(String[] args)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Framework.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] arg
, IServiceProvider serviceProvider)
   at dnx.host.Bootstrapper.RunAsync(List`1 args)
   at dnx.host.RuntimeBootstrapper.ExecuteAsync(String[] args)
   at dnx.host.RuntimeBootstrapper.Execute(String[] args)
Press any key to continue . . .
```

rustd:
@muratg this is for Beta5 for the templates

Tratcher:
FYI: The expected key names are:
```
server=Microsoft.AspNet.Server.WebListener
server.urls=http://localhost:8080
```
or
```
[...]

Issue #276, opened 2015-05-20
Hosting should take in hosting file name as an argument


rustd:
This would allow us to do the following.
```C#
    "web": "Microsoft.AspNet.Hosting --config hosting.ini",
```
The code that needs to be changed is https://github.com/aspnet/Hosting/blob/5b851c49a337e49001ed2351253ab0f52b696040/src/Microsoft.AspNet.Hosting/Program.cs

As a stretch it would be awesome if the hosting could read a json file so we could have hosting.json

rustd:
@muratg @Eilon This is for Beta5 from the templates review

muratg:
https://github.com/aspnet/Hosting/issues/277

davidfowl:
@lodejard

muratg:
@rustd Template change for the web command should be unblocked now.

PR #280, opened 2015-05-20
Fix ini loading, read location from command line.

Tratcher:
#277 Fix loading the hosting ini file, requires a base path
#276 Allow the ini file to be specified via command line '--config'
@davidfowl @muratg @rustd


muratg:
:shipit:

PR #285, opened 2015-05-28
Fix build break

kichalla:
@Tratcher


Tratcher:
:shipit: after updating the other deployers.

PR #294, opened 2015-06-16
Inline RequestServicesContainer

HaoK:
cc @davidfowl @Tratcher Should we go as far as inlining the entire request services middleware itself inside of the AutoServicesFilter?

Also it looks like today we pop ApplicationServices onto the RequestServices instead of returning it to null.  Is that what we actually want?


src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -21,9 +23,38 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
...
+
+            var serviceProvider = httpContext.ApplicationServices ?? _services;
+
+            var appServiceProvider = serviceProvider.GetRequiredService<IServiceProvider>();
+            var appServiceScopeFactory = serviceProvider.GetRequiredService<IServiceScopeFactory>();
+
+            if (serviceProvider != appServiceProvider)
+            {
+                appServiceProvider = serviceProvider;
+                appServiceScopeFactory = serviceProvider.GetRequiredService<IServiceScopeFactory>();
```
Tratcher: Does appServiceScopeFactory need to be resolved twice? Could you just resolve it once after setting appServiceProvider?

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -21,9 +23,38 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
...
+                {
+                    httpContext.ApplicationServices = appServiceProvider;
+                    httpContext.RequestServices = scope.ServiceProvider;
+
+                    await _next.Invoke(httpContext);
+                }
+            }
+            finally
+            {
+                httpContext.RequestServices = serviceProvider; // REVIEW: should this go back to null instead?
```
Tratcher: Maybe return it to its original value?

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
HaoK: Yeah and that makes this easier to read as well good idea

HaoK: updated

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
HaoK: The old request services value is always null since we no-op otherwise.  I think this was copied from code that was meant to be general purpose, and we no longer are general purpose, so I think this should just be null?

Tratcher: Ah, sure, that makes sense.

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -21,9 +23,37 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
...
-            using (var container = RequestServicesContainer.EnsureRequestServices(httpContext, _services))
+            // All done if we already have a request services
+            if (httpContext.RequestServices != null)
             {
-                await _next.Invoke(httpContext);
+                return;
+            }
+
+            // Resolve the ScopeFactory from the correct SP
+            var priorApplicationServices = httpContext.ApplicationServices;
```
HaoK: Tweaked the logic a bit to return the applicationServices to httpContext's ApplicationServices which may be null (instead of replacing it with _services)

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -21,9 +23,37 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
...
+                    httpContext.ApplicationServices = appServiceProvider;
+                    httpContext.RequestServices = scope.ServiceProvider;
+
+                    await _next.Invoke(httpContext);
+                }
+            }
+            finally
+            {
+                httpContext.RequestServices = null;
+                httpContext.ApplicationServices = priorApplicationServices;
```
HaoK: So now we return RequestServices to null, and ApplicationServices to the old context.ApplicationServices, instead of returning it to the effective serviceProvider which we were doing previously (I don't think this behavior actually matters to anyone except during cleanup (or if someone wraps our IStartupFilter with their own which adds middleware before us)

src/Microsoft.AspNet.Hosting/Internal/AutoRequestServicesStartupFilter.cs
```diff
@@ -13,6 +13,7 @@ public Action<IApplicationBuilder> Configure(IApplicationBuilder app, Action<IAp
         {
             return builder =>
             {
+                // REVIEW: we could inline the entire request services container middleware here instead?
```
Tratcher: No, I think it's cleaner this way.

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -21,9 +23,37 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
...
+            if (httpContext.RequestServices != null)
             {
-                await _next.Invoke(httpContext);
+                return;
+            }
+
+            // Resolve the ScopeFactory from the correct SP
+            var priorApplicationServices = httpContext.ApplicationServices;
+            var serviceProvider = priorApplicationServices ?? _services;
+            var appServiceProvider = serviceProvider.GetRequiredService<IServiceProvider>();
```
davidfowl: This logic looks so bizzare

davidfowl: Why not something like:

```
var appServiceProvider  = httpContext.ApplicationServices ?? _services;
```

Why do we call GetService on the IServiceProvider? Doesn't it just return itself?

HaoK: Hrm, this code has been moved around so often now, let me see if I can find trace the history, but is it guaranteed always that `GetService<IServiceProvider>` returns itself?  So in the code above appServiceProvider == serviceProvider always?

HaoK:
Updated cleaning up the ISP resolution, now we just use either httpContext.ApplicationServices or _services.  @Tratcher @davidfowl can you guys do a final review?

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -13,18 +15,40 @@ public class RequestServicesContainerMiddleware
...
 
-        public async Task Invoke(HttpContext httpContext)
+        public async Task Invoke([NotNull] HttpContext httpContext)
         {
-            using (var container = RequestServicesContainer.EnsureRequestServices(httpContext, _services))
+            // All done if there request services is set
+            if (httpContext.RequestServices != null)
             {
-                await _next.Invoke(httpContext);
+                return;
```
Tratcher: Why was `await _next.Invoke(httpContext);` removed here?

Tratcher:
:shipit: after fixing the no-op code path.

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
HaoK: Oops yeah, I should add a test that exercises the request services already being populated and not used

HaoK:
Fixed no-op and added a test that actually adds another request filter and stuffs a request services in

HaoK:
ee8baab1ed725cbce97e8daccd5ee7dbacad51b5

Issue #214, opened 2015-03-29
Simplify ServerLoader logic


HaoK:
Per @lodejard the logic should be simplified to:
- serverFactoryIdentifier should be treated only as the assembly name and directly used via Assembly.Load, the assembly will be scanned for a IServerFactory (no more split of name, and no more support for the serverFactoryIdentifier being a type)

davidfowl:
Nice

Tratcher:
@HaoK Did this get done? If not please move it to beta6.

HaoK:
Not yet I will take a look today/next week

HaoK:
bf6e293bfedcb921215f8c4273839912be205719

PR #295, opened 2015-06-16
Cleanup ServerLoader logic

HaoK:
Fixes https://github.com/aspnet/Hosting/issues/214

Note looks like the old code was doing a weird (finding all IServerLoaders and the last one wins).  Now we just select the first one (I'm not sure that edge case matters, but we could throw possibly instead and use SingleOrDefault?)

cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
+++ b/src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNet.Hosting.Server
                 throw new InvalidOperationException($"No server type found that implements IServerFactory in assembly: {assemblyName}.");
             }
 
-            return (IServerFactory)ActivatorUtilities.GetServiceOrCreateInstance(_services, serverType);
+            return (IServerFactory)ActivatorUtilities.GetServiceOrCreateInstance(_services, serverType.AsType());
         }
     }
 }
--- a/src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
+++ b/src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
@@ -4,8 +4,8 @@
 using System;
 using System.Linq;
 using System.Reflection;
-using Microsoft.AspNet.Hosting.Internal;
 using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Internal;
 
 namespace Microsoft.AspNet.Hosting.Server
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
```diff
@@ -18,59 +18,29 @@ public ServerLoader(IServiceProvider services)
...
-                foreach (var typeInfo in assembly.DefinedTypes)
-                {
-                    interfaceInfo = typeInfo.ImplementedInterfaces.FirstOrDefault(interf => interf.Equals(typeof(IServerFactory)));
-                    if (interfaceInfo != null)
-                    {
-                        type = typeInfo.AsType();
-                    }
-                }
+            var serverType = assembly.DefinedTypes.Where(
+                t => t.ImplementedInterfaces.FirstOrDefault(interf => interf.Equals(typeof(IServerFactory))) != null)
```
hishamco: Can't convert ```TypeInfo``` to ```System.Type```

HaoK: Ah thanks, I don't think we use this entry point very often, I'm not sure we have any unit test coverage at all currently of this...  I'll see if I can find someplace to add one for this

HaoK: Actually TypeInfo is a Type, so I'm not sure AsType is actually needed

hishamco: Or you can modify the query like this
```csharp
            var serverType = assembly.DefinedTypes.Where(ti=>ti.ImplementedInterfaces.Contains(typeof(IServerFactory))).Select(t=>t.AsType()).FirstOrDefault();
```

src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
```diff
@@ -18,59 +18,29 @@ public ServerLoader(IServiceProvider services)
...
         }
 
-        public IServerFactory LoadServerFactory(string serverFactoryIdentifier)
+        public IServerFactory LoadServerFactory([NotNull] string assemblyName)
         {
-            if (string.IsNullOrEmpty(serverFactoryIdentifier))
+            if (string.IsNullOrEmpty(assemblyName))
             {
-                throw new ArgumentException(string.Empty, nameof(serverFactoryIdentifier));
+                throw new ArgumentException(string.Empty, nameof(assemblyName));
```
henkmollema: Do we still need this or does `[NotNull]` not check for empty strings?

HaoK: Both, NotNull doesn't yet enforce anything at runtime, and no it won't prevent empty strings

src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
HaoK: Looks like the ServerTests repo exercise this code path today (WebListener as well)

HaoK:
@Tratcher can you review?

src/Microsoft.AspNet.Hosting/Server/ServerLoader.cs
```diff
@@ -18,59 +18,29 @@ public ServerLoader(IServiceProvider services)
             _services = services;
         }
 
-        public IServerFactory LoadServerFactory(string serverFactoryIdentifier)
+        public IServerFactory LoadServerFactory([NotNull] string assemblyName)
```
Tratcher: Methods implementing interfaces should use the same parameter names as the interface. `serverName` in this case.

Tratcher:
:shipit:

HaoK:
bf6e293bfedcb921215f8c4273839912be205719

Issue #296, opened 2015-06-17
IApplicationLifetime ApplicationStopping does not trigger when you press ctl+c from the command line


jsacapdev:
Possibly related to https://github.com/aspnet/Hosting/issues/151.

In the following code:

```
using Microsoft.AspNet.Builder;
using Microsoft.AspNet.Hosting;
using Microsoft.Framework.DependencyInjection;
using System;

namespace TestAppShutdownWebApi
{
    public class Startup
    {
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvc();
        }

        public void Configure(IApplicationBuilder app, IApplicationLifetime lifetime)
        {
            lifetime.ApplicationStopping.Register(OnStopping);

            app.UseMvc();
        }

        public void OnStopping()
        {
            Console.WriteLine("OnStopping");
        }
    }
}
```

...the OnStopping event is not raised.

The following code works.

```
using Microsoft.AspNet.Builder;
using Microsoft.AspNet.Hosting;
using Microsoft.AspNet.Http;
using Microsoft.Framework.DependencyInjection;
using System;

namespace TestAppShutdownWebApi
{
    public class Startup
    {
        public void ConfigureServices(IServiceCollection services)
        {
        }

        public void Configure(IApplicationBuilder app, IApplicationLifetime lifetime)
        {
            lifetime.ApplicationStopping.Register(OnStopping);
        }

        public void OnStopping()
        {
            Console.WriteLine("OnStopping");
        }
    }
}
```

This is the case for Microsoft.AspNet.Server.WebListener or Kestrel.

davidfowl:
What version of hosting are you using?

davidfowl:
Also, you're talking about ctrl + c not triggering stopping right?

jsacapdev:
Microsoft.AspNet.Hosting (1.0.0-beta6-11996)

Yes, CTRL+C does not trigger the stop.

Tratcher:
Does the application exit as expected or just hang?

This may be related to https://github.com/aspnet/WebListener/issues/71 (private)

jsacapdev:
exits

jsacapdev:
At the command line it does not seem to work at all (even without the MVC registration). Using 1.0.0-beta6-12085 coreclr x64, at the command line:
```
C:\Tmp\TestAppShutdownWebApi\src\TestAppShutdownWebApi>dnx . kestrel
Loaded module: dnx.win32.coreclr.dll
[...]

PR #297, opened 2015-06-19
Keep the process alive long enough to shutdown gracefully.

Tratcher:
#296 
Ctl+c from the command line kills the process as soon as the Console.CancelKeyPress event finishes, it doesn't wait for the Main method/thread to exit.
@muratg @davidfowl @jsacapdev @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Program.cs
+++ b/src/Microsoft.AspNet.Hosting/Program.cs
@@ -22,7 +22,7 @@ namespace Microsoft.AspNet.Hosting
 
         public void Main(string[] args)
         {
-            // Allow the location of the ini file to be specfied via a --config command line arg
+            // Allow the location of the ini file to be specified via a --config command line arg
             var tempBuilder = new ConfigurationBuilder().AddCommandLine(args);
             var tempConfig = tempBuilder.Build();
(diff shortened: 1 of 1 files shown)


Tratcher:
https://msdn.microsoft.com/en-us/library/system.console.cancelkeypress(v=vs.110).aspx

Tragetaschen:
LGTM. When I saw the bug, that was my first idea.

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -39,7 +39,12 @@ public void Main(string[] args)
             {
                 Console.WriteLine("Started");
                 var appShutdownService = host.ApplicationServices.GetRequiredService<IApplicationShutdown>();
-                Console.CancelKeyPress += delegate { appShutdownService.RequestShutdown(); };
+                Console.CancelKeyPress += (sender, eventArgs) =>
+                {
+                    appShutdownService.RequestShutdown();
+                    // Don't terminate the process immidately, wait for the Main thread to exit gracefully.
```
muratg: nit: immediately

muratg:
:shipit:

Issue #254, opened 2015-05-01
Search in more locations for Startup class


Matthew-Bonner:
We need the ability to specify the Startup type we want to use as in our case the Startup type we want to use resides in a referenced dependency project.

Our preferred solution is to extend the FindStartupType method so that just before the type is returned, a search of all assemblies is performed to look for a startup type, we think that this solution will be of the most benefit to a wider community.

A search of Startup types could be cached during initial compilation in a similar way to how the ControllerTypeCache performs a search for controllers and caches the result in the MVC stack.

**Use case scenario**
A sample use case scenario is a solution which takes on the following structure:
Project 1: Library of classes

Project 2: Scaffold of website
Depends on Project 1
(This will contain a Startup class)

Project 3: Website of models, views and controllers
Depends on Project 1 and Project 2

Project 4: Scaffold of administration website
Depends on Project 1 and Project 2
(This will contain a Startup class)

Project 5: Administration website for Project 3
Depends on Project 1, Project 2 and Project 3

This structure allows for the scaffold of a website or administration website to be reused by multiple websites and administration websites over and over again allowing for a rapid development while allowing for changes to be made to the scaffold which will be applied to the websites using the scaffolds.

In the scenario of Project 4 depending on Project 2 which already contains a Startup class, because Project 4 contains a Startup class the Startup class in Project 2 is not loaded.

**Code example**

*Project 2*
~/Startup.cs

```cs
using Microsoft.AspNet.Builder;
using Microsoft.AspNet.Hosting;
using Microsoft.AspNet.Mvc;
using Microsoft.Framework.DependencyInjection;
using Microsoft.Framework.Logging;

namespace Sapphire.Cms
{
    public class Startup
    {
        public virtual void ConfigureServices(IServiceCollection services)
        {
            services.AddMvc();
        }

        public virtual void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerfactory)
        {
            app.UseMvc(routes =>
            {
                routes.MapRoute(
                    name: "default",
                    template: "{controller=Home}/{action=Index}/{id?}"
                );
            });
        }
    }
}
```
*Project 3*
~/Startup.cs

```cs
namespace New606
{
    public class Startup : Sapphire.Cms.Startup
    {
    }
}
```

As you can see since Project 2 does everything required, we are currently left having to create a class that serves no purpose.

**Backwards compatibility**
[...]

PR #300, opened 2015-06-25
Locate startup type by name or assembly.

Tratcher:
#254 
Previously you could use the config variable `Hosting:Application` or `app` to specify the assembly in which to find your Startup class.

I've expanded that field so that you can optionally specify a fully qualified type name, including the assembly. `MyType, MyAssembly...`. This also combines with your current dev/production environment  setting so that we will auto locate `MyTypeDev` if you ask for `MyType`.
@muratg @davidfowl @Matthew-Bonner

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
+++ b/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
+using Microsoft.AspNet.Hosting.Internal;
 using Microsoft.Framework.DependencyInjection;
 
 namespace Microsoft.AspNet.Hosting.Startup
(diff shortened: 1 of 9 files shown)


muratg:
:shipit:

davidfowl:
Can we hair support the assembly name?

Tratcher:
Hair?

Tratcher:
Just Assembly name is still supported.

davidfowl:
I mean dropping support for type name. It's a pretty obscure syntax. Let's just use the pattern we have been using. You can specify the assembly but not he type name. It has to be called startup

HaoK:
yeah especially since I just went and nuked this same logic in ServerLoader for the sake of simplicity...

https://github.com/aspnet/Hosting/pull/295

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -38,37 +39,45 @@ public StartupLoader(IServiceProvider services, IHostingEnvironment hostingEnv)
...
-            var assembly = Assembly.Load(new AssemblyName(startupAssemblyName));
+            var nameParts = HostingUtilities.SplitTypeName(startupName);
+            var typeName = nameParts.Item1;
+            var assemblyName = nameParts.Item2;
+
+            var assembly = Assembly.Load(new AssemblyName(assemblyName));
             if (assembly == null)
             {
-                throw new InvalidOperationException(String.Format("The assembly '{0}' failed to load.", startupAssemblyName));
+                throw new InvalidOperationException($"The assembly '{assemblyName}' failed to load.");
```
Eilon: BTW for future reference we shouldn't change exception strings to use string interpolation. When these get localized we'll just have to switch back to `string.Format` anyway.

Eilon:
I agree with supporting only the assembly name. Even the type name lookup is already fancy - don't we support `Startup{EnvironmentName}` already?

HaoK:
Yeah Startup{Env}, Configure{Env}, Configure{Env}Services I think

Tratcher:
I thought one of the points of #254 was the ability to directly specify a startup type. If you want to keep it constrained to specifying an assembly the we don't need this PR at all. The startup assembly can already be specified via the `Hosting:Application` config entry.

Issue #212, opened 2015-03-27
AmbiguousMatchException when starting ASP.NET 5 Web Application


rausmaus:
I was told by @davidfowl to move this here from [Home/#402](https://github.com/aspnet/Home/issues/402)

Description
---
Starting a web application with multiple `Configure`-overloads in `Startup.cs` causes a `AmbigousMatchException`.

This is probably "by design", but the Exception does not provide much information about what the underlaying cause is.

Code:
--------

    using System;
    using Microsoft.AspNet.Builder;
    using Microsoft.AspNet.Hosting;
    using Microsoft.Framework.ConfigurationModel;
    using Microsoft.Framework.DependencyInjection;
    using Microsoft.Framework.Logging;
    using Microsoft.Framework.Logging.Console;
    using Microsoft.AspNet.Diagnostics;

    namespace MausSite
    {
        public class Startup
        {
            public IConfiguration Configuration { get; private set; }

            public Startup(IHostingEnvironment env)
            {
                Configuration = new Configuration()
                    .AddJsonFile("config.json")
                    .AddEnvironmentVariables();
            }

            public void ConfigureServices(IServiceCollection services)
            {
                services.AddMvc();
            }

            public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerfactory)
            {
                loggerfactory.AddConsole();

                // Add the following to the request pipeline only in development environment.
                if (String.Equals(env.EnvironmentName, "development", StringComparison.OrdinalIgnoreCase))
                {
                    app.UseBrowserLink();
                    app.UseErrorPage(ErrorPageOptions.ShowAll);
                }
                else
                {
                    // Add Error handling middleware which catches all application specific errors and
                    // send the request to the following path or controller action.
                    app.UseErrorHandler("/Home/Error");
                }

                app.UseStaticFiles();

                app.UseMvc(routes =>
                {
                    routes.MapRoute(
                        name: "default",
                        template: "{controller}/{action}/{id?}",
                        defaults: new { controller = "Home", action = "Index" });
                });
            }
            // Notice this second "Configure"-overload.
            public void Configure(IApplicationBuilder app) 
            {
                // For more information on how to configure your application, visit http://go.microsoft.com/fwlink/?LinkID=398940
            }
        }
    }

Stack trace:
-----
    System.Reflection.AmbiguousMatchException: Ambiguous match found.
[...]

PR #301, opened 2015-07-01
Check for duplicate Configure or ConfigureServices methods.

Tratcher:
#212 
@davidfowl @rausmaus @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
+++ b/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
@@ -105,8 +105,23 @@ namespace Microsoft.AspNet.Hosting.Startup
         {
             var methodNameWithEnv = string.Format(CultureInfo.InvariantCulture, methodName, environmentName);
             var methodNameWithNoEnv = string.Format(CultureInfo.InvariantCulture, methodName, "");
-            var methodInfo = startupType.GetMethod(methodNameWithEnv, BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static)
-                ?? startupType.GetMethod(methodNameWithNoEnv, BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static);
+
+            var methods = startupType.GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static);
+            var selectedMethods = methods.Where(method => method.Name.Equals(methodNameWithEnv)).ToList();
+            if (selectedMethods.Count > 1)
+            {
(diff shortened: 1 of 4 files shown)


HaoK:
:shipit:

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -105,8 +105,23 @@ private static MethodInfo FindMethod(Type startupType, string methodName, string
...
+            if (selectedMethods.Count == 0)
+            {
+                selectedMethods = methods.Where(method => method.Name.Equals(methodNameWithNoEnv)).ToList();
+                if (selectedMethods.Count > 1)
+                {
+                    throw new InvalidOperationException(string.Format("Having multiple overloads of method '{0}' is not supported.", methodNameWithNoEnv));
+                }
+            }
+
+            var methodInfo = selectedMethods.FirstOrDefault();
```
HaoK: nit: this should be safe to use Single() now right?

HaoK: Guess not it doesn't always throw

StevenLiekens:
Hey sorry to bring up old issues but why not just support overloads instead of throwing?

Right now it's only possible to `Configure` everything in one method. This method gets very large, very quickly.

```csharp
public void Configure(
    IApplicationBuilder app, 
    IHostingEnvironment env,
    ICustomService1 service1,
    ICustomService2 service2,
    ICustomService3 service3,
    ICustomServiceN serviceN,
    )
{
    // build app
    app.UseThis();
    app.UseThat();
    app.Etc();
   
   // initialize custom services
   service1.ConfigureStuff(/* with lots of options */);
   
   service2.MoreStuffToConfigure(/* with lots of options */);
   
   service3.NotDoneConfiguringYet(/* ... */);
   
   serviceN.ThisIsGettingBoring(/* I think you get the point /*);
}
```

I can't think of a technical reason why I shouldn't just be able to split this into overloads?

```csharp
public void Configure( IApplicationBuilder app, IHostingEnvironment env)
{
    // build app
    app.UseThis();
    app.UseThat();
    app.Etc();
}

public void Configure(ICustomService1 service2, IHostingEnvironment env)
{
    //  // initialize custom service 1
    service1.ConfigureStuff(/* with lots of options */);
}

public void Configure(ICustomService1 service2, IHostingEnvironment env)
{
    //  // initialize custom service 2
     service2.MoreStuffToConfigure(/* with lots of options */);
}

// etc
```

The framework could just as well call every overload instead of throwing.

Tratcher:
Comments on closed issues are not tracked, please open a new issue with the details for your scenario.

PR #304, opened 2015-07-19
instruct how to shutdown

mishfit:
seems trivial but even the node based static-server does it. Could prevent future incorrect usages of "Ctrl-C" (or whatever, depending on the platform).

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,7 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
+    <clear />
+    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetrelease/api/v2" />
     <add key="NuGet" value="https://nuget.org/api/v2/" />
   </packageSources>
 </configuration>
--- a/src/Microsoft.AspNet.Hosting/Program.cs
+++ b/src/Microsoft.AspNet.Hosting/Program.cs
@@ -38,6 +38,8 @@ namespace Microsoft.AspNet.Hosting
             using (host.Start())
             {
                 Console.WriteLine("Started");
+                Console.WriteLine("Press Ctrl+C to shutdown.");
+
                 var appShutdownService = host.ApplicationServices.GetRequiredService<IApplicationShutdown>();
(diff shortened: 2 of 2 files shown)


davidfowl:
Except it's actually ctrl+c to shutdown

mishfit:
Interesting... perhaps I had the wrong version. Before updating I had to hit __*enter*__...after updating dnvm, mono, and my default runtime __*ctrl+c*__ doesn't initially appear to work from the console (control has not been returned to bash). Only after a subsequent request do I get the following error and an exit:

```
System.Exception: Error -16 EBUSY resource busy or locked
  at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.Check (Int32 statusCode) [0x00000] in <filename unknown>:0 
  at Microsoft.AspNet.Server.Kestrel.Networking.Libuv.loop_close (Microsoft.AspNet.Server.Kestrel.Networking.UvLoopHandle handle) [0x00000] in <filename unknown>:0 
  at Microsoft.AspNet.Server.Kestrel.Networking.UvLoopHandle.ReleaseHandle () [0x00000] in <filename unknown>:0 
  at System.Runtime.InteropServices.SafeHandle.Dispose (Boolean disposing) [0x00000] in <filename unknown>:0 
  at System.Runtime.InteropServices.SafeHandle.Finalize () [0x00000] in <filename unknown>:0 
```
The browser immediately returns connection refused as the server is shutdown. I've had to __*ctrl+z*__, then find the pid and kill it also to get a sense of immediate termination. The following is on Yosemite 10.10.4:


#### dnvm
```
$ dnvm list

Active Version              Runtime Arch Location             Alias
------ -------              ------- ---- --------             -----
       1.0.0-beta4          mono         ~/.dnx/runtimes      
  *    1.0.0-beta5          mono         ~/.dnx/runtimes      default
```
#### mono
```
$ mono --version
Mono JIT compiler version 4.0.2 (Stable 4.0.2.5/c99aa0c Fri Jul 10 20:47:12 PDT 2015)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
```
#### dnx
```
$ dnx
Microsoft .NET Execution environment Mono-x64-1.0.0-beta5-12103
```

either way, I'll update and get back to you but the delayed shutdown is odd.

mishfit:
@davidfowl do I need to squash? Or are we good with 3 commits?

PR #308, opened 2015-07-23
Fixing dnx.exe arguments in deployer to fix server tests

moozzyk:


troydai:
:shipit: 

Not sure if there are other repo referencing this deployer.
Just need to make sure both Hosting and ServerTests tests pass.

moozzyk:
Both hosting and ServerTests passed on my machine after this change

PR #309, opened 2015-07-27
show how to exit

mishfit:
just display how to shutdown the server on startup

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Program.cs
+++ b/src/Microsoft.AspNet.Hosting/Program.cs
@@ -37,7 +37,8 @@ namespace Microsoft.AspNet.Hosting
             var host = new WebHostBuilder(_serviceProvider, config).Build();
             using (host.Start())
             {
-                Console.WriteLine("Started");
+                Console.WriteLine("Application started. Press Ctrl+C to shut down.");
+
                 var appShutdownService = host.ApplicationServices.GetRequiredService<IApplicationShutdown>();
                 Console.CancelKeyPress += (sender, eventArgs) =>
                 {


src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -38,6 +38,7 @@ public void Main(string[] args)
             using (host.Start())
             {
                 Console.WriteLine("Started");
+                Console.WriteLine("Press Ctrl+C to shutdown.");
```
Eilon: `shut down` is two words.

mishfit:
I don't think that it is (as seen below):
https://en.wikipedia.org/wiki/Shutdown_(computing)
http://www.merriam-webster.com/dictionary/shutdown

node static-server also uses this verbiage.

Do you still want me to change it?

Eilon:
`Shutdown` is a noun, e.g. "the shutdown process." `Shut down` is a compound verb, as in "please shut down the process." You need the verb here.

mishfit:
fair enough.

mishfit:
@eilon done.

Eilon:
@Tratcher look good to you?

Tratcher:
I'd prefer to see the two lines merged: "Application started. Press Ctrl+C to shut down."

Eilon:
Yeah that might be a tiny bit cleaner. @mishfit do you mind making that change? Then we can merge.

mishfit:
@Eilon done.

Eilon:
That was fast! :smile: 

@Tratcher merge please.

Tratcher:
Rebased and merged.

PR #316, opened 2015-08-03
Dispose logger factory when disposing hosting engine

kichalla:
@Tratcher This is related to the PR https://github.com/aspnet/Logging/pull/162

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -39,8 +39,8 @@ namespace Microsoft.AspNet.Hosting.Internal
         private IServerInformation _serverInstance;
 
         public HostingEngine(
-            [NotNull] IServiceCollection appServices, 
-            [NotNull] IStartupLoader startupLoader, 
+            [NotNull] IServiceCollection appServices,
+            [NotNull] IStartupLoader startupLoader,
             [NotNull] IConfiguration config)
         {
             _config = config;
@@ -64,7 +64,8 @@ namespace Microsoft.AspNet.Hosting.Internal
 
             var application = BuildApplication();
 
-            var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
+            var loggerFactory = _applicationServices.GetRequiredService<ILoggerFactory>();
+            var logger = loggerFactory.CreateLogger<HostingEngine>();
             var contextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
             var contextAccessor = _applicationServices.GetRequiredService<IHttpContextAccessor>();
             var server = ServerFactory.Start(_serverInstance,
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -87,6 +88,7 @@ public virtual IDisposable Start()
                 _applicationLifetime.NotifyStopping();
                 server.Dispose();
                 _applicationLifetime.NotifyStopped();
+                loggerFactory.Dispose();
```
kichalla: It feels odd to just dispose this service here. Not sure if there are any other services which need similar requirement.

Tratcher: Agreed, this is weird. Isn't this what DI scopes are for? Would it make sense to create a scope on line 66 and dispose it here? Then any scoped services would be disposed.

kichalla: Hmm...I have seen DI scopes used in `per-request` kind of scenarios where as logger factory here is an application wide service. (thinking of it in a different way...the application start and shutdown itself is a scope and probably we should just dispose off all disposable application services as how DI scope would do).

Tratcher: Consider in the test scenario where we start multiple TestHost/Servers in one process.  HostingEngine is only a scoped subset of an actual application, so I think DI scopes are an appropriate concept here.  Per-request scopes would be nested inside of the per-HostingEngine scopes.

kichalla: :+1:

kichalla:
@rynowak

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
kichalla: @HaoK  and I discussed about this and agreed that disposing loggerfactory alone for this PR is fine and we will file another issue to handle disposing all the hosting engine services as it requires more refactoring than expected.

@Tratcher

Tratcher: Why bother with the partial fix then? It's not urgent.

kichalla: The following PR depends on this
https://github.com/aspnet/Logging/pull/162

Tratcher: No, Hosting depends on Logging. That Logging PR is nice to have, but calling Dispose not a blocking issue for beta7.

Your partial fix here does not contribute towards the full fix, so don't bother.

kichalla:
Closing this PR as it will be handled as part of fix for https://github.com/aspnet/Hosting/issues/322

PR #325, opened 2015-08-12
Move HttpContextAccessor to HttpAbstractions

HaoK:
cc @Tratcher @davidfowl


HaoK:
I'll remove identity's reference to hosting once these are in and everything passes once

HaoK:
ad6e4b838a5daad0b9a36870e5d0e3f586e9ca06

PR #332, opened 2015-08-15
Removed builder arg parameter from IStartupFilter

davidfowl:
- It broke the composition model by allowing you to
reach out to the original app builder. This breaks
the ability to properly wrap and have all configure
methods see the wrapper.

/cc @HaoK @Tratcher @loudej


Tratcher:
:shipit:

Issue #320, opened 2015-08-07
Support HttpContext.RequestAborted in TestServer


tmds:
Request to be able to test aborted requests.

HttpContext.RequestAborted should be cancelled when:
- the CancellationToken passed to SendAsync is cancelled
- the HttpMessageHandler is Disposed

Tratcher:
Note that disposing the HttpClient calls HttpClient.CancelPendingRequests(), so we don't have to do anything special for that case. We just need to pass the CancellationToken through.

tmds:
I tried implementing the SendAsync-case.

In the ClientHandler, where the HttpContext is created. I set RequestAborted to be equal to the cancellationToken.

To test this I made a Website where I added a SemaphoreSlim which does a WaitAsync using the RequestAborted. I never release the SemaphoreSlim, so the only way it will unblock is by cancelling the token.

Now when I cancel the token, the SemaphoreSlim unblocks by throwing an Exception. This is okay. What is unexpected to me is that the exception is of type ObjectDisposedException and not of type OperationCancelledException.

I saw that the CancellationTokenSource of RequestAborted is different than the one I create for the SendAsync call. The HttpClient uses it in a LinkedTokenSource.

It is not clear to me why this is throwing an ObjectDisposedException instead of an OperationCancelledException. Any thoughts?

Tratcher:
It's hard to guess without seeing the test code and the exception stack trace.

tmds:
I made a repo with a test here: https://github.com/tmds/RequestAbortedTestCase.git

This is the small change made to the ClientHandler: https://github.com/tmds/RequestAbortedTestCase/blob/master/src/Microsoft.AspNet.TestHost/ClientHandler.cs#L120

These are the 3 things I observe:

When running the TestWebsite project and pressing the stop button in the browser, an OperationCancelledException is caught here: https://github.com/tmds/RequestAbortedTestCase/blob/master/src/TestWebSite/Startup.cs#L36. This is the behaviour I would like to get when using the TestServer.

When I run the [test](https://github.com/tmds/RequestAbortedTestCase/blob/master/src/Tests/TestCase.cs#L15) in 'debug', the call to [aquire the semaphore](https://github.com/tmds/RequestAbortedTestCase/blob/master/src/TestWebSite/Startup.cs#L31) does not return.

[...]

PR #335, opened 2015-08-20
Request aborted

tmds:
Support RequestAborted (fixes https://github.com/aspnet/Hosting/issues/320)
Add PipelineCompleteAsync to wait for pipeline operations for a request to complete.

Code based on 1.0.0-beta6 branch


src/Microsoft.AspNet.Hosting.Abstractions/project.json
```diff
@@ -1,14 +1,17 @@
...
     "description": "ASP.NET 5 Hosting abstractions.",
     "repository": {
         "type": "git",
         "url": "git://github.com/aspnet/hosting"
     },
     "dependencies": {
-        "Microsoft.AspNet.Http.Abstractions": "1.0.0-*",
-        "Microsoft.AspNet.FileProviders.Abstractions": "1.0.0-*",
-        "Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" }
+        "Microsoft.AspNet.Http.Abstractions": "1.0.0-beta6",
```
Tratcher: Please go back to `*`, we're working on beta8 now.

Tratcher:
lock.json files should not be checked in.

Tratcher:
Most of these changes are noise because your branch is based on master and you're trying to merge into dev. please rebase your changes onto dev directly.

src/Microsoft.AspNet.TestHost/HttpResponseMessageExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.TestHost
+{
+    public static class HttpResponseMessageExtensions
+    {
+        public static Task PipelineCompleteAsync(this HttpResponseMessage message)
```
Tratcher: This seems overkill. Why can't the client just wait for the end of the response stream?

tmds: I agree this is not necessary for implementing RequestAborted.
The test I was planning to write using RequestAborted has an effect on subsequent requests. I noticed the effects weren't visible immediately because when the ResponseStream returns, the pipeline hadn't finished yet.
I think it is an interesting feature independent of RequestAborted.
If you like, I can leave it out here and open another feature request?

Tratcher: Why is the response stream completing before the end of the pipeline? Is it checking content-length? Or did somebody call Dispose on it?

tmds: The test called Dispose on it to abort the request.

Tratcher: I think this is more complicated than we want the test client to be. We want HttpClient to be used like it normally would. If you really need to control the server data structures and flow then you can call TestServer.Invoke directly.

test/Microsoft.AspNet.TestHost.Tests/TestClientTests.cs
```diff
@@ -7,6 +7,8 @@
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
 using Xunit;
+using System.Threading;
```
Tratcher: Ordering, System goes first.

test/Microsoft.AspNet.TestHost.Tests/TestClientTests.cs
```diff
@@ -111,5 +113,38 @@ public TestClientTests()
...
+                }
+            };
+
+            // Act
+            var server = TestServer.Create(app => app.Run(appDelegate));
+            var client = server.CreateClient();
+            var request = new HttpRequestMessage(HttpMethod.Get, "http://localhost:12345");
+            var response = await client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead);
+            // Abort Request
+            response.Dispose();
```
Tratcher: There should be a separate test for aborting a request before the response is received. You could do this with a CancellationToken or just dispose the HttpClient.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -70,33 +70,36 @@ public ClientHandler([NotNull] Func<IFeatureCollection, Task> next, PathString p
...
+        private class RequestState
         {
             private readonly HttpRequestMessage _request;
             private TaskCompletionSource<HttpResponseMessage> _responseTcs;
             private ResponseStream _responseStream;
             private ResponseFeature _responseFeature;
+            private CancellationTokenSource _requestAbortedSource;
+            private bool _pipelineFinished;
 
             internal RequestState(HttpRequestMessage request, PathString pathBase, CancellationToken cancellationToken)
```
Tratcher: This incoming CT needs to be linked to _requestAbortedSource.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -146,9 +149,10 @@ internal RequestState(HttpRequestMessage request, PathString pathBase, Cancellat
                     }
                 }
 
-                _responseStream = new ResponseStream(CompleteResponse);
+                _responseStream = new ResponseStream(ReturnResponseMessage, AbortRequest);
                 HttpContext.Response.Body = _responseStream;
                 HttpContext.Response.StatusCode = 200;
+                HttpContext.RequestAborted = _requestAbortedSource.Token;
```
Tratcher: You can remove line 137 now, it's redundant with this.

NuGet.Config
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetvnext/api/v2" />
-    <add key="NuGet" value="https://nuget.org/api/v2/" />
+    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetlitedev/api/v2" />
+    <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
```
Tratcher: there's still something odd with your rebase if you are including changes in this file.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
Tratcher: nevermind, that's on line 65

tmds: There is a registration in SendAsync which calls AbortRequest, which then Cancels _requestAbortedSource. AbortRequest also calls Complete on the _responseStream.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -160,7 +164,28 @@ public Task<HttpResponseMessage> ResponseTask
...
+            public void AbortRequest()
+            {
+                if (!_pipelineFinished)
+                {
+                    _requestAbortedSource.Cancel();
+                }
+                _responseStream.Complete();
+            }
+
+            internal void PipelineComplete()
```
Tratcher: Method names should be verbs/actions. It's not clear why you renamed most of these methods.

test/Microsoft.AspNet.TestHost.Tests/TestClientTests.cs
```diff
@@ -111,5 +113,67 @@ public TestClientTests()
...
+                {
+                    tcs.SetException(e);
+                }
+            };
+
+            // Act
+            var server = TestServer.Create(app => app.Run(appDelegate));
+            var client = server.CreateClient();
+            var cts = new CancellationTokenSource();
+            cts.Cancel();
```
Tratcher: It's really odd to pass in an already cancelled CT. use a 500ms time instead.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
tmds: Renames where part of PipelineCompleteAsync feature, which has been removed.

Tratcher:
I synced your branch and I finally understand what's wrong with your history. Your source tree is a merger of dev and master which causes a big mess. The PR needs to be based on the dev branch with no mergers. I tried cherry picking things over to dev but ran into a bunch of conflicts. Try resetting to dev and pulling your changes over by source. That will help you squash all the commits in the process.

tmds:
Replaced by https://github.com/aspnet/Hosting/pull/345

Issue #315, opened 2015-08-02
Support WebSocket tests using TestServer


tmds:
Request to be able to test WebSockets using the TestServer.

WebSocket testing could start by calling a new method TestServer.CreateWebSocketClient(). This method returns a WebSocketClient. This is similar to the normal http request testing: TestServer.CreateClient method which returns an HttpClient.

Tratcher:
It sounds doable, and kindof fun actually. I don't think it's a very high priority though, so I can't say how soon it would happen.

davidfowl:
Sounds like it would make a good contribution

tmds:
@Tratcher, @davidfowl, if you want, you can assign the issue to me.

davidfowl:
@tmds issues can't be assinged to non core contributors

tmds:
@davidfowl I guess you mean the tool doesn't allow it.
I want to check whether it is worth spending my time, that is: you guys won't work on it in the short term and it is 'up for grabs'.

davidfowl:
We don't have any plans to do it in the short term so feel free to do it

PR #336, opened 2015-08-22
TestHost: support WebSocket testing

tmds:
Implement https://github.com/aspnet/Hosting/issues/315


src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public Action<HttpResponseMessage> InspectResponse
+        {
+            get;
+            set;
+        }
+
+        public async Task<WebSocket> ConnectAsync(Uri uri, CancellationToken cancellationToken)
+        {
+            HttpClient client = _testServer.CreateClient();
```
Tratcher: I think the WebSocketClient can bypass the HttpClient entirely. HttpClient doesn't have any concept of WebSockets, so tacking them on is pretty awkward. Instead of creating an HttpRequestMessage here just create an IFeatureCollection & DefaultHttpContext directly (This may be a little easer after https://github.com/aspnet/HttpAbstractions/issues/367).

src/Microsoft.AspNet.TestHost/project.json
```diff
@@ -7,6 +7,7 @@
     },
     "dependencies": {
         "Microsoft.AspNet.Hosting": "1.0.0-*",
+        "Microsoft.AspNet.WebSockets.Protocol": "1.0.0-*",
```
Tratcher: This dependency will cause a build order issue on our CI (HttpAbstractions -> Hosting -> Servers -> WebSockets -> Hosting). I think there's a simpler way to implement it that doesn't require the dependency anyways.  Because the data is never sent out over an actual network, we don't even have to serialize the web socket messages down to a Stream, we can just pass the full messages back and forth between the 'client' and 'server' ends of the websocket. You'd need to change your reader/writer buffer to store full messages rather than just bytes.

test/Microsoft.AspNet.TestHost.Tests/TestClientTests.cs
```diff
@@ -111,5 +113,35 @@ public TestClientTests()
             // Assert
             Assert.Equal("Hello world POST Response", await response.Content.ReadAsStringAsync());
         }
+
+        [Fact]
+        public async Task WebSocketWorks()
```
Tratcher: Needs at least one test that reads and writes a few messages back and forth.

src/Microsoft.AspNet.TestHost/ResponseMessage.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.TestHost
+{
+    class ResponseMessage : HttpResponseMessage
```
davidfowl: All classes need the visibility. Please see our [Engineering Guidelines](https://github.com/aspnet/Home/wiki/Engineering-guidelines)

tmds:
@Tratcher I re-implemented the feature based on your feedback
I still need to add the read/write test. Also, the WebSocket.Abort/Dispose logic needs to be checked.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -66,22 +66,22 @@ public ClientHandler([NotNull] Func<IFeatureCollection, Task> next, PathString p
 
             // Async offload, don't let the test code block the caller.
             var offload = Task.Factory.StartNew(async () =>
+            {
```
Tratcher: You can revert this file if only the formatting has changed.

src/Microsoft.AspNet.TestHost/WebSocket.cs
```diff
@@ -0,0 +1,339 @@
...
+using System.IO;
+using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.TestHost
+{
+    class WebSocket : System.Net.WebSockets.WebSocket
```
Tratcher: Needs an actual name. TestWebSocket? InMemoryWebSocket?

Tratcher: And a visibility

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,177 @@
...
+        public Action<HttpRequestMessage> ConfigureRequest
+        {
+            get;
+            set;
+        }
+
+        public async Task<System.Net.WebSockets.WebSocket> ConnectAsync(Uri uri, CancellationToken cancellationToken)
+        {
+            // clientRequest
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, uri);
```
Tratcher: You can skip the HttpRequestMessage here and go directly to the DefaultHttpContext.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                // HttpContext
+                FeatureCollection = new FeatureCollection();
+                var httpContext = new DefaultHttpContext(FeatureCollection);
+
+                // Request
+                httpContext.SetFeature<IHttpRequestFeature>(new RequestFeature());
+                var serverRequest = httpContext.Request;
+                serverRequest.Protocol = "HTTP/" + clientRequest.Version.ToString(2);
+                serverRequest.Scheme = clientRequest.RequestUri.Scheme;
```
Tratcher: If the user passes in `ws` or `wss` it should be converted to `http` or `https` respectively for the server.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+            public void PipelineComplete()
+            {
+                PipelineFailed(null);
+            }
+
+            public void PipelineFailed(Exception ex)
+            {
+                _websocketTcs.TrySetException(new InvalidOperationException("The websocket was not accepted.", ex));
```
Tratcher: Include the status code if ex is null?

Tratcher:
This looks much better overall. It's definitely going to need some tests though.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,177 @@
...
+            {
+                get
+                {
+                    return true;
+                }
+            }
+
+            Task<System.Net.WebSockets.WebSocket> IHttpWebSocketFeature.AcceptAsync(WebSocketAcceptContext context)
+            {
+                var websockets = WebSocket.CreatePair(context.SubProtocol);
```
Tratcher: Store both of these and abort/dispose them when _next finishes or throws (e.g. PipelineComplete or PipelineFailed).

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,177 @@
...
+                }
+            });
+
+            return await state.WebSocketTask;
+        }
+
+        private class RequestState : IHttpWebSocketFeature
+        {
+            public IFeatureCollection FeatureCollection { get; private set; }
+            public Task<System.Net.WebSockets.WebSocket> WebSocketTask { get { return _websocketTcs.Task; } }
```
davidfowl: Why is the full namespace expanded here?

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,177 @@
...
+
+            private TaskCompletionSource<System.Net.WebSockets.WebSocket> _websocketTcs;
```
davidfowl: And here?

src/Microsoft.AspNet.TestHost/WebSocket.cs
```diff
@@ -0,0 +1,339 @@
...
+                    {
+                        throw new ObjectDisposedException(typeof(WebSocket).FullName);
+                    }
+                    if (_receiverClosed)
+                    {
+                        throw new IOException("The remote end closed the connection.", new ObjectDisposedException(typeof(WebSocket).FullName));
+                    }
+
+                    // we return immediately so we need to copy the buffer since the sender can re-use it
+                    byte[] array = new byte[message.Buffer.Count];
```
davidfowl: needs more var

src/Microsoft.AspNet.TestHost/WebSocket.cs
davidfowl: Is this largely a reuse of our websocket middleware code?

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
tmds: This is useful for the ConfigureRequest callback. Should I keep this or drop the ConfigureRequest?

src/Microsoft.AspNet.TestHost/WebSocket.cs
tmds: Indeed, code is based on websocket middleware.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
tmds: Because of the name clash with WebSocket class in the Hosting namespace, will rename WebSocket to TestWebSocket as suggested by @Tratcher.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
tmds: Will do.
I will also set the status code to 101 in the AcceptAsync call.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
Tratcher: ConfigureRequest is a good feature, but it can use HttpContext just as well. ConfigureHttpContext?

src/Microsoft.AspNet.TestHost/WebSocket.cs
davidfowl: @Tratcher Surely there's a way to reuse that code without copying the implementation here? We can shuffle the CI dependencies or move things to other repos if it causes problems.

Tratcher: It's not the same code, it's only 'inspired by' the websocket middleware. It cuts out the wire-serialization layer and only passes messages back and forth. This should make the test scenarios simpler and easier to debug.

tmds:
@Tratcher, @davidfowl, I improved the code further based on your comments
Two points:
- WebSocketClient disposes the server websocket when the pipeline ends. Disposing the client websocket is the responsibility of the client (i.e. the test)
- I replaced HttpRequestMessage by HttpRequest for WebSocketClient.ConfigureRequest. I limited this to the request, as this is also the case for the [real WebSocketClient](https://github.com/aspnet/WebSockets/blob/552b336a78c4835e7f84a94b1e5a07df4bc09b39/src/Microsoft.AspNet.WebSockets.Client/WebSocketClient.cs#L62).

src/Microsoft.AspNet.TestHost/TestWebSocket.cs
```diff
@@ -0,0 +1,354 @@
...
+using System.IO;
+using System.Linq;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.TestHost
+{
+    internal class TestWebSocket : System.Net.WebSockets.WebSocket
```
Tratcher: Do you still need the full type name here? `System.Net.WebSockets.WebSocket`

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,179 @@
...
+            private set;
+        }
+
+        public Action<HttpRequest> ConfigureRequest
+        {
+            get;
+            set;
+        }
+
+        public async Task<System.Net.WebSockets.WebSocket> ConnectAsync(Uri uri, CancellationToken cancellationToken)
```
Tratcher: Or here?

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,179 @@
...
+                    state.Dispose();
+                }
+            });
+
+            return await state.WebSocketTask;
+        }
+
+        private class RequestState : IDisposable, IHttpWebSocketFeature
+        {
+            private TaskCompletionSource<System.Net.WebSockets.WebSocket> _clientWebSocketTcs;
```
Tratcher: etc.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -0,0 +1,179 @@
...
+                // HttpContext
+                FeatureCollection = new FeatureCollection();
+                HttpContext = new DefaultHttpContext(FeatureCollection);
+
+                // Request
+                HttpContext.SetFeature<IHttpRequestFeature>(new RequestFeature());
+                var request = HttpContext.Request;
+                request.Protocol = "HTTP/1.1";
+                var scheme = uri.Scheme;
+                scheme = (scheme == "ws") ? "http" : scheme;
```
Tratcher: case insensitive

tmds: The Uri class converts the scheme to lower case.

test/Microsoft.AspNet.TestHost.Tests/TestClientTests.cs
```diff
@@ -7,6 +7,10 @@
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Http;
 using Xunit;
+using System.Threading;
```
Tratcher: Sort, System first.

tmds:
@Tratcher, is this ready for merging?

Tratcher:
@tmds Sorry, this is on hold while we resolve an internal discussion about WebSockets. We'll get back to you soon.

cc: @DamianEdwards

tmds:
@Tratcher no problem, take your time

Tratcher:
:shipit: I'll try rebasing myself.

Tratcher:
Rebased and merged.

PR #340, opened 2015-08-25
Add test verifying ApplicationServices replacement

HaoK:
cc @Tratcher @davidfowl


HaoK:
Replacing with PR with extra test and fix of bug

Issue #337, opened 2015-08-22
HostEngine Start not setting HttpContext.ApplicationServices - uses default DI provider


BillKrat:
I am returning a UnityServiceProvider from ConfigureServices and I can see that it is being set to the ApplicationServices property (above the HostingEngine.Start method in top right of image below), however the httpContext.ApplicationServices is null which causes it to use _services resulting in the default DI provider to be used.  I see the DefaultControllerActivator retrieves the HttpContext.RequestServices (default provider) resulting in my HomeController not resolving IFoo as expected (line 104 left pane) since it is not using my UnityServiceProvider.

I am making assumptions based on the limited documentation available during internet research, but will be glad to blog the topic if I am missing something and you can elaborate...

UPDATED: My work around was to add the following line in HostEngine.Start() that sets RequestServices as follows:

                    using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
                    {
                        // BillKrat.2015.08.23 - work around so DI container is used
                        httpContext.RequestServices = ApplicationServices;

                        contextAccessor.HttpContext = httpContext;
                        await application(httpContext);
                    }

![serviceprovider](https://cloud.githubusercontent.com/assets/10552946/9426301/634dcea8-48fd-11e5-83e7-a1cd91c81014.png)

davidfowl:
Looks like a bug.

glennc:
@HaoK Can you investigate?

HaoK:
Ah, this is probably a sideeffect of 'cleaning up' some of the weird logic last time... Yeah I'll take a look and add a test for this

HaoK:
So I added a test which is just verifying that ApplicationServices is set to the IServiceProvider returned via ConfigureServices (shown below), which works fine.  Are you not seeing context.ApplicationServices being set to the one returned from your ConfigureServices method?

```
        public class CustomContainerStartup
        {
            public IServiceProvider Services;
            public IServiceProvider ConfigureServices(IServiceCollection services)
            {
                Services = services.BuildServiceProvider();
                return Services;
            }

            public void Configure(IApplicationBuilder app)
            {
                app.Run(async context =>
                {
                    await context.Response.WriteAsync("ApplicationServicesEqual:" + (context.ApplicationServices == Services));
                });
            }
        }
```

BillKrat:
[...]

PR #341, opened 2015-08-25
Set applicationServices on context explicitly

HaoK:
We were relying on the RequestServices magic to set it before

Fixes https://github.com/aspnet/Hosting/issues/337

cc @Tratcher @davidfowl


Tratcher:
:shipit:

HaoK:
55b28abeab524c39bacc1a0e0717d5126ec3a6a7

PR #352, opened 2015-09-14
Replacing NotNullAttribute with thrown exceptions

pranavkm:


src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
```diff
@@ -142,22 +165,32 @@ public WebHostBuilder UseServer(IServerFactory factory)
...
 
-        public WebHostBuilder UseStartup([NotNull] Type startupType)
+        public WebHostBuilder UseStartup(Type startupType)
         {
             if (startupType == null)
             {
                 throw new ArgumentNullException(nameof(startupType));
             }
+
+            if (startupType == null)
```
Tratcher: duplicate

Tratcher:
:shipit:

tmds:
Is NotNullAttribute deprecated?

pranavkm:
@tmds, the feature to support it isn't ready for v1.

tmds:
I saw the announcement about 'dnu sources' and I thought it only had impact on the repository section on the json file. But it also means that Sources dependencies are not supported?

pranavkm:
@tmds, no the feature that's not ready is discovery of precompiled modules in packages.Precompiled modules work when added as a source (which is how I'm generating these changes), but cannot be brought in from packages as yet.

tmds:
ok, thanks

PR #355, opened 2015-09-15
Pool DefaultHttpContext [Design]

benaadams:
Resolves https://github.com/aspnet/HttpAbstractions/issues/402 needs PR https://github.com/aspnet/HttpAbstractions/pull/404 done first

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -71,7 +71,6 @@ namespace Microsoft.AspNet.Hosting.Internal
                 async features =>
                 {
                     var httpContext = contextFactory.CreateHttpContext(features);
-                    httpContext.Response.RegisterForDispose(httpContext);
 
                     httpContext.ApplicationServices = _applicationServices;
                     var requestIdentifier = GetRequestIdentifier(httpContext);
@@ -82,6 +81,7 @@ namespace Microsoft.AspNet.Hosting.Internal
                         await application(httpContext);
                     }
 
+                    return httpContext;
                 });
 
             _applicationLifetime.NotifyStarted();
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -70,14 +70,16 @@ namespace Microsoft.AspNet.Hosting.Internal
             var server = ServerFactory.Start(_serverInstance,
                 async features =>
                 {
-                    var httpContext = contextFactory.CreateHttpContext(features);
-                    httpContext.ApplicationServices = _applicationServices;
-                    var requestIdentifier = GetRequestIdentifier(httpContext);
-
-                    using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
+                    using (var httpContext = contextFactory.CreateHttpContext(features))
                     {
-                        contextAccessor.HttpContext = httpContext;
-                        await application(httpContext);
+                        httpContext.ApplicationServices = _applicationServices;
+                        var requestIdentifier = GetRequestIdentifier(httpContext);
+
+                        using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
+                        {
+                            contextAccessor.HttpContext = httpContext;
+                            await application(httpContext);
+                        }
                     }
                 });
 
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -70,17 +70,18 @@ namespace Microsoft.AspNet.Hosting.Internal
             var server = ServerFactory.Start(_serverInstance,
                 async features =>
                 {
-                    using (var httpContext = contextFactory.CreateHttpContext(features))
-                    {
-                        httpContext.ApplicationServices = _applicationServices;
-                        var requestIdentifier = GetRequestIdentifier(httpContext);
+                    var httpContext = contextFactory.CreateHttpContext(features);
+                    httpContext.Response.RegisterForDispose(httpContext);
 
-                        using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
-                        {
-                            contextAccessor.HttpContext = httpContext;
-                            await application(httpContext);
-                        }
+                    httpContext.ApplicationServices = _applicationServices;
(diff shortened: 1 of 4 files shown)


benaadams:
See https://github.com/aspnet/HttpAbstractions/issues/402#issuecomment-140336990 for details on server run improvements (61.5k rps -> 1.49M rps)

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -70,14 +70,16 @@ public virtual IDisposable Start()
             var server = ServerFactory.Start(_serverInstance,
                 async features =>
                 {
-                    var httpContext = contextFactory.CreateHttpContext(features);
-                    httpContext.ApplicationServices = _applicationServices;
-                    var requestIdentifier = GetRequestIdentifier(httpContext);
-
-                    using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
+                    using (var httpContext = contextFactory.CreateHttpContext(features))
```
Tratcher: The scope of HttpContext is not this deterministic. IHttpResponseFeature has two events, OnStarting and OnCompleted, that may not have fired before the end of this using block. Both of these events may have user state that contains the HttpContext.  I'm not sure there is a simple workaround.

Tratcher:
This is a little dangerous. HttpContext is the primary touchpoint for application APIs and is likely to be referenced in non-linear ways like events, background threads, etc.. Consider SignalR which does much of its work on background threads and could easily get into race conditions with this change. Arguably SignalR needs to guard against referencing the HttpContext after the 'end' of the request, but before the worst that could happen was an ObjectDisposedException. Now recycling the HttpContext underneath them could cause pretty terrible side effects like having the wrong user credentials.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
benaadams: Isn't FireOnCompleted/OnCompleted called by the end of ```await application(httpContext);```?

e.g. https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Http/Frame.cs#L226

Else it will never be called which would be a problem?

benaadams:
>  Consider SignalR which does much of its work on background threads and could easily get into race conditions with this change. 

Will get back to you on this; I think I can add a safety layer that will throw errors at a cost of one small object per request.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
Tratcher: I think you have your stack backwards. 'Application' in that kestrel code block refers to the async lamda you're in here (LN71-83), so FireOnCompleted runs after your using block.

benaadams: Ahh... hmm...

benaadams: K, I think the problem is also the solution :)

benaadams: Removed ```using``` and added ```httpContext.Response.RegisterForDispose(httpContext);```

Will need to change it so Disposing is done after OnCompleted rather than during; but probably should be anyway?

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -71,6 +71,8 @@ public virtual IDisposable Start()
                 async features =>
                 {
                     var httpContext = contextFactory.CreateHttpContext(features);
+                    httpContext.Response.RegisterForDispose(httpContext);
```
davidfowl: That is pretty weird...

benaadams: I think its the weirdest code I've written...

Most sensibly it should return Task<IDisposable> rather than Task; but then' I'd have to add a PR to Kestrel as well... Ok will do that?

Tratcher: There's still and ordering problem. If this registers first it will run first. I think you'd have to register after `await application` to ensure you'd be last.

benaadams: Going to return IDisposable from the closure; HttpContext has always been IDisposable, just never Disposed so will be just closing that loop.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -79,6 +80,8 @@ public virtual IDisposable Start()
                         contextAccessor.HttpContext = httpContext;
                         await application(httpContext);
                     }
+
+                    return httpContext;
```
halter73: Is this so httpContext gets disposed? What are we waiting for?

I would just wrap the call to `await application(httpContext)` in a using statement. Once the Task returned from the application is completed, you should be safe to dispose the httpContext.

Tratcher: @halter73 He tried that first, see https://github.com/aspnet/Hosting/pull/355#discussion_r39513731

halter73: Is the HttpContext supposed to be usable in OnCompleted? I think the simplest solution is to not support it. It's not like you would be able to change the response at that point.

I'm not a big fan of returning an IDisposable here, because not obvious why it's necessary. It could definitely be a gotcha for anyone implementing or using the IServerFactory interface.

I would even prefer having a "OnTearDown" IHttpResponseFeature that runs after OnCompleted, only because the purpose is more clear.

Tratcher: I'm not recommending any specific solution, just trying to outline what's possible. HttpContext is supported in OnCompleted, but more importantly it's supported in OnStarted which can fire after `await application(httpContext)` in no-body scenarios.

benaadams: Is open question on whether the Context is the place to dispose of the features collection: https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http/DefaultHttpContext.cs#L172

Currently it is in Dispose; but the dispose never gets called as its not returned, so the features collection is never disposed?

Have raised related issue https://github.com/aspnet/Hosting/issues/359 as these have all been PRs where its slightly tangential.

davidfowl:
Closing this as discussed with @benaadams

Issue #354, opened 2015-09-14
Publish begin/end request and unhandled errors to telemetry source


Eilon:
Publish these:
1. Request start/end, response start/end (or the closest we can get to these)
2. Unhandled errors that get bubbled up to the servers

Example in MVC: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/Actions/MvcRouteHandler.cs#L73-L78

Eilon:
Related: https://github.com/aspnet/Diagnostics/issues/180

Issue #353, opened 2015-09-14
Move MVC's TelemetrySource registration to Hosting


Eilon:
This avoids multiple components each trying to register the services themselves.

Source codes here: https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L139-L148

Eilon:
No, get it done in beta 6 :trollface:

PR #360, opened 2015-09-16
Add telemetry for begin/end request and unhandled exceptions.

Tratcher:
#353, #354 
@Eilon @rynowak @lodejard @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics.Tracing;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Hosting.Builder;
 using Microsoft.AspNet.Hosting.Internal;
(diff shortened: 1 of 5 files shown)


Eilon:
Looks pretty straightforward, :shipit:

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
```diff
@@ -91,6 +92,10 @@ private IServiceCollection BuildHostingServices()
             services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
             services.AddLogging();
 
+            var telemetrySource = new TelemetryListener("Microsoft.AspNet");
+            services.AddInstance<TelemetrySource>(telemetrySource);
+            services.AddInstance<TelemetryListener>(telemetrySource);
```
rynowak: Looks very nice - I'll remove this bit from MVC.

rynowak:
/cc @delmyers @avanderhoorn as FYI

**Annnnnnnnnnnnnnnnnnnnnnnnnd......................................** :shipit:

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
rynowak: Orthogonal to this PR, but @davidfowl - do you still think we need to do something different here?

delmyers:
Thanks!

Issue #322, opened 2015-08-10
Dispose hosting engine services once the application is shutting down


kichalla:
Scenario: LoggerFactory needs to be disposed once the application shutdown so that dispose can be called on all the registered logger providers. This is so that these logger providers can flush any data that they might be holding.

@Tratcher @HaoK 
The proposal is to create  a DI scope for hosting engine services which get disposed once the hosting engine is being disposed.

HaoK:
Is there anything other than the logger that is going to be disposed as part of this scope cleanup?  If not is it really worth creating a scope to dispose one service?

davidfowl:
+1 to @HaoK's comment. Why would we create a scope just to dispose the ILoggerFactory.

dougbu:
@Tratcher and I suspect aspnet/Mvc#3066 is related. Could fix that one in other ways but could also piggy-back on a solution here.

PR #362, opened 2015-09-16
Dispose the service provider on app shutdown

davidfowl:
- Added a unit test

#322


src/Microsoft.AspNet.Hosting/Internal/Application.cs
```diff
@@ -32,6 +32,7 @@ public IServiceProvider Services
         public void Dispose()
         {
             _stop.Dispose();
+            (_services as IDisposable)?.Dispose();
```
Tratcher: Move this into the Disposable created in HostingEngine.Start

Tratcher:
:shipit:

Issue #358, opened 2015-09-15
[HttpPlatformHandler]: Support HTTP_PLATFORM_PORT env variable as an address


davidfowl:
The HttpPlatformHandler passes the port via the env variable `HTTP_PLATFORM_PORT`.

davidfowl:
We should also support port

/cc @Tratcher

Tratcher:
Have Hosting read HTTP_PLATFORM_PORT and add it to IServerAddressesFeature.Addresses?

davidfowl:
Yep

Tratcher:
Actually, we can do it in the new UseHttpPlatformHandler extension.

davidfowl:
We decided to put this in hosting

Tratcher:
Move the default address from Kestrel to hosting and make it http://localhost:5000/. Make HTTP_PLATFORM_PORT into http://localhost:port/

PR #365, opened 2015-09-18
Support HTTP_PLATFORM_PORT.

Tratcher:
#358 @davidfowl @muratg @halter73

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -22,6 +22,9 @@ namespace Microsoft.AspNet.Hosting.Internal
 {
     public class HostingEngine : IHostingEngine
     {
+        // This is defined by IIS's HttpPlatformHandler.
+        private static readonly string ServerPort = "HTTP_PLATFORM_PORT";
+
         private readonly IServiceCollection _applicationServiceCollection;
         private readonly IStartupLoader _startupLoader;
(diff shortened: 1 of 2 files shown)


halter73:
:shipit:

halter73:
The ServerTests will rely on this to pass, correct?

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -188,6 +191,16 @@ private RequestDelegate BuildApplication()
             var builder = builderFactory.CreateBuilder(_serverInstance);
             builder.ApplicationServices = _applicationServices;
 
+            var port = _config[ServerPort];
+            if (!string.IsNullOrEmpty(port))
+            {
+                var addresses = builder.ServerFeatures.Get<IServerAddressesFeature>()?.Addresses;
+                if (addresses != null && !addresses.IsReadOnly)
```
davidfowl: When is this ever read only.

davidfowl:
Add a test

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
Tratcher: If the server doesn't support modifying it at runtime. e.g. IIS/Helios.

Tratcher:
@halter73 Yes, the new HttpPlatformHandler tests will require it.

Tratcher:
Added test.

davidfowl:
:shipit:

PR #369, opened 2015-09-23
Switch to using feature for RequestServices

HaoK:


Tratcher:
Why? Bug #?

davidfowl:
@Tratcher Performance. Making this lazy. The only problem is the fact that IServiceProviders feature is pretty hidden.

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerFeature.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace Microsoft.AspNet.Hosting.Internal
+{
+    public class RequestServicesFeature : IServiceProvidersFeature, IDisposable
+    {
+        private readonly IServiceScopeFactory _scopeFactory;
+        private IServiceProvider _requestServices;
+        private IServiceScope _scope;
+        private bool _requestServicesSet;
+
+        public RequestServicesFeature(IServiceProvider applicationServices, IServiceScopeFactory scopeFactory)
```
davidfowl: Do we even need to resolve the factory this early?

HaoK: Resolve it from AppServices in the getter instead?

HaoK:
Added a test to verify we no-op if there's an existing ISPsFeature.  Removed IServiceScope from ctors and just GetRequiredService when needed

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerFeature.cs
```diff
@@ -0,0 +1,58 @@
...
+            set
+            {
+                _requestServicesSet = true;
+                RequestServices = value;
+            }
+        }
+
+        public void Dispose()
+        {
+            if (_scope != null)
```
davidfowl: I was gonna say this is so old school:

```
_scope?.Dispose();
```

HaoK: Sure, can just nuke the if then

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerFeature.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public IServiceProvider ApplicationServices { get; set; }
+
+        public IServiceProvider RequestServices
+        {
+            get
+            {
+                if (!_requestServicesSet)
+                {
+                    _scope = ApplicationServices.GetRequiredService<IServiceScopeFactory>().CreateScope();
```
davidfowl: Does this need to be defensive since ApplicationServices is settable?

HaoK: Yeah looks like it.  Do we want to throw an exception if you try to set it to null, or just treat that as no-op and basically do nothing and leave RequestServices null if you set ApplicationServices to null?

davidfowl: This behavior is totally undefined. To repro this you would have to set ApplicationServices to null then before ever touching RequestServices (I just know somebody is going to do something stupid and report a bug :smile:). This works today because nothing is lazy. We could store the original application services and use that as a fallback but that might be perplexing. Throwing might be the most reasonable thing. Null isn't bad either so lets flip a coin. If you did this as a customer an exception might be the best since you'd want to know why instead of getting a null ref. On the other hand if you get RequestServices and then set ApplicationServices to null, should that throw?

HaoK: Is there any real scenario where we want to allow them to set ApplicationServices to null?  If not, I can just un auto the property and throw for null sets...    Won't we fall over all over the place if ApplicationServices is null?  I don't think we are checking anywhere in our code today

davidfowl: I don't know why anybody would set it to null unless they were trying to clear it out. Maybe if the old value was null?

HaoK: Ok well this is a subtle breaking change then, setting to null will throw an exception now...  updated and added a test

src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerFeature.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            get
+            {
+                return _appServices;
+            }
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(ApplicationServices));
```
davidfowl: This should be value not ApplicationServices afaik

davidfowl:
Small change then :shipit: once CI is green

HaoK:
49520a2a73251a9f718743df4a43d26514021767

Issue #77, opened 2014-09-23
Catch selfhost Startup errors and show them in the browser


Tratcher:
When anything goes wrong with Helios' startup code path it will catch those errors, finish starting the server, and send an error page in response to any HTTP requests.

When anything goes wrong with the Self-Host startup code path it will throw an exception and crash the process. If you're lucky the exception will be printed on the console, but even this requires setting the ASPNET_ENV to Development.  This is even harder to debug in AppFabric or Azure Worker roles.

Proposal: Move/duplicate some of Helios' startup logic into the Hosting package. When errors occur locating the startup class/method or executing it, catch those and generate a default error page application instead.

There are only a small class of errors that would prevent the server from actually starting and serving such an error page (address conflicts, compiler errors, etc.), so we should be able to provide the error page for a range of common errors.

davidfowl:
I like this suggestion

muratg:
@Tratcher feel free to move this to RC1.

davidfowl:
This cannot be moved to RC1

PR #370, opened 2015-09-23
Catch startup exceptions and show them in the browser.

Tratcher:
#77 @davidfowl @DamianEdwards @muratg @Eilon 
Catch any application startup related exceptions and display them in the browser rather than on the console. Note this only applies to Hosting.Program.Main scenarios, not to self-hosted WebHostBuilder or to TestHost
Handled exceptions:
- App compilation failures
- Missing Startup class
- Startup's constructor or static constructor throw, or fail to resolve dependencies
- Startup does not have the expected methods.
- ConfigureServices or Configure throws

image

The error details are only displayed if your in Development. Otherwise you get:
image

There is one open question:
Should we attempt any content-negotiation to avoid displaying HTML to non-browser clients?

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Http.Features.Internal;
 using Microsoft.AspNet.Server.Features;
+using Microsoft.Dnx.Runtime;
 using Microsoft.Framework.Configuration;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Logging;
@@ -24,11 +25,13 @@ namespace Microsoft.AspNet.Hosting.Internal
     {
         // This is defined by IIS's HttpPlatformHandler.
         private static readonly string ServerPort = "HTTP_PLATFORM_PORT";
+        private static readonly string DetailedErrors = "Hosting:DetailedErrors";
 
         private readonly IServiceCollection _applicationServiceCollection;
         private readonly IStartupLoader _startupLoader;
         private readonly ApplicationLifetime _applicationLifetime;
         private readonly IConfiguration _config;
+        private readonly bool _captureStartupErrors;
 
         private IServiceProvider _applicationServices;
 
@@ -45,7 +48,8 @@ namespace Microsoft.AspNet.Hosting.Internal
         public HostingEngine(
             IServiceCollection appServices,
             IStartupLoader startupLoader,
-            IConfiguration config)
+            IConfiguration config,
+            bool captureStartupErrors)
         {
             if (appServices == null)
             {
@@ -65,6 +69,7 @@ namespace Microsoft.AspNet.Hosting.Internal
             _config = config;
             _applicationServiceCollection = appServices;
             _startupLoader = startupLoader;
+            _captureStartupErrors = captureStartupErrors;
             _applicationLifetime = new ApplicationLifetime();
         }
 
@@ -79,8 +84,6 @@ namespace Microsoft.AspNet.Hosting.Internal
 
         public virtual IApplication Start()
         {
-            EnsureApplicationServices();
-
             var application = BuildApplication();
 
             var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
@@ -175,6 +178,67 @@ namespace Microsoft.AspNet.Hosting.Internal
 
         private RequestDelegate BuildApplication()
         {
+            try
+            {
+                EnsureApplicationServices();
+                EnsureServer();
+
+                var builderFactory = _applicationServices.GetRequiredService<IApplicationBuilderFactory>();
+                var builder = builderFactory.CreateBuilder(_serverInstance);
+                builder.ApplicationServices = _applicationServices;
+
+                var startupFilters = _applicationServices.GetService<IEnumerable<IStartupFilter>>();
+                var configure = Startup.ConfigureDelegate;
+                foreach (var filter in startupFilters)
+                {
+                    configure = filter.Configure(configure);
+                }
+
+                configure(builder);
+
+                return builder.Build();
+            }
+            catch (Exception ex)
+            {
+                if (!_captureStartupErrors)
+                {
+                    throw;
+                }
+
+                // EnsureApplicationServices may have failed due to a missing or throwing Startup class.
+                if (_applicationServices == null)
+                {
+                    _applicationServices = _applicationServiceCollection.BuildServiceProvider();
+                }
+
+                EnsureServer();
+
+                // Write errors to standard out so they can be retrieved when not in development mode.
+                Console.Out.WriteLine("Application startup exception: " + ex.ToString());
+                var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
+                logger.LogError("Application startup exception", ex);
+
+                // Generate an HTML error page.
+                var runtimeEnv = _applicationServices.GetRequiredService<IRuntimeEnvironment>();
+                var hostingEnv = _applicationServices.GetRequiredService<IHostingEnvironment>();
+                var showDetailedErrors = hostingEnv.IsDevelopment()
+                    || string.Equals("true", _config[DetailedErrors], StringComparison.OrdinalIgnoreCase)
+                    || string.Equals("1", _config[DetailedErrors], StringComparison.OrdinalIgnoreCase);
+                var errorBytes = StartupExceptionPage.GenerateErrorHtml(showDetailedErrors, runtimeEnv, ex);
+
+                return context =>
+                {
+                    context.Response.StatusCode = 500;
+                    context.Response.Headers["Cache-Control"] = "private, max-age=0";
+                    context.Response.ContentType = "text/html; charset=utf-8";
+                    context.Response.ContentLength = errorBytes.Length;
+                    return context.Response.Body.WriteAsync(errorBytes, 0, errorBytes.Length);
+                };
+            }
+        }
+
+        private void EnsureServer()
+        {
             if (ServerFactory == null)
             {
                 // Blow up if we don't have a server set at this point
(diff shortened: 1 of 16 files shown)


DamianEdwards:
So pretty

DamianEdwards:
RE the content-negotiation question: That sounds really nice actually. In the case that the client doesn't send an `text/html` accept, what we return in the body? Nothing at all?

davidfowl:
RE content-negotiation: no :smile:

DamianEdwards:
What did we decide to do RE making this easier to debug in non-development environment? Where does this message go in that case?

DamianEdwards:
Spoke to @Tratcher, decided we should also print this message (sans-HTML) to stdout. With the HttpPlatformHandler config this will go to the file so you can see the error easily in Azure, etc.

davidfowl:
Sweet! We need to figure out if we want to use development as the only switch. We wouldn't want people to change the env to Development do we?

DamianEdwards:
@davidfowl we would not. Are you saying there should be a knob to allow show detailed errors that overrides the development check? Or are you referring to the stdout printing?

davidfowl:
Yep. That horrible one we had before ASPNET_DETAILED_ERRORS=true

DamianEdwards:
Yep sounds good. Can you add that pls @Tratcher?

src/Microsoft.AspNet.Hosting/Startup/StartupErrorUtilities.cs
```diff
@@ -0,0 +1,537 @@
...
+
+            // Generate the footer
+            string footer = showDetails ? GenerateFooterEncoded(runtimeEnvironment) : null;
+
+            // And generate the full markup
+            return Encoding.UTF8.GetBytes(String.Format(CultureInfo.InvariantCulture, _errorPageFormatString, builder, rawExceptionDetails, footer));
+        }
+
+
+#if !DNXCORE50
```
davidfowl: They added StackFrame to CoreCLR https://www.nuget.org/packages/System.Diagnostics.StackTrace/4.0.1-beta-23225

Tratcher:
Updated

src/Microsoft.AspNet.Hosting/Startup/FailSafeStartupLoader.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        internal class FailSafeStartup
+        {
+            private readonly byte[] _errorBytes;
+
+            public FailSafeStartup(bool showDetails, IRuntimeEnvironment env, Exception ex)
+            {
+                // Write errors to standard out so they can be retrieved when not in development mode.
+                Console.Out.WriteLine("Application startup exception: " + ex.ToString());
```
davidfowl: Log as well. One thing we said was that logging is *always* good to do. Console.WriteLine is the thing to question.

Tratcher: This can't log, logging is app configured and probably wasn't set up yet.

davidfowl: You can log. Hosting creates the logger factory.

Tratcher: But the logs don't go anywhere until the app configures them, and the app failed to start.

davidfowl: Don't assume that, we should just log. You can imagine hosting registering an EventLogProvider on windows so that critical exceptions like this show up there.

Tratcher:
Updated.

src/Microsoft.AspNet.Hosting/Startup/FailSafeStartupLoader.cs
```diff
@@ -0,0 +1,168 @@
...
+        private Exception _startupException;
+
+        public FailSafeStartupLoader(IServiceProvider services, IHostingEnvironment hostingEnv, IRuntimeEnvironment runtimeEnv, ILoggerFactory loggerFactory)
+        {
+            _nestedLoader = new StartupLoader(services, hostingEnv);
+            _services = services;
+            _runtimeEnv = runtimeEnv;
+            _loggerFactory = loggerFactory;
+            ShowDetailedErrors = hostingEnv.IsDevelopment()
+                || string.Equals("true", Environment.GetEnvironmentVariable("ASPNET_DETAILED_ERRORS"), StringComparison.OrdinalIgnoreCase);
```
DamianEdwards: `true || 1` ?

src/Microsoft.AspNet.Hosting/Startup/FailSafeStartupLoader.cs
```diff
@@ -0,0 +1,168 @@
...
+using Microsoft.Dnx.Runtime;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.AspNet.Hosting.Startup
+{
+    /// <summary>
+    /// Catches common startup exceptions and shows an error page.
+    /// </summary>
+    public class FailSafeStartupLoader : IStartupLoader
```
lodejard: Don't implement as an IStartupLoader. Instead put `try {} catch {}` in `HostingEngine.BuildApplication` catch should determine if the error should be rethrown or displayed. If displayed it should return a simple delegate - doesn't need to follow Startup-class logic paths.

This also means the call to EnsureApplicationServices in HostingEngine.Start should move down into the BuildApplication method.

TBH, the application-services assigning and telemetry-providing delegate wrapper should be also be moved into the BuildApplication success path instead of the engine Start method.

davidfowl: I'm going to +1, this feels like a larger than needs to be change. If your current tests are covering all of the cases, we can restructure to be a simple try catch in the right places (we can make things happen in a single place). Rather than using the startup loader.

Tratcher:
Redone.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -175,48 +178,96 @@ private void EnsureStartup()
...
                 {
-                    addresses.Add("http://localhost:5000");
+                    _applicationServices = _applicationServiceCollection.BuildServiceProvider();
                 }
+
+                EnsureServer();
+
+                // Write errors to standard out so they can be retrieved when not in development mode.
+                Console.Out.WriteLine("Application startup exception: " + ex.ToString());
+                var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
```
davidfowl: Dont use `Logger<T>` it's hella inefficient (unfortunately) :smile:. Get the LoggerFactory

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -175,48 +178,96 @@ private void EnsureStartup()
...
+                logger.LogError("Application startup exception", ex);
+
+                // Generate an HTML error page.
+                var runtimeEnv = _applicationServices.GetRequiredService<IRuntimeEnvironment>();
+                var hostingEnv = _applicationServices.GetRequiredService<IHostingEnvironment>();
+                var showDetailedErrors = hostingEnv.IsDevelopment()
+                    || string.Equals("true", _config[DetailedErrors], StringComparison.OrdinalIgnoreCase);
```
davidfowl: support true or 1

src/Microsoft.AspNet.Hosting/Startup/StartupExceptionPage.cs
```diff
@@ -0,0 +1,520 @@
...
+        private static void WriteMessage(string message, StringBuilder builder)
+        {
+            // Build the <div>
+            builder.AppendFormat(CultureInfo.InvariantCulture, _errorMessageFormatString,
+                HtmlEncodeAndReplaceLineBreaks(message));
+        }
+
+        private static IEnumerable<Exception> FlattenAndReverseExceptionTree(Exception ex)
+        {
+            List<Exception> list = new List<Exception>();
```
davidfowl: var

davidfowl:
minor changes, this change looks *much* better

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
rynowak: Should we just delete that?

davidfowl: Its like LINQ, nice to use in some cases but mostly easy to do something in efficient. It's worth trying to see why it's so slow. TBH tho, this code path is error path so it doesn't matter that much.

davidfowl: Plus you'll make @DamianEdwards cry, but thats ok

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
Tratcher: make up your mind.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
Tratcher: This executes once per app and only in a catastrophic failure scenario, forget perf.

davidfowl: ya

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
davidfowl: I meant the or statement :smile: Support both 1 and true

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
DamianEdwards: What happened to my beloved `ILogger<T>`?!?! :sob:

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
DamianEdwards: Yeah I'd said that earlier, think it got lost in the latest diff.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
muratg: I say keep ILogger<T> since perf is not that important in the failure case.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
muratg: Do we want to have standardization across the stack for boolean env variables?  For example for `DNX_TRACE` we seem to have 0 or 1 only.

Tratcher: Exactly, if we're not consistent about this sort of thing we're just going to confuse people. On that note the name should probably be `Hosting:DetailedErrors` like the other hosting keys.

Tratcher: Exactly, if we're not consistent about this sort of thing we're just going to confuse people. On that note the name should probably be `Hosting:DetailedErrors` like the other hosting keys.

davidfowl: DNX_TRACE is a bad comparison. That's potentially a level, all the other boolean flags in DNX support true or 1.

davidfowl: on unix Hosting:DetailedErrors has to be set has `Hosting__DetailedErrors` as `:` isn't allowed in env vars.

Tratcher:
Updated.

Tratcher:
*ping* @davidfowl

davidfowl:
:shipit:

davidfowl:
Thought you already merged this change

davidfowl:
:smile:

PR #375, opened 2015-09-26
Proposal: Start log scope only if logging is enabled

### LINKED ISSUE #367: [Perf] High Allocator LoggingScope when no loggers
Using works with null references, so this is simpler than outlined below; added change to logging to improve efficiency.

---
Creating a scope is expensive (at benchmarking speeds) and the techempower benchmarks specify in [general requirements](https://www.techempower.com/benchmarks/#section=code) item 2:

> All test implementations **must disable all disk logging**. For many reasons, we expect all tests will run without writing logs to disk. Most importantly, the volume of requests is sufficiently high to fill up disks even with only a single line written to disk per request. Please disable all forms of disk logging. We recommend but do not require disabling console logging as well.

For fair comparison, only create the logging scope if logging is enabled:

Add ```LogLevel``` of ```None = 7```? 

Change the following from:

```csharp
var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
...
    async features =>
    {
        try
        {
            using (logger.BeginScope("Request Id: {RequestId}", requestIdentifier))
            {
            }
        }
        catch (Exception ex)
        {
            if (telemetrySource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
            {
                telemetrySource.WriteTelemetry("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
            }

            throw;
        }
```

To something similar to:

```csharp
var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
...
var isLogging = logger.IsEnabled(LogLevel.Critical); // ? not sure about how not logging is checked
    async features =>
    {
        ILogger loggerScope;
        try
        {
            if (isLogging) 
            {
                loggerScope = logger.BeginScope("Request Id: {RequestId}", requestIdentifier);
            }

        }
        catch (Exception ex)
        {
            if (telemetrySource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
            {
                telemetrySource.WriteTelemetry("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
            }

            throw;
        }
        finally
        {
            if (loggerScope != null) loggerScope.Dispose();
        }
```

Ensure having no local logging scope doesn't cause issues.

benaadams:
Start log scope only if logging is enabled at some level

Resolves #367

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -104,12 +104,7 @@ namespace Microsoft.AspNet.Hosting.Internal
 
                     try
                     {
-                        using (logger.IsEnabled(LogLevel.Critical) ||
-                               logger.IsEnabled(LogLevel.Error) ||
-                               logger.IsEnabled(LogLevel.Warning) ||
-                               logger.IsEnabled(LogLevel.Information) ||
-                               logger.IsEnabled(LogLevel.Verbose) ||
-                               logger.IsEnabled(LogLevel.Debug)
+                        using (logger.IsEnabled(LogLevel.Critical)
                             ? logger.BeginScope("Request Id: {RequestId}", requestIdentifier) 
                             : null)
                         {
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -95,7 +95,6 @@ namespace Microsoft.AspNet.Hosting.Internal
                 {
                     var httpContext = contextFactory.CreateHttpContext(features);
                     httpContext.ApplicationServices = _applicationServices;
-                    var requestIdentifier = GetRequestIdentifier(httpContext);
 
                     if (telemetrySource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
                     {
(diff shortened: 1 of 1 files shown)


davidfowl:
This doesn't look right. Scopes and IsEnabled don't really overlap /cc @lodejard

benaadams:
For scoping disposal the ```RequestServicesContainerMiddleware``` creates a ```RequestServices``` scope as first step of ```await application(httpContext);```

If there is no logging is there a logging scope? :see_no_evil: :hear_no_evil: :speak_no_evil:

benaadams:
@davidfowl related https://github.com/aspnet/Logging/pull/254 though that doesn't solve it alone and can't get away structs; as it would do a lot of boxing with paramsArray and ```ILogger.BeginScopeImpl(object state)``` so this just short cruits the whole path.

davidfowl:
@lodejard Actually has a nice fix coming for logger in performance critical scenarios (0 allocations)

benaadams:
Performing correct test - Yuk

benaadams:
Still cuts it to zero alloc for the logging scope though is performing 6 ```IsEnabled``` tests rather than the 1 ```IsDisabled``` type test I was hoping for.

logging-scope

benaadams:
You could cut the CPU work a little more by moving ```GetRequestIdentifier(httpContext)``` to be the last parameter of ```BeginScope``` though I think you want to put it onto HttpContext so that wouldn't work - although it could be a lazy eval'd property; and still the last param.

lodejard:
Nice! You should also be able to test for the most restrictive, Critical, only... If any of the others are true that will be true as well (the severity is a level rather than a bitmask)

benaadams:
@lodejard Ah, did that first, will revert. Also see https://github.com/aspnet/Logging/pull/254 which will make the tests faster.

Will still need changes to improve individual Log calls though; but you may be working on that?

davidfowl:
I think we want to take this. @lodejard can you think of anything this breaks?

lodejard:
:shipit:

PR #377, opened 2015-09-28
Set ContextAccessor before the BeginRequest event runs

avanderhoorn:
Tools like Glimpse are expecting to have access to the ContextAccessor to determine if profiling should be turns on or not. The contextAccessor is key to being able to determine this and based of the code flow, it shouldn't have any side effect.


Tratcher:
:shipit: 
@muratg @davidfowl Consider for beta8?

davidfowl:
:shipit:

Tratcher:
Rebased and merged.

avanderhoorn:
THANKS!!!

PR #380, opened 2015-10-02
Sanitise loops

benaadams:
Use jit recognised ```for (var i; i < a.length; i++)``` for bounds check elimination.


Tratcher:
:shipit:

src/Microsoft.AspNet.Hosting/Startup/ConfigureDelegate.cs
```diff
@@ -27,7 +27,7 @@ private void Invoke(object instance, IApplicationBuilder builder)
             var serviceProvider = builder.ApplicationServices;
             var parameterInfos = MethodInfo.GetParameters();
             var parameters = new object[parameterInfos.Length];
-            for (var index = 0; index != parameterInfos.Length; ++index)
+            for (var index = 0; index < parameterInfos.Length; index++)
```
Eilon: YESSSSSSSSS!!!!!!!!!!!!!!!!!!!!!!!!

Issue #310, opened 2015-07-28
[Perf] HttpContextFactory should only wrap ReadOnly feature collections


Tratcher:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs#L14

IFeatureCollections can be marked as IsReadOnly or not. HttpContextFactory currently assumes they are all read only and wraps them in a mutable FeatureCollection. This has some perf overhead where it creates extra Dictionaries, etc, and is redundant if the server already provided a mutable collection.

Recommendation: Only wrap the given feature collection if is marked as IsReadOnly.

Tratcher:
Nevermind, FeatureCollection has made significant improvements in this scenario.

Tratcher:
@lodejard @benaadams

benaadams:
I did have a look at ```FeatureCollection``` when checking the codebase and *in isolation* its pretty efficient at what it does; so couldn't come up with any improvements.

Taking a step back there are FeatureCollection improvements coming in Kestrel; which regardless of metrics, have the distinct allocation advantage of already being created, and this wrapping would prevent it being utilised passed the point the context is created.

However, I don't have the holistic view on what circumstances; if any, ```IFeatureCollection``` could be readonly, and whether  ```FeatureCollection``` is needed at all; but if it is, you could make the one line change from:
```csharp
return new DefaultHttpContext(new FeatureCollection(featureCollection));
```
to (in preferred line break formatting) 
```csharp
return new DefaultHttpContext(
    featureCollection.IsReadOnly 
    ? new FeatureCollection(featureCollection) 
    : featureCollection);
```
which would work and allow the changes in Kestrel's collection to flow through. So that would be an easy fix?

Then open question on when would the ```IFeatureCollection``` be readonly?

Tratcher:
Allowing the IFeatureCollection to be read-only can make for simpler server implementations. FeatureCollection makes up the difference.

Or we cut that and just ask the simple servers to use FeatureCollection directly.

benaadams:
```FeatureCollection``` is a very interesting collection; especially when you consider you can pass an ```FeatureCollection``` to its constructor; boom you've got javascript's prototype inheritance right there; change the original and it flows through if its not overridden at some stage. I digress...

> Allowing the IFeatureCollection to be read-only can make for simpler server implementations. 

[...]

PR #381, opened 2015-10-04
Only wrap readonly IFeatureCollections

benaadams:
Perf improvement suggested by @Tratcher in https://github.com/aspnet/Hosting/issues/310 also needed to full take advantage of Kestrel's feature collection improvements coming down the pipe.

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
+++ b/src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
@@ -11,7 +11,10 @@ namespace Microsoft.AspNet.Hosting.Builder
     {
         public HttpContext CreateHttpContext(IFeatureCollection featureCollection)
         {
-            return new DefaultHttpContext(new FeatureCollection(featureCollection));
+            return new DefaultHttpContext(
+                            featureCollection.IsReadOnly 
+                            ? new FeatureCollection(featureCollection) 
+                            : featureCollection);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Tratcher:
Ok @benaadams, you can remove the wrapper completely.

benaadams:
Done

benaadams:
I need to change some tests?

benaadams:
Uncomfortable with the test change

davidfowl:
We need to fix the owing feature collection

benaadams:
There is a lot of angry in the tests, trying to get to bottom of it, but can't run them locally

benaadams:
Highlights for items where entire point of test is now bypassed:

https://github.com/benaadams/Hosting/blob/only-wrap-readonly-features/test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs#L322

https://github.com/benaadams/Hosting/blob/only-wrap-readonly-features/test/Microsoft.AspNet.Hosting.Tests/HttpContextFactoryFacts.cs#L19

benaadams:
They are bypassed by wrapping with ```FeatureCollection``` so may be desired outcome?

davidfowl:
@Tratcher assigning this one to you. We should merge this in when the servers support it. We probably need to update the OwinFeatureCollection as well?

Tratcher:
Done. Filed a separate issue for the OwinFeatureCollection: https://github.com/aspnet/HttpAbstractions/issues/431

Tratcher:
FYI: The servers already supported it.

borgdylan:
As things stand owinfeaturecollection is still read-only yet it is no longer being wrapped. This makes all requests fail.

Issue #306, opened 2015-07-22
Don't use guid in GetRequestIdentifier


halter73:
Guid.NewGuid and Guid.ToString are a little slow. Using a incrementing long instead with an random starting seed will likely be faster.

We should also look at the most efficient way to generate a string from this long.

![profile](https://cloud.githubusercontent.com/assets/54385/8839344/8206c8a4-3089-11e5-8c85-2bc35e80841e.jpg)

Eilon:
We've had a lot of discussion before on `Guid` vs. other types of ids. Could we still use a Guid but generate it ourselves? Just thinking out loud here.

Tratcher:
@halter73's description sounds like what System.Web did previously.

You could also make Kestrel generate it directly using something like the socket handle and a request counter.

PR #384, opened 2015-10-05
Don't use GUID in GetRequestIdentifier

DamianEdwards:
Lazily increments a long and formats it to a HEX string. Results in SmurfLab show a performance increase of ~10% in the plaintext benchmark.

#306


src/Microsoft.AspNet.Hosting/Internal/FastHttpRequestIdentifierFeature.cs
```diff
@@ -0,0 +1,60 @@
...
+                }
+                return _id;
+            }
+            set
+            {
+                _id = value;
+            }
+        }
+
+        private string GenerateRequestId(long id)
```
DamianEdwards: Meant to make this `static`, will fix.

davidfowl:
:shipit:

PR #385, opened 2015-10-05
requestId ticks seed, reduced GenerateRequestId allocs

benaadams:
Seed requestId with DateTime.UtcNow.Ticks rather than Random
Reduced allocs in GenerateRequestId

/cc @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/FastHttpRequestIdentifierFeature.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/FastHttpRequestIdentifierFeature.cs
@@ -9,9 +9,12 @@ namespace Microsoft.AspNet.Hosting.Internal
 {
     public class FastHttpRequestIdentifierFeature : IHttpRequestIdentifierFeature
     {
-        private static readonly string _hexChars = "0123456789ABCDEF";
-        // Seed the _requestId for this application instance with a random int
-        private static long _requestId = new Random().Next();
+        // Base64 encoding - but in ascii sort order for easy text based sorting 
+        private static readonly string _encode32Chars = "0123456789ABCDEFGHIJKLMNOPQRSTUV";
+        // Seed the _requestId for this application instance with 
+        // the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001
+        // for a roughly increasing _requestId over restarts
+        private static long _requestId = DateTime.UtcNow.Ticks;
         
         private string _id = null;
         
@@ -32,29 +35,31 @@ namespace Microsoft.AspNet.Hosting.Internal
             }
         }
 
-        private static string GenerateRequestId(long id)
+        private static unsafe string GenerateRequestId(long id)
         {
-            // The following routine is ~33% faster than calling long.ToString() when testing in tight loops of
-            // 1 million iterations.
-            var charBuffer = new char[sizeof(long) * 2];
-            charBuffer[0] = _hexChars[(int)(id >> 60) & 0x0f];
-            charBuffer[1] = _hexChars[(int)(id >> 56) & 0x0f];
-            charBuffer[2] = _hexChars[(int)(id >> 52) & 0x0f];
-            charBuffer[3] = _hexChars[(int)(id >> 48) & 0x0f];
-            charBuffer[4] = _hexChars[(int)(id >> 44) & 0x0f];
-            charBuffer[5] = _hexChars[(int)(id >> 40) & 0x0f];
-            charBuffer[6] = _hexChars[(int)(id >> 36) & 0x0f];
-            charBuffer[7] = _hexChars[(int)(id >> 32) & 0x0f];
-            charBuffer[8] = _hexChars[(int)(id >> 28) & 0x0f];
-            charBuffer[9] = _hexChars[(int)(id >> 24) & 0x0f];
-            charBuffer[10] = _hexChars[(int)(id >> 20) & 0x0f];
-            charBuffer[11] = _hexChars[(int)(id >> 16) & 0x0f];
-            charBuffer[12] = _hexChars[(int)(id >> 12) & 0x0f];
-            charBuffer[13] = _hexChars[(int)(id >> 8) & 0x0f];
-            charBuffer[14] = _hexChars[(int)(id >> 4) & 0x0f];
-            charBuffer[15] = _hexChars[(int)(id >> 0) & 0x0f];
+            // The following routine is ~310% faster than calling long.ToString() on x64
+            // and ~600% faster than calling long.ToString() on x86 in tight loops of 1 million+ iterations
+            // See: https://github.com/aspnet/Hosting/pull/385
 
-            return new string(charBuffer);
+            // stackalloc to allocate array on stack rather than heap
+            char* charBuffer = stackalloc char[13];
+
+            charBuffer[0] = _encode32Chars[(int)(id >> 60) & 31];
+            charBuffer[1] = _encode32Chars[(int)(id >> 55) & 31];
+            charBuffer[2] = _encode32Chars[(int)(id >> 50) & 31];
+            charBuffer[3] = _encode32Chars[(int)(id >> 45) & 31];
+            charBuffer[4] = _encode32Chars[(int)(id >> 40) & 31];
+            charBuffer[5] = _encode32Chars[(int)(id >> 35) & 31];
+            charBuffer[6] = _encode32Chars[(int)(id >> 30) & 31];
+            charBuffer[7] = _encode32Chars[(int)(id >> 25) & 31];
+            charBuffer[8] = _encode32Chars[(int)(id >> 20) & 31];
+            charBuffer[9] = _encode32Chars[(int)(id >> 15) & 31];
+            charBuffer[10] = _encode32Chars[(int)(id >> 10) & 31];
+            charBuffer[11] = _encode32Chars[(int)(id >> 5) & 31];
+            charBuffer[12] = _encode32Chars[(int)id & 31];
+
+            // string ctor overload that takes char*
+            return new string(charBuffer, 0, 13);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


benaadams:
```ini
BenchmarkDotNet=v0.7.8.0
OS=Microsoft Windows NT 6.2.9200.0
Processor=Intel(R) Core(TM) i7-4720HQ CPU @ 2.60GHz, ProcessorCount=8
HostCLR=MS.NET 4.0.30319.42000, Arch=32-bit
Type=LongToString  Mode=Throughput  .NET=HostFramework
```

                     Method | Platform |       Jit |    AvrTime |    StdDev |           op/s |
--------------------------- |--------- |---------- |----------- |---------- |--------------- |
LongToString |      X86 | LegacyJit | 48.6370 ns | 0.5701 ns |  20,560,474.66 |
RequestIdCharHeap |      X86 | LegacyJit |  7.9431 ns | 0.1111 ns | 125,896,043.99 |
RequestIdCharStack |      X86 | LegacyJit |  9.6136 ns | 0.1804 ns | 104,019,793.55 |
LongToString |      X64 |    RyuJit | 21.9177 ns | 0.7149 ns |  45,625,348.42 |
RequestIdCharHeap |      X64 |    RyuJit |  7.7770 ns | 0.0258 ns | 128,584,042.38 |
RequestIdCharStack |      X64 |    RyuJit |  7.2523 ns | 0.0137 ns | 137,887,101.48 |

Allocations aside, is a little slower on x86, but faster on x64 and x64 is faster than x86

src/Microsoft.AspNet.Hosting/Internal/FastHttpRequestIdentifierFeature.cs
```diff
@@ -10,8 +10,10 @@ namespace Microsoft.AspNet.Hosting.Internal
     public class FastHttpRequestIdentifierFeature : IHttpRequestIdentifierFeature
     {
         private static readonly string _hexChars = "0123456789ABCDEF";
-        // Seed the _requestId for this application instance with a random int
-        private static long _requestId = new Random().Next();
+        // Seed the _requestId for this application instance with 
+        // the number of 100-nanosecond intervals that have elapsed since 12:00:00 midnight, January 1, 0001
+        // for a roughtly increasing _requestId over restarts
```
davidfowl: roughtly -> roughly

stephentoub:
LGTM.

Is performance here really critical?  If so, there are other things that could be done to cut down on memory usage further:
- If instead of using DateTime.UtcNow.Ticks you were to use Environment.TickCount (which is what Random uses internally), then the majority of requests would have a relatively small ID, e.g. the hex encoding of Environment.TickCount on my machine right now is 62EF26 whereas for DateTime.UtcNow.Ticks it's 8D2CDAE45A15844.  If you were to go with the smaller value, and if you were to size the resulting string based on the actual number of characters used (as long.ToString does by default), then in this example most of the allocated strings would be only 6-characters long rather than 15-characters long.
- If instead of hex you were to use a larger set of characters, e.g. 64 of them rather than 16 of them, you could encode 6 bits per character instead of only 4 bits per character, and save a few additional characters that way (assuming you were doing the above change as well that limited the length of the string to how many non-zero characters there were).

benaadams:
> If instead of hex you were to use a larger set of characters, e.g. 64 of them rather than 16 of them

That trumps them all!

                     Method | Platform |       Jit |    AvrTime |    StdDev |           op/s |
--------------------------- |--------- |---------- |----------- |---------- |--------------- |
LongToString |      X86 | LegacyJit | 48.6370 ns | 0.5701 ns |  20,560,474.66 |
RequestIdCharHeap |      X86 | LegacyJit |  7.9431 ns | 0.1111 ns | 125,896,043.99 |
RequestIdCharStack |      X86 | LegacyJit |  9.6136 ns | 0.1804 ns | 104,019,793.55 |
RequestIdChar64Stack |      X86 | LegacyJit | 7.4285 ns | 0.0076 ns | 134,617,091.36 |
LongToString |      X64 |    RyuJit | 21.9177 ns | 0.7149 ns |  45,625,348.42 |
RequestIdCharHeap |      X64 |    RyuJit |  7.7770 ns | 0.0258 ns | 128,584,042.38 |
RequestIdCharStack |      X64 |    RyuJit |  7.2523 ns | 0.0137 ns | 137,887,101.48 |
RequestIdChar64Stack |      X64 |    RyuJit | 6.0165 ns | 0.0120 ns | 166,209,448.85 |

```Environment.TickCount``` I'm not sure about as it wraps every 50 days; and this value will be like used in tracing and logging; so with a ```DateTime.UtcNow.Ticks``` based start you can have a correlated event stream spanning years; but ```Environment.TickCount``` would give a shorter period, would would work for DevOps and diagnostic correlation; but not necessarily for big data (though a timestamp is probably thrown in also so its probably not too important).

The full string also gives simple alpha sort (case sensitive) to get all the events in order; but again not sure is important as will likely be a timestamp somewhere? If not can go with the standard base64 set.

benaadams:
However using ```Environment.TickCount``` and an ```int``` based ```RequestId``` (6 char without string shrinking) is a very significant boost

                     Method | Platform |       Jit |    AvrTime |    StdDev |           op/s |
--------------------------- |--------- |---------- |----------- |---------- |--------------- |
LongToString |      X86 | LegacyJit | 48.6370 ns | 0.5701 ns |  20,560,474.66 |
RequestIdCharHeap |      X86 | LegacyJit |  7.9431 ns | 0.1111 ns | 125,896,043.99 |
RequestIdCharStack |      X86 | LegacyJit |  9.6136 ns | 0.1804 ns | 104,019,793.55 |
RequestIdChar64Stack |      X86 | LegacyJit | 7.4285 ns | 0.0076 ns | 134,617,091.36 |
**IntChar64Stack** |      X86 | LegacyJit | 5.9278 ns | 0.0069 ns | 168,697,749.18 |
LongToString |      X64 |    RyuJit | 21.9177 ns | 0.7149 ns |  45,625,348.42 |
RequestIdCharHeap |      X64 |    RyuJit |  7.7770 ns | 0.0258 ns | 128,584,042.38 |
RequestIdCharStack |      X64 |    RyuJit |  7.2523 ns | 0.0137 ns | 137,887,101.48 |
RequestIdChar64Stack |      X64 |    RyuJit | 6.0165 ns | 0.0120 ns | 166,209,448.85 |
**IntChar64Stack** |      X64 |    RyuJit | 4.2964 ns | 0.0117 ns | 232,752,905.73 |

benaadams:
Would need call on whether to do any of
* Move ```int``` based with ```Environment.TickCount```
* Size the array to remove "leading zeros" rather than fully filled
* Change any of ```_encode64Chars``` character choices

bbowyersmyth:
If this is performance critical, seeing it is a string creation method you could do the unsafe index assignment on a pre-sized string itself and ditch the temp buffer and copy. Don't know this repo stance on such things though.

benaadams:
@bbowyersmyth String looks like its implemented in clr native and exposes no accepted way to do this so it is crossing an implementation detail boundary and moving outside the managed implementation; which may be a step to far? Its further than reflecting over private members which is quite dodgy.

benaadams:
@bbowyersmyth maybe something like ```LockBits``` that exists on ```System.Drawing.Bitmap```

bbowyersmyth:
@benaadams It may be a step to far but I though I'd mention it if this is still on the hot path. Something as simple as duplicating data can have an impact on the RPS. All good.

benaadams:
@bbowyersmyth zero alloc, zero copy are my watch words ;-)

DamianEdwards:
This certainly isn't on the hot path anymore. I think that the change that is currently in this PR is a worth step, to reduce the heap allocations but anything beyond that is most likely overkill right now (or at least until this shows up in our profile again :smile:).

benaadams:
need a call on the base64 set though; won't change perf

DamianEdwards:
@benaadams base64 implies case sensitivity yes? I don't really want these IDs to be case sensitive. Perhaps base36 is enough.

benaadams:
@DamianEdwards base32 would be easier for a clean 5 bit; shave off a couple letter from the end?

DamianEdwards:
Sure

benaadams:
Done. ~310% faster than calling long.ToString() on x64 and ~600% faster on x32

                     Method | Platform |       Jit |    AvrTime |    StdDev |           op/s |
--------------------------- |--------- |---------- |----------- |---------- |--------------- |
LongToString |      X86 | LegacyJit | 48.6370 ns | 0.5701 ns |  20,560,474.66 |
CharHeap |      X86 | LegacyJit |  7.9431 ns | 0.1111 ns | 125,896,043.99 |
CharStack |      X86 | LegacyJit |  9.6136 ns | 0.1804 ns | 104,019,793.55 |
**Char32Stack**  |      X86 | LegacyJit | 8.1148 ns | 0.0441 ns | **123,231,165.02** |
Char64Stack |      X86 | LegacyJit | 7.4285 ns | 0.0076 ns | 134,617,091.36 |
IntChar64Stack |      X86 | LegacyJit | 5.9278 ns | 0.0069 ns | 168,697,749.18 |
LongToString |      X64 |    RyuJit | 21.9177 ns | 0.7149 ns |  45,625,348.42 |
CharHeap |      X64 |    RyuJit |  7.7770 ns | 0.0258 ns | 128,584,042.38 |
CharStack |      X64 |    RyuJit |  7.2523 ns | 0.0137 ns | 137,887,101.48 |
**Char32Stack** |      X64 |    RyuJit | 6.9184 ns | 0.0076 ns | **144,542,881.19** |
Char64Stack |      X64 |    RyuJit | 6.0165 ns | 0.0120 ns | 166,209,448.85 |
IntChar64Stack |      X64 |    RyuJit | 4.2964 ns | 0.0117 ns | 232,752,905.73 |

benaadams:
Back of paper calc says can reduce it from  13 bytes to 12 bytes but cutting the representable range to: from now (or open sourcing of .net; or start of asp.net 5) to 365 years from now rather than +/- 9999 years

DamianEdwards:
Looks good to me. @davidfowl please review

davidfowl:
Looks good!

davidfowl:
@benaadams can you squash your commits?

benaadams:
Done

PR #396, opened 2015-10-11
Late init-requestId

### LINKED ISSUE #392: HostingEngine.Start shouldn't get request ID unless it begins a scope
Generating request ID is expensive so `HostingEngine.Start` shouldn't be grabbing the request ID from the feature collection unless it's actually going to call `BeginScope`.

Effectively we need to merge line 99 into line 110.

by @DamianEdwards

benaadams:
Also resolves #392 but breaks test ```HostingEngine_CreatesDefaultRequestIdentifierFeature_IfNotPresent``` which  aspnet/HttpAbstractions#435 may fix?


Tratcher:
@DamianEdwards @lodejard This isn't going to work soon because you were talking about reverting the `logger.IsEnabled(LogLevel.Critical)` check, correct?

benaadams:
Closing this as better PRs out there.

PR #399, opened 2015-10-11
pass command line args in ApplicationEnvironment

grahamehorner:
allow commandline args to be passed in application environment

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Program.cs
+++ b/src/Microsoft.AspNet.Hosting/Program.cs
@@ -29,7 +29,9 @@ namespace Microsoft.AspNet.Hosting
             var tempConfig = tempBuilder.Build();
             var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
 
-            var appBasePath = _serviceProvider.GetRequiredService<IApplicationEnvironment>().ApplicationBasePath;
+            var appEnv = _serviceProvider.GetRequiredService<IApplicationEnvironment>();
+            appEnv.SetData("CommandLineArguments",args);
+            var appBasePath = appEnv.ApplicationBasePath;
             var config = new ConfigurationBuilder()
                 .SetBasePath(appBasePath)
                 .AddJsonFile(configFilePath, optional: true)


src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -29,7 +29,9 @@ public void Main(string[] args)
             var tempConfig = tempBuilder.Build();
             var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
 
-            var appBasePath = _serviceProvider.GetRequiredService<IApplicationEnvironment>().ApplicationBasePath;
+            var appEnv = _serviceProvider.GetRequiredService<IApplicationEnvironment>();
+            appEnv.SetData("CommandLineArguments",args);
```
davidfowl: Why not expose something a bit more first class? Like a new interface to get the extra arguments? Also, do you want all arguments or unconsumed arguments?

grahamehorner: I can see a need for the needing to know about all of the command line arguments supplied via dnx/hosting to be accessible by the web application; with regards to making the command line arguments a bit more first class; I had played with this idea but opted for a simple string[] so the web application could parse the string[] however it liked and not be constrained to any implementation that was put in place.

davidfowl: I don't like it. If we were going to expose the command line arguments passed into hosting, it would need to be something a bit more first class. Using the SetData, GetData is too much of a hack IMO.

grahamehorner: If the SetData/GetData is a hack and not to be used; it should be marked obsolete; in my opinion otherwise developers will use it.

grahamehorner: so what you thoughts be if I created a IApplicationCommandLineConfiguration interface which is derived from IConfigurationRoot and used (IApplicationCommandLineConfiguration) new ConfigurationBuilder( ).AddCommandLine(args) to build an instance; which I then registered with service provider via the WebHostBuilder by using :--

            hostBuilder.UseServices(cnfgService =>
            {
                cnfgService.AddSingleton<IApplicationCommandLine>(_commandlineConfig);
            });

davidfowl: Yes, something like this in hosting would be better. It might require some changes to the config system if you want to get unconsumed command line options.

PR #403, opened 2015-10-13
Reduce logging overhead in hosting

lodejard:
* TraceIdentifier is done at the last moment, or not at all
* Request starting and finished messages are added
* This pair provide many of the top-level values that you would have
  found in server log files.


src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            private readonly HttpContext _httpContext;
+
+            private FeatureReference<IHttpRequestIdentifierFeature> _requestIdentifierFeatureReference;
+            private IEnumerable<KeyValuePair<string, object>> _cachedGetValues;
+            private string _cachedToString;
+
+            public HostingRequestScope(HttpContext httpContext)
+            {
+                this._httpContext = httpContext;
```
Tratcher: nit: remove `this.`

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            public IHttpRequestIdentifierFeature RequestIdFeature =>
+                _requestIdentifierFeatureReference.Fetch(_httpContext.Features) ??
+                _requestIdentifierFeatureReference.Update(_httpContext.Features, new FastHttpRequestIdentifierFeature());
+
+            public override string ToString() => _cachedToString ?? Interlocked.CompareExchange(
+                ref _cachedToString,
+                $"RequestId:{RequestIdFeature.TraceIdentifier} RequestPath:{_httpContext.Request.Path}",
```
Tratcher: Why just the request path? Why not PathBase, query, etc?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+            public override string ToString() => _cachedToString ?? Interlocked.CompareExchange(
+                ref _cachedToString,
+                $"Request starting {_httpContext.Request.Protocol} {_httpContext.Request.Method} {_httpContext.Request.Scheme}://{_httpContext.Request.Host}{_httpContext.Request.PathBase}{_httpContext.Request.Path}{_httpContext.Request.QueryString} {_httpContext.Request.ContentType} {_httpContext.Request.ContentLength}",
+                null);
+
+            public IEnumerable<KeyValuePair<string, object>> GetValues() => _cachedGetValues ?? Interlocked.CompareExchange(
+                ref _cachedGetValues,
+                new[]
+                {
+                    new KeyValuePair<string, object>("EventName", "RequestStarting"),
```
Tratcher: What is `EventName`? Why didn't you set this for HostingRequestScope?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+            private string _cachedToString;
+
+            public HostingRequestFinished(HttpContext httpContext)
+            {
+                _httpContext = httpContext;
+            }
+
+            public override string ToString() => _cachedToString ?? Interlocked.CompareExchange(
+                ref _cachedToString,
+                $"Request finished {_httpContext.Response.StatusCode} {_httpContext.Response.ContentType} {_httpContext.Response.Body.Length}",
```
Tratcher: You can't assume Length is supported.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+            public override string ToString() => _cachedToString ?? Interlocked.CompareExchange(
+                ref _cachedToString,
+                $"RequestId:{RequestIdFeature.TraceIdentifier} RequestPath:{_httpContext.Request.Path}",
+                null);
+
+            public IEnumerable<KeyValuePair<string, object>> GetValues() => _cachedGetValues ?? Interlocked.CompareExchange(
+                ref _cachedGetValues,
+                new[]
+                {
+                    new KeyValuePair<string, object>("RequestId", RequestIdFeature.TraceIdentifier),
```
Tratcher: Should TraceIdentifier also be included in the other events to facilitate correlation?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+        public static IDisposable RequestScope(this ILogger logger, HttpContext httpContext)
+        {
+            return logger.BeginScopeImpl(new HostingRequestScope(httpContext));
+        }
+
+        public static void RequestStarting(this ILogger logger, HttpContext httpContext)
+        {
+            if (logger.IsEnabled(LogLevel.Information))
+            {
+                logger.Log(LogLevel.Information, 1, new HostingRequestStarting(httpContext), null, HostingRequestStarting.Callback);
```
davidfowl: null and 1 should use the optional args syntax.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+    internal static class HostingLoggerExtensions
+    {
+        public static IDisposable RequestScope(this ILogger logger, HttpContext httpContext)
+        {
+            return logger.BeginScopeImpl(new HostingRequestScope(httpContext));
+        }
+
+        public static void RequestStarting(this ILogger logger, HttpContext httpContext)
+        {
+            if (logger.IsEnabled(LogLevel.Information))
```
Tratcher: This is going to be very noisy. Is there a downside to making this Verbose?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
Tratcher: Or do you depend on the scope to manage that? Will all providers support scopes?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
Tratcher: Or, why bother providing any information at all?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
Tratcher: E.g. this will throw on WebListener.

Tratcher: It wouldn't account for SendFile calls either. I recommend just displaying Content-Length like you do on the request.

davidfowl:
This ci broke

lodejard:
See also aspnet/Logging#261

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: :ok:

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: Just a question of what properties do we want to have as scope decoration on all requests. The TraceIdentifier is fantastic for a unique value, but the Path is second-most-invaluable for knowing what context an event was related to. BasePath being a constant never changes, and QueryString is too variable and less associated with knowing what the context was.

lodejard: For "why bother at all" it's that it's saving the manual step of taking a request id and scrolling back in the logs for the matching record that says what it connected to. It's a scenario specific value judgement.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                logger.Log(LogLevel.Information, 2, new HostingRequestFinished(httpContext), null, HostingRequestFinished.Callback);
+            }
+        }
+
+        private class HostingRequestScope : ILogValues
+        {
+            private readonly HttpContext _httpContext;
+
+            private FeatureReference<IHttpRequestIdentifierFeature> _requestIdentifierFeatureReference;
```
benaadams: With https://github.com/aspnet/HttpAbstractions/pull/440 merged, remove

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,135 @@
...
+            private IEnumerable<KeyValuePair<string, object>> _cachedGetValues;
+            private string _cachedToString;
+
+            public HostingRequestScope(HttpContext httpContext)
+            {
+                this._httpContext = httpContext;
+            }
+
+            public IHttpRequestIdentifierFeature RequestIdFeature =>
```
benaadams: remove

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: Scopes are not log messages, they're one or two properties added as decoration to log messages inside their active range. So adding the same EventName to every message in the scope's range isn't good.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
benaadams: Change ```RequestIdFeature.TraceIdentifier```to ```_httpContext.TraceIdentifier```

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
benaadams: Change ```RequestIdFeature.TraceIdentifier```to ```_httpContext.TraceIdentifier```

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: Content-Length is less frequently assigned. Though now that you mention it SendFiles added as a feature would work because it calls through the Write methods. We can discuss what the alternatives to Stream.Length are.

This was coming from looking at some of the things that are traditionally logged from iis. sc-bytes and sc-bytes in this case. See also http://www.nsi.bg/nrnm/Help/iisHelp/iis/htm/core/iiintlg.htm

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: Exactly right - that's what scopes are for - that way the trace identifier can be used to correlate even on messages that don't include it as their own property. And yeah, the application of scope properties and their correlation is totally provider-specific.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: :ok:

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: These two messages part of the "ideal ten-ish" messages @DamianEdwards and friends called out as visible from a default web application configuration

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
Tratcher: This maps roughly to bytes-sent which can be useful information, but it's hard for Hosting to track that information in detail. Do servers need a stats Feature where they can report bytes received, sent, time-taken, etc.?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: Nice! Hmm... Why don't I see it yet? I'm using `1.0.0-rc1-15826`... Must still be working its way through the system

benaadams: Will merge back into other PR, and add comment changes - hopefully it will flow though soon :)

benaadams: Not sure, updated #409 with all the changes but isn't picking up ```TraceIdentifier``` on ```HttpContext```yet :(

Tratcher: Coherence hasn't passed, but you can reference https://www.myget.org/F/aspnetvolatiledev/api/v2 instead

benaadams: Was test break, applied fix; build is now happy against aspnetvolatiledev for other PR

benaadams: Is working now! :dancer:

benaadams:
PR #409 is working now; which you'll need to merge as the HttpAbstraction changes look like they have flowed through; so currently this PR will break the CI. Added all my comment changes from this to it - as they were around the upcoming break.

benaadams:
Sorry, was wrong, don't think this PR will break the CI.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+                logger.Log(
+                    logLevel: LogLevel.Information,
+                    eventId: 2,
+                    state: new HostingRequestFinished(httpContext),
+                    exception: null,
+                    formatter: HostingRequestFinished.Callback);
+            }
+        }
+
+        private class HostingRequestScope : ILogValues
```
davidfowl: HostingLogScope. RequestScope sounds like DI

lodejard: Okay, but the other two are still HostingRequestXxx

benaadams:
Rebased #409 onto last commit and picked up remaining clean up and ci break

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -0,0 +1,141 @@
...
+                    eventId: 2,
+                    state: new HostingRequestFinished(httpContext),
+                    exception: null,
+                    formatter: HostingRequestFinished.Callback);
+            }
+        }
+
+        private class HostingLogScope : ILogValues
+        {
+            private readonly HttpContext _httpContext;
```
benaadams: Could move this up to ```HttpAbstractions``` implement in safe general way, and thus drop the ```HostingLogScope``` class create in ```RequestScope(...)```

```csharp
public abstract class HttpContext : ILogValues
{
    ...
   
    public override string ToString() => $"RequestId:{TraceIdentifier} RequestPath:{Request.Path}";

    IEnumerable<KeyValuePair<string, object>> ILogValues.GetValues()
    {
        yield return new KeyValuePair<string, object>("RequestId", TraceIdentifier);
        yield return new KeyValuePair<string, object>("RequestPath", Request.Path.ToString());
    }
}
```

Specific derived classes could then override with caching as per ```HostingLogScope```

Tratcher: I wouldn't recommend it. Hosting has an opinion on what is appropriate to log at this moment in the request's execution. That opinion is not valid or useful elsewhere in the app.

benaadams: Hmm, then switch 
```csharp
public static IDisposable RequestScope(this ILogger logger, HttpContext httpContext)
{
    return logger.BeginScopeImpl(new HostingLogScope(httpContext));
}
```
for
```csharp
public static IDisposable RequestScope(this ILogger logger, HttpContext httpContext)
{
    return logger.IsEnabled(LogLevel.Critical) 
                ? logger.BeginScopeImpl(new HostingLogScope(httpContext)) 
                : null;
}
```
Would that work?

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -104,24 +103,27 @@ public virtual IApplication Start()
...
+                            {
+                                telemetrySource.WriteTelemetry("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
+                            }
+#pragma warning restore 0618
+                            throw;
+                        }
+                        finally
                         {
-                            telemetrySource.WriteTelemetry("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
+                            logger.RequestFinished(httpContext);
```
Tratcher: It's odd to log the response when there's an unhandled exception. E.g. the status code will probably be 200 until after the server catches and handles the exception. Maybe FinishRequest(exception) for the failure case?

The response also may not be complete at this point. E.g. if there is no body, or if the body was buffered, OnStarting may not have fired yet. You may be logging incomplete headers.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
rynowak: Per discussion with lou, that's not really correct. Scopes are relevant to all loggers, not just to the current one. We'd need a sort of global 'all loggers disabled' check.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
rynowak: Per discussion with lou, going to take this out and follow up later on adding the length here.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
benaadams: Doesn't it go via `Microsoft.Extensions.Logging.Logger` which enumerates all loggers in the `IsEnabled` method bailing on first true?

Also have fast path PR for no loggers for the check: https://github.com/aspnet/Logging/pull/254/files

An `IsDisabled` with no params would be prettier, but would change the interface.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
rynowak: Agreed, this is better done at the server level, but we want to get something basic done here to flesh out the experience for RC. Opened https://github.com/aspnet/Hosting/issues/443

rynowak:
@Tratcher  updated

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -94,28 +94,33 @@ public virtual IApplication Start()
...
+                            if (diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
+                            {
+                                diagnosticSource.Write("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
+                            }
+                            throw;
                         }
-                        throw;
+                        finally
+                        {
+                            logger.RequestFinished(httpContext);
```
Tratcher: It's still weird that this fires for unhandled exceptions and logs 200 OK.

rynowak: How would you feel about just logging the elapsed time on failure then - no contenttype/status code. It really seems like the best we can do for now

Tratcher: Better, but you don't have the elapsed time here, do you? I would expect `logger.RequestFailed()`, with or without the exception.

lodejard: An alternative is to hook the request completed event to have the correct status code and possibly more accurate overall time. That would need to be weighed against the overhead of the event hook per request - but it can be done only if that level is enabled - so it's no cost at all when no loggers aren't added.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
benaadams: Ah, you mean because its an extension method so can be used in multiple places rather than just on the top level logger?

rynowak:
@Tratcher updated

benaadams:
Remove `src/Microsoft.AspNet.Hosting/Internal/FastHttpRequestIdentifierFeature.cs`

Remove function `private string GetRequestIdentifier(HttpContext httpContext)` from `src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs`

Remove `"allowUnsafe": true` from `src/Microsoft.AspNet.Hosting/project.json`

Remove `test/Microsoft.AspNet.Hosting.Tests/Internal/FastHttpRequestIdentifierFeatureTests.cs`

Change function `HostingEngine_CreatesDefaultRequestIdentifierFeature_IfNotPresent` in `test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs` from

````csharp
Assert.NotNull(httpContext);
Assert.IsType<FastHttpRequestIdentifierFeature>(httpContext.Features.Get<IHttpRequestIdentifierFeature>());
````
to
````csharp
Assert.NotNull(httpContext);
Assert.IsType<HttpRequestIdentifierFeature>(httpContext.Features.Get<IHttpRequestIdentifierFeature>());
Assert.False(string.IsNullOrWhiteSpace(httpContext.TraceIdentifier));
Assert.Same(httpContext.TraceIdentifier, httpContext.Features.Get<IHttpRequestIdentifierFeature>().TraceIdentifier);
````
As is overhead and otherwise now redundant (has been moved into `HttpAbstractions`)

File: src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs [hunk 14]
```diff
@@ -93,27 +93,40 @@ public virtual IApplication Start()
                 {
                     var httpContext = contextFactory.CreateHttpContext(features);
                     httpContext.ApplicationServices = _applicationServices;
+
                     var requestIdentifier = GetRequestIdentifier(httpContext);
```
Tratcher: Remove this line.

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -93,27 +93,40 @@ public virtual IApplication Start()
...
+                            logger.RequestFailed(httpContext, startTime);
+
+                            if (diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
+                            {
+                                diagnosticSource.Write("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
+                            }
+                            throw;
                         }
-                        throw;
+                        finally
```
Tratcher: remove empty finally block

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -93,27 +93,35 @@ public virtual IApplication Start()
                 {
                     var httpContext = contextFactory.CreateHttpContext(features);
                     httpContext.ApplicationServices = _applicationServices;
-                    var requestIdentifier = GetRequestIdentifier(httpContext);
+
```
Tratcher: nit: empty line

Tratcher:
:shipit:

rynowak:
Will wait until CI for hosting is green to push. Currently broken due to issues not related to this change.

rynowak:
f37375f43cd9e49ebcd61b62130d3455c8d41944

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
lodejard: It turns out it was a bug to check the IsEnabled to short-circuit the scope. Even if messages in the `logger`'s category are entirely disabled, there can be other messages in the application that will be logged and will need the logger scope available... We might be able to optimize the BeginScopeImpl a little bit more and get the per-request overhead down to a single object with one field, then measure it from there to see how much impact that is having

PR #413, opened 2015-10-16
Add different logic to get runtime for coreclr on mac

kirthik:
cc @kichalla, @Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
@@ -99,41 +99,60 @@ namespace Microsoft.AspNet.Server.Testing
 
         protected string PopulateChosenRuntimeInformation()
         {
-            // ex: runtimes/dnx-coreclr-win-x64.1.0.0-rc1-15844/bin
-            var currentRuntimeBinPath = Path.GetDirectoryName(Process.GetCurrentProcess().MainModule.FileName);
-            Logger.LogInformation($"Current runtime path is : {currentRuntimeBinPath}");
-
-            var targetRuntimeName = new StringBuilder()
-                .Append("dnx")
-                .Append((DeploymentParameters.RuntimeFlavor == RuntimeFlavor.CoreClr) ? "-coreclr" : "-clr")
-                .Append($"-{OSPrefix}")
-                .Append((DeploymentParameters.RuntimeArchitecture == RuntimeArchitecture.x86) ? "-x86" : "-x64")
-                .ToString();
-
-            string targetRuntimeBinPath;
-            // Ex: When current runtime is Mono and the tests are being run for CoreClr
-            if (currentRuntimeBinPath.Contains("dnx-mono"))
+            if (TestPlatformHelper.IsMac && DeploymentParameters.RuntimeFlavor == RuntimeFlavor.CoreClr)
             {
-                targetRuntimeBinPath = currentRuntimeBinPath.Replace("dnx-mono", targetRuntimeName);
+                var path = Environment.GetEnvironmentVariable("PATH");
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -99,41 +99,60 @@ protected string DnxCommandName
...
-                .Append("dnx")
-                .Append((DeploymentParameters.RuntimeFlavor == RuntimeFlavor.CoreClr) ? "-coreclr" : "-clr")
-                .Append($"-{OSPrefix}")
-                .Append((DeploymentParameters.RuntimeArchitecture == RuntimeArchitecture.x86) ? "-x86" : "-x64")
-                .ToString();
-
-            string targetRuntimeBinPath;
-            // Ex: When current runtime is Mono and the tests are being run for CoreClr
-            if (currentRuntimeBinPath.Contains("dnx-mono"))
+            if (TestPlatformHelper.IsMac && DeploymentParameters.RuntimeFlavor == RuntimeFlavor.CoreClr)
```
kichalla: Instead of this, we could following the approach what the MonoDeployer is doing (and probably extract that common logic in the application deployer so that even MonoDeployer can use it)

https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Server.Testing/Deployers/MonoDeployer.cs#L31-L38

kirthik:
@kichalla : I updated the code to use the replacement logic in `ApplicationDeployer`. I looked at merging `MonoDeployer` code with `ApplicationDeployer` but their logic is different. So kept it separate for now. Please let me know if you disagree.

src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -99,16 +99,31 @@ protected string DnxCommandName
...
+
+                if (string.IsNullOrWhiteSpace(currentRuntimeBinPath))
+                {
+                    throw new Exception("Runtime not detected on the machine.");
+                }
+            }
+            else
+            {
+                // ex: runtimes/dnx-coreclr-win-x64.1.0.0-rc1-15844/bin
+                currentRuntimeBinPath = Path.GetDirectoryName(Process.GetCurrentProcess().MainModule.FileName);
```
kichalla: You need not have this condition any more as we can have one way of getting current runtime bin path and it could be used both on mono and regular deployers...
```
protected string getcurrentruntimepath()
{
            var path = Environment.GetEnvironmentVariable("PATH");
            var runtimeBin = path.Split(new char[] { ':' }, StringSplitOptions.RemoveEmptyEntries).
                Where(c => c.Contains("dnx-mono") || c.Contains("dnx-clr") || c.Contains("dnx-coreclr")).FirstOrDefault();

            if (string.IsNullOrWhiteSpace(runtimeBin))
            {
                throw new Exception("Runtime not detected on the machine.");
            }

     return path;
}
```

src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -133,7 +148,7 @@ protected string PopulateChosenRuntimeInformation()
 
             ChosenRuntimePath = targetRuntimeBinDir.FullName;
             ChosenRuntimeName = targetRuntimeBinDir.Parent.Name;
-            DeploymentParameters.DnxRuntime = ChosenRuntimeName;
+            DeploymentParameters.DnxRuntime = PopulateChosenRuntimeInformation();
```
kichalla: this change is not required?

kirthik: Oops. That got added by mistake.

kichalla:
:shipit:

kirthik:
merged 5f4f6194b8014ea0f63cd812db3b17143ef07890

PR #421, opened 2015-10-19
Added doc-comments to IHostingEnvironment

DamianEdwards:
Coherence-Signed#75

@Eilon

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
+++ b/src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
@@ -6,6 +6,7 @@ namespace Microsoft.AspNet.Hosting
     public static class EnvironmentName
     {
         public static readonly string Development = "Development";
+        public static readonly string Staging = "Staging";
         public static readonly string Production = "Production";
     }
 }
--- a/src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
+++ b/src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
@@ -3,6 +3,9 @@
 
 namespace Microsoft.AspNet.Hosting
 {
+    /// <summary>
+    /// Commonly used environment names.
+    /// </summary>
     public static class EnvironmentName
     {
         public static readonly string Development = "Development";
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Hosting.Abstractions/IHostingEnvironment.cs
```diff
@@ -5,14 +5,27 @@
...
 namespace Microsoft.AspNet.Hosting
 {
+    /// <summary>
+    /// Provides information about the web hosting environment an application is running in.
+    /// </summary>
     public interface IHostingEnvironment
     {
+        /// <summary>
+        /// Gets or sets the name of the environment. This property is automatically set by the host to the value
+        /// of the <c>ASPNET_ENV</c> environment variable.
```
davidfowl: Hosting:Environment or ASPNET_ENV

Tratcher: Don't encourage the use of old environment variable names. We should just doc the current/new names.

davidfowl: On linux it's Hosting__Environment since we can't use : there. So we might need to start using the platform neutral ones.

DamianEdwards: So, which is it?

Tratcher: Hosting:Environment

DamianEdwards: But that's not x-plat, so what should we put here?

src/Microsoft.AspNet.Hosting.Abstractions/HostingEnvironmentExtensions.cs
```diff
@@ -9,10 +9,10 @@ namespace Microsoft.AspNet.Hosting
     public static class HostingEnvironmentExtensions
```
Eilon: Can add the generic extension class blurb here.

```c#
    /// <summary>
    /// Extension methods for doing BLAH to a <see cref="BLAH"/>.
    /// </summary>
```

src/Microsoft.AspNet.Hosting.Abstractions/EnvironmentName.cs
```diff
@@ -6,6 +6,7 @@ namespace Microsoft.AspNet.Hosting
     public static class EnvironmentName
```
Eilon: WUT? No doc comments here?

DamianEdwards:
@Eilon I added doc comments to extension class.

Eilon:
:shipit:

PR #425, opened 2015-10-20
React to dnx dependency injection changes

pakrym:
@davidfowl

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Program.cs
+++ b/src/Microsoft.AspNet.Hosting/Program.cs
@@ -4,9 +4,11 @@
 using System;
 using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Server.Features;
+using Microsoft.Dnx.Compilation;
 using Microsoft.Dnx.Runtime;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 
 namespace Microsoft.AspNet.Hosting
 {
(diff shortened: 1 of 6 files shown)


Tratcher:
Next you'll need to update Kestrel and WebListener: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNet.Server.Kestrel/Program.cs

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -15,21 +17,24 @@ public class Program
...
 
-            var appBasePath = _serviceProvider.GetRequiredService<IApplicationEnvironment>().ApplicationBasePath;
+            var services = new ServiceCollection();
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.Application));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.Runtime));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.AssemblyLoadContextAccessor));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.AssemblyLoaderContainer));
+            services.TryAdd(ServiceDescriptor.Instance(PlatformServices.Default.LibraryManager));
+            services.TryAdd(ServiceDescriptor.Instance(CompilationServices.Default.LibraryExporter));
+            var serviceProvider = services.BuildServiceProvider();
```
Tratcher: @davidfowl Since this is all static now can we just move it into a WebHostBuilder constructor? That should also fix TestHost.

pakrym: I tried, some tests rely on the fact that some of services are not initialized (ThrowsIfNoApplicationEnvironmentIsRegisteredWithTheProvider) and with ApplicationHost always providing static ApplicationEnvironment it fails

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
```diff
@@ -73,18 +70,35 @@ public WebHostBuilder(IServiceProvider services, IConfiguration config, bool cap
...
             var services = new ServiceCollection();
 
-            // Import from manifest
-            var manifest = _services.GetService<IRuntimeServices>();
-            if (manifest != null)
+            if (_services != null)
             {
-                foreach (var service in manifest.Services)
+                // Import from manifest
+                var manifest = _services.GetService<IRuntimeServices>();
```
davidfowl: `_services?.`

davidfowl:
Remove the ISP from WebHostBuilder.

pakrym:
@davidfowl done

davidfowl:
Looks good! :shipit:

rynowak:
Is there a reason why we're still waiting on this? MVC is broken

muratg:
@rynowak @pakrym should be able to check this in early in the morning.

davidfowl:
@rynowak no this was https://github.com/aspnet/dnx/commit/6d132950fc2fbbead501cbd10bcb6dbfdb10ac0c

JunTaoLuo:
@pakrym The appveyor failure is due to the renaming of telemetry and should be fixed now. See if you can restart the test. Even if you can't, I think it's safe to checkin since travis is passing now. This is blocking quite a few things.

Issue #387, opened 2015-10-05
Private Configure message should be more explicit


victorhurdugaci:
Change the Configure method to private and run:

```
private void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
```

- Expected: an error message saying that `Configure` should be public
- Actual error message: "A method named 'ConfigureDevelopment' or 'Configure' in the type 'WebApplication1.Startup' could not be found."

JunTaoLuo:
merged to dev

PR #427, opened 2015-10-20
Johluo/387 more descriptive message for private configure

JunTaoLuo:
#387 I'm assuming that if `ConfigureServices` is set to private, we'll just silently ignore it, similar to how we currently handle `ConfigureServices` with incompatible return type.

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
+++ b/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
@@ -126,7 +126,7 @@ namespace Microsoft.AspNet.Hosting.Startup
             {
                 if (required)
                 {
-                    throw new InvalidOperationException(string.Format("A method named '{0}' or '{1}' in the type '{2}' could not be found.",
+                    throw new InvalidOperationException(string.Format("A public method named '{0}' or '{1}' could not be found in the '{2}' type.",
                         methodNameWithEnv,
                         methodNameWithNoEnv,
                         startupType.FullName));
--- a/test/Microsoft.AspNet.Hosting.Tests/Fakes/StartupPrivateConfigure.cs
+++ b/test/Microsoft.AspNet.Hosting.Tests/Fakes/StartupPrivateConfigure.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Builder;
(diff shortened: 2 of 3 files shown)


src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -49,7 +49,7 @@ public Type FindStartupType(string startupAssemblyName, IList<string> diagnostic
             var assembly = Assembly.Load(new AssemblyName(startupAssemblyName));
             if (assembly == null)
             {
-                throw new InvalidOperationException(String.Format("The assembly '{0}' failed to load.", startupAssemblyName));
+                throw new InvalidOperationException($"The assembly '{startupAssemblyName}' failed to load.");
```
Tratcher: These should be moved to resource strings instead.

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -106,18 +103,18 @@ private static MethodInfo FindMethod(Type startupType, string methodName, string
             var methodNameWithEnv = string.Format(CultureInfo.InvariantCulture, methodName, environmentName);
             var methodNameWithNoEnv = string.Format(CultureInfo.InvariantCulture, methodName, "");
 
-            var methods = startupType.GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static);
+            var methods = startupType.GetMethods(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static);
```
Tratcher: This will cause new conflicts on line 108 if you have a public and a private overload of the method. How about just changing the original exceptions to "A public method named..."

JunTaoLuo:
ping @victorhurdugaci

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -126,7 +126,7 @@ private static MethodInfo FindMethod(Type startupType, string methodName, string
             {
                 if (required)
                 {
-                    throw new InvalidOperationException(string.Format("A method named '{0}' or '{1}' in the type '{2}' could not be found.",
+                    throw new InvalidOperationException(string.Format("A public method named '{0}' or '{1}' in the type '{2}' could not be found.",
```
victorhurdugaci: This message sounds bad. It should be "A public method named ... could not be found in the ... type"

victorhurdugaci:
:shipit: after you investigate/fix travis and appveyor

Issue #411, opened 2015-10-16
Move webroot out of project.json


davidfowl:
We should have a default convention based `wwwroot` in a known location relative to the app base. As an override, it can also be specified as a config option.

muratg:
@Tratcher Let's get this done before other RC1 items, and remove project.json parsing from Hosting layer.

@BillHiebert FYI. This will have tooling impact that will need to go in RC1.  Do you have a tracking bug on your side.

BillHiebert:
Do now,  TFS# Bug 156338 

@sayedihashimi

muratg:
Note: project.json schema will need to change.

muratg:
@Tratcher Could you file the DNX bug for the publish, test etc change (RC1) as well?

Tratcher:
@muratg https://github.com/aspnet/dnx/issues/3035

Tratcher:
@BillHiebert does tooling read the webroot from project.json for anything?

davidfowl:
@Tratcher please post an announcement

Tratcher:
@davidfowl waiting on the details of aspnet/dnx#3035

PR #431, opened 2015-10-20
Move webroot from project.json to config.

Tratcher:
#411 @davidfowl @muratg


test/Microsoft.AspNet.TestHost.Tests/TestServerTests.cs
```diff
@@ -357,12 +357,12 @@ public ContextHolder(IHttpContextAccessor accessor)
...
 
         [Fact]
-        public void WebRootCanBeResolvedWhenNotInTheProjectJson()
+        public void WebRootCanBeResolvedWhenNotInTheConfig()
         {
             TestServer server = TestServer.Create(app =>
             {
                 var env = app.ApplicationServices.GetRequiredService<IHostingEnvironment>();
-                Assert.Equal(Directory.GetCurrentDirectory(), env.WebRootPath);
+                Assert.Equal(Directory.GetCurrentDirectory() + "\\wwwroot", env.WebRootPath);
```
davidfowl: Path.Combine for xplat

Tratcher:
Minor test update.

davidfowl:
Why does travis hate this?

davidfowl:
One behavior change is that we will always create a wwwroot. I think we shouldn't do that. We need to only do it if there is a wwwroot. So the create if not exists logic only runs if it is defined, we can still have a default wwwroot anyways.

Tratcher:
You mean default WebRootPath to applicationBasePath instead of applicationBasePath\wwwroot if webroot isn't specified in the config?

davidfowl:
@Tratcher That's what we did before right?

Tratcher:
Yes. Updated.

davidfowl:
:shipit: tho I dunno why the ci is failing

PR #435, opened 2015-10-21
Use UseServices to init services instead of UseStartup

pakrym:
@Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNet.TestHost/TestServer.cs
+++ b/src/Microsoft.AspNet.TestHost/TestServer.cs
@@ -48,7 +48,11 @@ namespace Microsoft.AspNet.TestHost
 
         public static TestServer Create(Action<IApplicationBuilder> configureApp, Func<IServiceCollection, IServiceProvider> configureServices)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -71,7 +71,7 @@ public static WebHostBuilder CreateBuilder(IConfiguration config, Action<IApplic
 
         public static WebHostBuilder CreateBuilder(IConfiguration config, Action<IApplicationBuilder> configureApp, Func<IServiceCollection, IServiceProvider> configureServices)
         {
-            return CreateBuilder(config).UseStartup(configureApp, configureServices);
+            return CreateBuilder(config).UseServices((services)=>configureServices(services)).UseStartup(configureApp);
```
davidfowl: This isn't right

Tratcher:
I don't understand why. Bug#? Reason?

davidfowl:
:shipit: It's super ugly though

Issue #405, opened 2015-10-14
Assign IHttpContextAccessor in the default implementation of IHttpContextFactory


davidfowl:
As part of https://github.com/aspnet/Hosting/issues/404, we should be setting the IHttpContextAccessor in CreateHttpContext

benaadams:
Is ```CreateHttpContext``` per request or per connection? Per connection is fine; but per request setting an ```AsyncLocal``` incurs a ```new Dictionary``` cost per set.

davidfowl:
It's per request. Nothing is changing there. We're just moving things around so that application code can remove it if required.

benaadams:
Ah, kk

JunTaoLuo:
merged to dev

PR #436, opened 2015-10-21
Setting IHttpContextAccessor in CreateHttpContext #405

JunTaoLuo:
#405

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
+++ b/src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
@@ -9,9 +9,18 @@ namespace Microsoft.AspNet.Hosting.Builder
 {
     public class HttpContextFactory : IHttpContextFactory
     {
+        private IHttpContextAccessor _httpContextAccessor;
+
+        public HttpContextFactory(IHttpContextAccessor httpContextAccessor)
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
         public HttpContext CreateHttpContext(IFeatureCollection featureCollection)
         {
-            return new DefaultHttpContext(featureCollection);
+            var httpContext = new DefaultHttpContext(featureCollection);
+            _httpContextAccessor.HttpContext = httpContext;
+            return httpContext;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
```diff
@@ -9,9 +9,18 @@ namespace Microsoft.AspNet.Hosting.Builder
...
+        public HttpContextFactory(IHttpContextAccessor httpContextAccessor)
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
         public HttpContext CreateHttpContext(IFeatureCollection featureCollection)
         {
-            return new DefaultHttpContext(featureCollection);
+            var httpContext = new DefaultHttpContext(featureCollection);
+            _httpContextAccessor.HttpContext = httpContext;
```
benaadams: Should this be
```csharp
if (_httpContextAccessor != null) 
{
    _httpContextAccessor.HttpContext = httpContext;
}
```

JunTaoLuo: Yes that would be a good idea.

src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
```diff
@@ -9,9 +9,23 @@ namespace Microsoft.AspNet.Hosting.Builder
...
+        {
+            _httpContextAccessor = httpContextAccessor;
+        }
+
         public HttpContext CreateHttpContext(IFeatureCollection featureCollection)
         {
-            return new DefaultHttpContext(featureCollection);
+            var httpContext = new DefaultHttpContext(featureCollection);
+
+            if (_httpContextAccessor != null)
```
davidfowl: Actually we don't support null constructor injection so we may need to adjust that

Tratcher:
:shipit:

davidfowl:
:shipit:

JunTaoLuo:
I merged this to dev even though tests failed since this is a small change and travis won't get the updated packages fixes until late today or early tomorrow when coherence passes again. The tests previously passed on AppVeyor.

Issue #437, opened 2015-10-22
Only add services from PlatformServices if the Default is set


davidfowl:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs#L98

davidfowl:
I guess the easiest way to do this would be to make it a lambda

davidfowl:
Done

PR #438, opened 2015-10-22
Don't add platforms services if no available

davidfowl:
#437

Depends on https://github.com/aspnet/dnx/pull/3042


pakrym:
:shipit:

PR #440, opened 2015-10-22
Default webroot to wwwroot if the directory exists.

Tratcher:
#411 @davidfowl @BrennanConroy


davidfowl:
Use Path.GetFullPath to resolve the full path. Otherwise looks good :shipit:

PR #441, opened 2015-10-22
Remove compilation options from services.

NTaylorMullen:
- This was a temporary work around to allow MVC to build.
- Reacts to aspnet/Mvc#3383

/cc @pranavkm


pranavkm:
:shipit:

Issue #430, opened 2015-10-20
Create WebApplication class for starting web apps application entry points (static void Main)


DamianEdwards:
Create a new type that will be called from web applications in their `static void Main(string args)`. Note this depends on #425

``` C#
public static class WebApplication
{
    public static void Start<TStartup>()
    {
        Start(typeof(TStartup), null);
    }

    public static void Start<TStartup>(string[] args)
    {
        Start(typeof(TStartup), args);
    }

    public static void Start(Type startupType)
    {
        Start(startupType, null);
    }

    public static void Start(Type startupType, string[] args)
    {
        // Basically the body of Microsoft.AspNet.Hosting.Program.Main, which will now call into this instead
        ...
    }
}
```

PR #444, opened 2015-10-22
Move Program.Main into WebApplication.Start.

Tratcher:
#430 @davidfowl @DamianEdwards @muratg 
This is a really strange class and it's going to confuse anyone that used Katana. It's especially strange that Start never returns. Can we rename Start to Main or Run?

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Program.cs
+++ b/src/Microsoft.AspNet.Hosting/Program.cs
@@ -1,65 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using Microsoft.AspNet.Http.Features;
-using Microsoft.AspNet.Server.Features;
-using Microsoft.Dnx.Compilation;
-using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.Extensions.PlatformAbstractions;
-
 namespace Microsoft.AspNet.Hosting
 {
     public class Program
     {
-        private const string HostingJsonFile = "Microsoft.AspNet.Hosting.json";
-        private const string ConfigFileKey = "config";
-
         public static void Main(string[] args)
         {
-            // Allow the location of the json file to be specified via a --config command line arg
-            var tempBuilder = new ConfigurationBuilder().AddCommandLine(args);
-            var tempConfig = tempBuilder.Build();
-            var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
-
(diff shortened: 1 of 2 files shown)


muratg:
Perhaps we should keep the Program class, and add the Start methods there?

Tratcher:
`public static void Main(string[] args) => Program.Start<Startup>(args);`? Hmm, I think that looses some context because you don't see who's Program your calling. Nor does it address the name Start.

davidfowl:
Call it run instead of start

davidfowl:
Also optionally take in the startup class

Tratcher:
An instance of the startup class rather than just the type? We don't actually support that anywhere in Hosting right now and it won't be trivial to add. We can work on that later.

davidfowl:
@Tratcher no, WebApplication.Run<TStartup>()

davidfowl:
Oh wait you have that :small_blue_diamond:

davidfowl:
Rebase and :shipit:

davidfowl:
Hmmm hold on

src/Microsoft.AspNet.Hosting/WebApplication.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        public static void Start(Type startupType, string[] args)
+        {
+            // Allow the location of the json file to be specified via a --config command line arg
+            var tempBuilder = new ConfigurationBuilder().AddCommandLine(args);
+            var tempConfig = tempBuilder.Build();
+            var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
+
+            var appBasePath = PlatformServices.Default.Application.ApplicationBasePath;
```
davidfowl: We need to tweak this a bit. Use this logic:

https://github.com/aspnet/Configuration/blob/dev/src/Microsoft.Extensions.Configuration.FileExtensions/FileConfigurationExtensions.cs#L52-L57

To resolve the application base path

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -1,65 +1,13 @@
...
-                Console.CancelKeyPress += (sender, eventArgs) =>
-                {
-                    appLifetime.StopApplication();
-                    // Don't terminate the process immediately, wait for the Main thread to exit gracefully.
-                    eventArgs.Cancel = true;
-                };
-
-                appLifetime.ApplicationStopping.WaitHandle.WaitOne();
-            }
+            WebApplication.Start(args);
```
davidfowl: Rename Start to Run

src/Microsoft.AspNet.Hosting/WebApplication.cs
Tratcher: Why isn't that done inside of PlatformServices.Default.Application.ApplicationBasePath?

davidfowl:
:shipit:

Issue #395, opened 2015-10-10
Revisit IServerFactory.Start


davidfowl:
Does it really need to pass back in the IFeatureCollection returned in Initialize?

```
IDisposable Start(IFeatureCollection serverFeatures, Func<IFeatureCollection, Task> application);
```

Why not just:

```
IDisposable Start(Func<IFeatureCollection, Task> application);
```

The current design does allow you to pass in different instances but I'm not sure if that's a goal since we don't use it that way currently.

Tratcher:
Well as a factory it can be used more than once (not that we actually do this at the moment). Maybe a factory isn't the pattern you want here? Removing that param makes it a statefull single use class. At that point you would just make the factory itself disposable rather than returning anything from start.

davidfowl:
Thats what I'm wondering. I know the pattern allows for that but I'm not sure we use it or plan to use it anywhere that way. I know this stuff came from katana but I'd like to understand if we ever plan to use it in that way in the current stack.

/cc @loudej @lodejard

davidfowl:
```C#
public interface IServerFactory : IDisposable
{
    IFeatureCollection Initialize(IConfiguration configuration);
    void Start(Func<IFeatureCollection, Task> application);
}
```

I'm not sure. I guess it does feel a bit off that some of those things are inferred (e.g. the fact that the returned IFeatureCollection should be used later by Start?)

Tratcher:
How about:
```c#
public interface IServer : IDisposable
{
    IFeatureCollection Features { get; }
    void Initialize(IConfiguration configuration);
    void Start(Func<IFeatureCollection, Task> application);
}
```

Note initialize only exists in this model because we don't have a good way of getting the IConfiguration into the DI constructor.

davidfowl:
What about:

```C#
public interface IServerFactory
{
    IServer CreateServer(IConfiguration configuration);
}
```

```c#
public interface IServer : IDisposable
{
    IFeatureCollection Features { get; }
    void Start(Func<IFeatureCollection, Task> application);
}
```

Tratcher:
*shrug* we can try it. I don't think it solves any issues, it's just API cleanup.

davidfowl:
This whole issue is about API cleanup. I was demoing the interface recently and decided to file this issue because I thought the stateless nature of the server factory was a little weird.

Eilon:
After discussion with @lodejard @davidfowl , slight change:

```C#
public interface IServerFactory
{
[...]

PR #445, opened 2015-10-23
Refactoring IServerFactory #395

JunTaoLuo:
#395

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Server.Abstractions/IServer.cs
+++ b/src/Microsoft.AspNet.Hosting.Server.Abstractions/IServer.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Hosting.Server
+{
+    /// <summary>
+    /// Represents a server.
+    /// </summary>
+    public interface IServer : IDisposable
+    {
+        /// <summary>
+        /// A collection of HTTP features of the server.
+        /// </summary>
+        IFeatureCollection Features { get; }
+
+        /// <summary>
+        /// Start the server with the given function that processes an HTTP request.
+        /// </summary>
+        /// <param name="requestDelegate">A function that processes an HTTP request.</param>
+        void Start(RequestDelegate requestDelegate);
+    }
+}
--- a/src/Microsoft.AspNet.Hosting.Server.Abstractions/IServerFactory.cs
+++ b/src/Microsoft.AspNet.Hosting.Server.Abstractions/IServerFactory.cs
@@ -1,16 +1,20 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Threading.Tasks;
-using Microsoft.AspNet.Http.Features;
 using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.AspNet.Hosting.Server
 {
+    /// <summary>
+    /// Represents a factory for creating servers.
+    /// </summary>
     public interface IServerFactory
     {
-        IFeatureCollection Initialize(IConfiguration configuration);
-        IDisposable Start(IFeatureCollection serverFeatures, Func<IFeatureCollection, Task> application);
+        /// <summary>
+        /// Creates <see cref="IServer"/> based on the given configuration.
+        /// </summary>
+        /// <param name="configuration">An instance of <see cref="IConfiguration"/>.</param>
+        /// <returns>The created server.</returns>
+        IServer CreateServer(IConfiguration configuration);
     }
 }
--- a/src/Microsoft.AspNet.Hosting.Server.Abstractions/project.json
+++ b/src/Microsoft.AspNet.Hosting.Server.Abstractions/project.json
@@ -6,7 +6,7 @@
     "url": "git://github.com/aspnet/hosting"
   },
   "dependencies": {
-    "Microsoft.AspNet.Http.Features": "1.0.0-*",
+    "Microsoft.AspNet.Http.Abstractions": "1.0.0-*",
     "Microsoft.Extensions.Configuration.Abstractions": "1.0.0-*"
   },
   "frameworks": {
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -42,7 +42,7 @@ namespace Microsoft.AspNet.Hosting.Internal
         // Only one of these should be set
         internal IServerFactory ServerFactory { get; set; }
         internal string ServerFactoryLocation { get; set; }
-        private IFeatureCollection _serverFeatures;
+        internal IServer Server { get; set; }
 
         public HostingEngine(
             IServiceCollection appServices,
@@ -87,15 +87,13 @@ namespace Microsoft.AspNet.Hosting.Internal
             var application = BuildApplication();
 
             var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
-            var contextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
             var diagnosticSource = _applicationServices.GetRequiredService<DiagnosticSource>();
 
             logger.Starting();
 
-            var server = ServerFactory.Start(_serverFeatures,
-                async features =>
+            Server.Start(
+                async httpContext =>
                 {
-                    var httpContext = contextFactory.Create(features);
                     httpContext.ApplicationServices = _applicationServices;
 
                     if (diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
@@ -135,11 +133,11 @@ namespace Microsoft.AspNet.Hosting.Internal
             _applicationLifetime.NotifyStarted();
             logger.Started();
 
-            return new Application(ApplicationServices, _serverFeatures, new Disposable(() =>
+            return new Application(ApplicationServices, Server.Features, new Disposable(() =>
             {
                 logger.Shutdown();
                 _applicationLifetime.StopApplication();
-                server.Dispose();
+                Server.Dispose();
                 _applicationLifetime.NotifyStopped();
                 (_applicationServices as IDisposable)?.Dispose();
             }));
@@ -191,7 +189,7 @@ namespace Microsoft.AspNet.Hosting.Internal
                 EnsureServer();
 
                 var builderFactory = _applicationServices.GetRequiredService<IApplicationBuilderFactory>();
-                var builder = builderFactory.CreateBuilder(_serverFeatures);
+                var builder = builderFactory.CreateBuilder(Server.Features);
                 builder.ApplicationServices = _applicationServices;
 
                 var startupFilters = _applicationServices.GetService<IEnumerable<IStartupFilter>>();
@@ -246,21 +244,21 @@ namespace Microsoft.AspNet.Hosting.Internal
 
         private void EnsureServer()
         {
-            if (ServerFactory == null)
+            if (Server == null)
             {
-                // Blow up if we don't have a server set at this point
-                if (ServerFactoryLocation == null)
+                if (ServerFactory == null)
                 {
-                    throw new InvalidOperationException("IHostingBuilder.UseServer() is required for " + nameof(Start) + "()");
+                    // Blow up if we don't have a server set at this point
+                    if (ServerFactoryLocation == null)
+                    {
+                        throw new InvalidOperationException("IHostingBuilder.UseServer() is required for " + nameof(Start) + "()");
+                    }
+
+                    ServerFactory = _applicationServices.GetRequiredService<IServerLoader>().LoadServerFactory(ServerFactoryLocation);
                 }
 
-                ServerFactory = _applicationServices.GetRequiredService<IServerLoader>().LoadServerFactory(ServerFactoryLocation);
-            }
-
-            if (_serverFeatures == null)
-            {
-                _serverFeatures = ServerFactory.Initialize(_config);
-                var addresses = _serverFeatures?.Get<IServerAddressesFeature>()?.Addresses;
+                Server = ServerFactory.CreateServer(_config);
+                var addresses = Server.Features?.Get<IServerAddressesFeature>()?.Addresses;
                 if (addresses != null && !addresses.IsReadOnly)
                 {
                     var port = _config[ServerPort];
--- a/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
@@ -42,6 +42,7 @@ namespace Microsoft.AspNet.Hosting
         // Only one of these should be set
         private string _serverFactoryLocation;
         private IServerFactory _serverFactory;
+        private IServer _server;
 
         public WebHostBuilder()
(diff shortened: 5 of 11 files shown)


src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -245,10 +245,10 @@ private void EnsureServer()
...
             }
 
-            if (_serverFeatures == null)
+            if (_server == null)
             {
-                _serverFeatures = ServerFactory.Initialize(_config);
-                var addresses = _serverFeatures?.Get<IServerAddressesFeature>()?.Addresses;
+                _server = ServerFactory.CreateServer(_config);
+                var addresses = _server.Features?.Get<IServerAddressesFeature>()?.Addresses;
                 if (addresses != null && !addresses.IsReadOnly)
```
cesarbs: nit: you could do if (!addresses?.IsReadOnly)

src/Microsoft.AspNet.Hosting.Server.Abstractions/project.json
```diff
@@ -6,6 +6,7 @@
     "url": "git://github.com/aspnet/hosting"
   },
   "dependencies": {
+    "Microsoft.AspNet.Http.Abstractions": "1.0.0-*",
     "Microsoft.AspNet.Http.Features": "1.0.0-*",
```
Tratcher: You can remove Features, it's transitive.

Tratcher:
You're off to a good start. Next is moving IHttpContextFactory to Http.Abstractions and HttpContextFactory to Http. Then you need to update TestHost, Kestrel, and Entropy.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IServer.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Hosting.Server
+{
+    public interface IServer : IDisposable
```
davidfowl: Add some doc comments

src/Microsoft.AspNet.Hosting.Server.Abstractions/IServerFactory.cs
```diff
@@ -1,16 +1,12 @@
...
-using System;
-using System.Threading.Tasks;
-using Microsoft.AspNet.Http.Features;
 using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.AspNet.Hosting.Server
 {
     public interface IServerFactory
     {
-        IFeatureCollection Initialize(IConfiguration configuration);
```
davidfowl: Same here, add some doc comments

test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
```diff
@@ -26,9 +27,8 @@
 
 namespace Microsoft.AspNet.Hosting
 {
-    public class HostingEngineTests : IServerFactory
+    public class HostingEngineTests
```
davidfowl: To reduce the test churn, make this type implement both IServer and IServerFactory

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -14,18 +14,20 @@
 
 namespace Microsoft.AspNet.TestHost
 {
-    public class TestServer : IServerFactory, IDisposable
+    public class TestServer : IDisposable
```
davidfowl: Same comment here, make this implement both to reduce churn

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -14,18 +14,20 @@
...
     {
         private const string DefaultEnvironmentName = "Development";
         private const string ServerName = nameof(TestServer);
-        private static readonly IFeatureCollection ServerInfo = new FeatureCollection();
-        private Func<IFeatureCollection, Task> _appDelegate;
+        private RequestDelegate _appDelegate;
         private IDisposable _appInstance;
         private bool _disposed = false;
+        private IHttpContextAccessor _httpContextAccessor;
```
davidfowl: This looks wrong, you shouldn't have to do any of this. It's using the WebHostBuilder under the covers which does the right thing.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -121,6 +124,12 @@ internal RequestState(HttpRequestMessage request, PathString pathBase)
                 }
 
                 HttpContext = new DefaultHttpContext();
+
+                if (accessor != null)
```
davidfowl: This also looks wrong.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -23,14 +24,15 @@ namespace Microsoft.AspNet.TestHost
     /// </summary>
     public class ClientHandler : HttpMessageHandler
```
davidfowl: This type shouldn't have changed

src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
```diff
@@ -1,26 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using Microsoft.AspNet.Http;
-using Microsoft.AspNet.Http.Features;
-using Microsoft.AspNet.Http.Internal;
-
-namespace Microsoft.AspNet.Hosting.Builder
-{
-    public class HttpContextFactory : IHttpContextFactory
```
Tratcher: Make a separate PR to fix the build ASAP.

src/Microsoft.AspNet.TestHost/TestServer.cs
Tratcher: No, IHttpContextAccessor usage was moved into IHttpContextFactory which isn't useful to TestServer since it prefers to create the DefaultHttpContext directly.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -14,18 +14,20 @@
...
 
         public TestServer(WebHostBuilder builder)
         {
-            _appInstance = builder.UseServer(this).Build().Start();
+            var hostingEngine = builder.UseServer(new ServerFactory(requestDelegate => { _appDelegate = requestDelegate; })).Build();
```
Tratcher: Would it simplify matters if you added a UserServer overload that took an IServer directly? Then you could bypass the ServerFactory layer.

src/Microsoft.AspNet.Hosting/Builder/HttpContextFactory.cs
JunTaoLuo: Thanks for doing this yesterday. I was hoping to do everything in one go but that was overzealous.

src/Microsoft.AspNet.TestHost/TestServer.cs
davidfowl: We should just use the default factory

src/Microsoft.AspNet.TestHost/ClientHandler.cs
JunTaoLuo: I'm assuming this is because we need to decide whether to use the default factory or keep this hack for the custom factory?

src/Microsoft.AspNet.TestHost/ClientHandler.cs
JunTaoLuo: What's wrong here?

src/Microsoft.AspNet.TestHost/TestServer.cs
JunTaoLuo: I can. What's the order of precedence here? Look for the server, if not found try to create one with the resolved serverfactory?

davidfowl:
Based on what Chris said. We should update we host builder and test server to support taking and IServer or ISeverFactory. There's probably an effect on the restartability

src/Microsoft.AspNet.TestHost/ClientHandler.cs
davidfowl: We should be using the default http context factory that sets the accessor for us automatically.

test/Microsoft.AspNet.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -40,60 +41,65 @@ public void Build_honors_UseStartup_with_string()
         [Fact]
         public async Task StartupMissing_Fallback()
         {
+            var application = new RequestDelegate(context => { return Task.FromResult(0); });
```
davidfowl: Why the ceremony here? Shouldn't this just look like it did before?

JunTaoLuo: This will go away once `UseServer` takes an `IServer`. This hack was added so the `RequestDelegate` passed into the `IServer` could be accessed for use in `AssertResponseContains`.

JunTaoLuo: Alternatively, we can just make this test extend `IServer` and `IServerFactory` like `HostingEngineTests`.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
Tratcher: The default http context factory is less convienet to use here, but it would make it so you don't have to think about the accessor.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -14,22 +14,26 @@
 
 namespace Microsoft.AspNet.TestHost
 {
-    public class TestServer : IServerFactory, IDisposable
+    public class TestServer : IServerFactory, IServer, IDisposable
```
Tratcher: IDisposable is now redundant.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -14,22 +14,26 @@
...
         {
-            _appInstance = builder.UseServer(this).Build().Start();
+            var hostingEngine = builder.UseServer(this).Build();
+            _httpContextAccessor = hostingEngine.ApplicationServices.GetService<IHttpContextAccessor>();
+            _appInstance = hostingEngine.Start();
         }
 
         public Uri BaseAddress { get; set; } = new Uri("http://localhost/");
 
+        public IFeatureCollection Features { get; }
```
Tratcher: Some of these interface members don't actually make sense here, use the explicit implementation to hide them:
`IServer.Features { get; }`

Tratcher: Is this ever set?

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -145,5 +137,15 @@ public void Dispose()
             _disposed = true;
             _appInstance.Dispose();
         }
+
+        public IServer CreateServer(IConfiguration configuration)
```
Tratcher: Explicit

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -145,5 +137,15 @@ public void Dispose()
...
+        {
+            return this;
+        }
+
+        public void Start(RequestDelegate requestDelegate)
```
Tratcher: explicit

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -94,8 +94,7 @@ public RequestState(Uri uri, PathString pathBase, CancellationToken cancellation
                 _clientWebSocketTcs = new TaskCompletionSource<WebSocket>();
 
                 // HttpContext
-                FeatureCollection = new FeatureCollection();
-                HttpContext = new DefaultHttpContext(FeatureCollection);
+                HttpContext = new DefaultHttpContext(new FeatureCollection());
```
Tratcher: I guess this should also use the context factory.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
davidfowl: Yah, so use it

src/Microsoft.AspNet.TestHost/ClientHandler.cs
davidfowl: Use the default factory everywhere

test/Microsoft.AspNet.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -125,36 +125,31 @@ private WebHostBuilder CreateWebHostBuilder()
...
             httpContext.Response.Body.Seek(0, SeekOrigin.Begin);
             var bodyText = new StreamReader(httpContext.Response.Body).ReadToEnd();
             Assert.Contains(expectedText, bodyText);
         }
 
-        private class TestServerFactory : IServerFactory
+        private class TestServer : IServer
         {
-            public Func<IFeatureCollection, Task> Application { get; set; }
+            private Action<RequestDelegate> _setAppDelegate;
```
JunTaoLuo: removing this

JunTaoLuo:
updated

src/Microsoft.AspNet.Hosting.Server.Abstractions/IServerFactory.cs
```diff
@@ -1,16 +1,20 @@
...
-        IDisposable Start(IFeatureCollection serverFeatures, Func<IFeatureCollection, Task> application);
+        /// <summary>
+        /// Creates <see cref="IServer"/> based on the given configuration.
+        /// </summary>
+        /// <param name="configuration">An instance of <see cref="IConfiguration"/>.</param>
+        /// <returns>The created server.</returns>
+        IServer CreateServer(IConfiguration configuration);
```
Tratcher: Create or CreateServer?

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
```diff
@@ -172,7 +185,7 @@ public WebHostBuilder UseServer(string assemblyName)
             return this;
         }
 
-        public WebHostBuilder UseServer(IServerFactory factory)
+        public WebHostBuilder UseServerFactory(IServerFactory factory)
         {
             _serverFactory = factory;
```
Tratcher: Add a null check here for consistency?

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -120,7 +130,8 @@ internal RequestState(HttpRequestMessage request, PathString pathBase)
                     request.Headers.Host = request.RequestUri.GetComponents(UriComponents.HostAndPort, UriFormat.UriEscaped);
                 }
 
-                HttpContext = new DefaultHttpContext();
+                HttpContext = factory.Create(new FeatureCollection());
```
Tratcher: nit: contextFactory

Tratcher: _contextFactory

src/Microsoft.AspNet.Hosting.Server.Abstractions/IServerFactory.cs
JunTaoLuo: According to the issue filed, CreateServer.

Tratcher:
:shipit: (once we get the servers updated)

davidfowl:
This looks great :shipit:

Issue #434, opened 2015-10-21
RC1(15958): Unable to resolve service for type 'Microsoft.AspNet.Hosting.IApplicationLifetime'


xatabhk:
Running website under IISExpress in IDE gives no detailed error; but 'dnx web' prints the following error messages:
System.InvalidOperationException: Unable to resolve service for type 'Microsoft.AspNet.Hosting.IApplicationLifetime' while attempting to activate 'Microsoft.AspNet.Server.Kestrel.ServerFactory'.
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.ConstructorMatcher.CreateInstance(IServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateInstance(IServiceProvider provider, Type instanceType, Object[] parameters)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.GetServiceOrCreateInstance(IServiceProvider provider, Type type)
   at Microsoft.AspNet.Hosting.Server.ServerLoader.LoadServerFactory(String assemblyName)
   at Microsoft.AspNet.Hosting.Internal.HostingEngine.EnsureServer()
   at Microsoft.AspNet.Hosting.Internal.HostingEngine.BuildApplication()
   at Microsoft.AspNet.Hosting.Internal.HostingEngine.Start()
   at Microsoft.AspNet.Hosting.Program.Main(String[] args)
   at Microsoft.AspNet.Server.Kestrel.Program.Main(String[] args)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Dnx.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] args, IServiceProvider serviceProvider)
   at Microsoft.Dnx.ApplicationHost.Program.ExecuteMain(DefaultHost host, String applicationName, String[] args)
   at Microsoft.Dnx.ApplicationHost.Program.Main(String[] args)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Dnx.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] args, IServiceProvider serviceProvider)
[...]

PR #446, opened 2015-10-25
Make sure IApplicationLifetime gets added even if startup throws.

Tratcher:
#434 @davidfowl @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -70,6 +70,7 @@ namespace Microsoft.AspNet.Hosting.Internal
             _startupLoader = startupLoader;
             _captureStartupErrors = captureStartupErrors;
             _applicationLifetime = new ApplicationLifetime();
+            _applicationServiceCollection.AddInstance<IApplicationLifetime>(_applicationLifetime);
         }
 
         public IServiceProvider ApplicationServices
@@ -144,7 +145,6 @@ namespace Microsoft.AspNet.Hosting.Internal
             if (_applicationServices == null)
             {
                 EnsureStartup();
-                _applicationServiceCollection.AddInstance<IApplicationLifetime>(_applicationLifetime);
                 _applicationServices = Startup.ConfigureServicesDelegate(_applicationServiceCollection);
             }
         }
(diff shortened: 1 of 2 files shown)


davidfowl:
Add a test

Tratcher:
Updated

test/Microsoft.AspNet.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -74,6 +75,20 @@ public void Build_honors_UseStartup_with_string()
         }
 
         [Fact]
+        public async Task IApplicationLifetimeRegisteredEvenWhenStartupCtorThrows_Fallback()
+        {
+            var builder = CreateWebHostBuilder();
+            var serverFactory = new TestServerFactory();
+            var engine = (HostingEngine)builder.UseServer(serverFactory).UseStartup<StartupCtorThrows>().Build();
```
davidfowl: Why does this need to cast?

src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
```diff
@@ -70,6 +70,7 @@ public class HostingEngine : IHostingEngine
             _startupLoader = startupLoader;
             _captureStartupErrors = captureStartupErrors;
             _applicationLifetime = new ApplicationLifetime();
```
davidfowl: Is there any reason not to add this in the hosting services?

Tratcher: Yes, it needs the actual instance to call things like NotifyStarted.

test/Microsoft.AspNet.Hosting.Tests/WebHostBuilderTests.cs
Tratcher: Some of the tests assert on non-interface members, but not this one. removed.

Tratcher:
Updated.

davidfowl:
Why is travis failing?

Tratcher:
IHttpContextFactory due to the lack of a recent coherence build.

Tratcher:
@muratg

HaoK:
:shipit:

Issue #374, opened 2015-09-24
Add/Update logging in hosting


rynowak:
From offline logging discussion:

Add Verbose logging where it can be useful for debugging
- Time box to 2 dev-days maximum (across all of k-core)
- Use EventIds at each logging statement
- Update any existing logging to include an EventId

Relevant logging points we want to include for sure:

```
    Hosting:
        -Request - Information
        -Response - Information
        -Startup - Information
        -Shutdown - Information
```

muratg:
cc @Tratcher (also see the StaticFiles one)

PR #447, opened 2015-10-26
Event ids and more logging

victorhurdugaci:
Fixes https://github.com/aspnet/Hosting/issues/363 https://github.com/aspnet/Hosting/issues/374

- Event ids to logger events
- Startup/shutdown event ids

Please review @Tratcher @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
@@ -22,7 +22,7 @@ namespace Microsoft.AspNet.Hosting.Internal
             {
                 logger.Log(
                     logLevel: LogLevel.Information,
-                    eventId: 1,
+                    eventId: LoggerEventIds.RequestStarting,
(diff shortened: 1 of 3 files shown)


Tratcher:
LGTM, @rynowak?

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -50,13 +50,51 @@ public static void RequestFailed(this ILogger logger, HttpContext httpContext, i
...
+        }
+
+        public static void Shutdown(this ILogger logger)
+        {
+            if (logger.IsEnabled(LogLevel.Information))
+            {
+                logger.LogInformation(
+                    eventId: LoggerEventIds.Shutdown,
+                    message: "Hosting shutdown");
+            }
```
rynowak: None of these should be information

victorhurdugaci:
Updated

rynowak:
:shipit:

PR #448, opened 2015-10-26
Use "webroot" instead of "hosting:webroot"

Tratcher:
#411 @davidfowl @DamianEdwards @muratg @BrennanConroy


muratg:
:shipit:  @BrennanConroy @phenning

muratg:
@Tratcher please follow up on the DNX and Template changes as well.  Thanks!

PR #455, opened 2015-10-28
[Fixes #454] Log messages from HostingEngine are sometimes null 

### LINKED ISSUE #454: Log messages from HostingEngine are sometimes null
Example line of code causing the issue: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs#L140

Suggestion (from @rynowak ): Change it to a lazy initialize pattern ex:
```
If (_cached == null)
{
  _cached = …;
}
```

Suggestion (from @loudejard): Add the missing piece
```
public override string ToString() => 
_cachedToString ?? Interlocked.CompareExchange(
                ref _cachedToString,
                $"Request starting {_httpContext.Request.Protocol} {_httpContext.Request.Method} {_httpContext.Request.Scheme}://{_httpContext.Request.Host}{_httpContext.Request.PathBase}{_httpContext.Request.Path}{_httpContext.Request.QueryString} {_httpContext.Request.ContentType} {_httpContext.Request.ContentLength}",
                null) ?? _cachedToString;
```

The `CompareExchange` will return `null` when it's successful, but we want it to return the new value (now stored in `_cachedToString`).

@loudej @lodejard

kichalla:
@rynowak @loudej @lodejard @davidfowl 

I have tested this with the Diagnostics repo where the tests were failing and also on the other thread where we the scope was not being printed.

Also tested on Mono on Linux (however there seems to be an issue with Diagnostics repo on Linux which seems to be unrelated to my fix...will talk to @rowanmiller about it)

cc: @JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
@@ -172,7 +172,7 @@ namespace Microsoft.AspNet.Hosting.Internal
                         new KeyValuePair<string, object>("Path", _request.Path.ToString()),
                         new KeyValuePair<string, object>("QueryString", _request.QueryString.ToString()),
                     };
-                };
+                }
 
                 return _cachedGetValues;
             }
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
@@ -108,55 +108,75 @@ namespace Microsoft.AspNet.Hosting.Internal
                 _httpContext = httpContext;
             }
 
-            public override string ToString() => _cachedToString ?? Interlocked.CompareExchange(
-                ref _cachedToString,
-                $"RequestId:{_httpContext.TraceIdentifier} RequestPath:{_httpContext.Request.Path}",
-                null);
-
-            public IEnumerable<KeyValuePair<string, object>> GetValues() => _cachedGetValues ?? Interlocked.CompareExchange(
-                ref _cachedGetValues,
-                new[]
+            public override string ToString()
+            {
+                if (_cachedToString == null)
                 {
-                    new KeyValuePair<string, object>("RequestId", _httpContext.TraceIdentifier),
-                    new KeyValuePair<string, object>("RequestPath", _httpContext.Request.Path.ToString()),
-                },
-                null);
+                    _cachedToString = $"RequestId:{_httpContext.TraceIdentifier} RequestPath:{_httpContext.Request.Path}";
+                }
+
+                return _cachedToString;
+            }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -108,55 +107,75 @@ public HostingLogScope(HttpContext httpContext)
...
+                        new KeyValuePair<string, object>("Method", _request.Method),
+                        new KeyValuePair<string, object>("ContentType", _request.ContentType),
+                        new KeyValuePair<string, object>("ContentLength", _request.ContentLength),
+                        new KeyValuePair<string, object>("Scheme", _request.Scheme.ToString()),
+                        new KeyValuePair<string, object>("Host", _request.Host.ToString()),
+                        new KeyValuePair<string, object>("PathBase", _request.PathBase.ToString()),
+                        new KeyValuePair<string, object>("Path", _request.Path.ToString()),
+                        new KeyValuePair<string, object>("QueryString", _request.QueryString.ToString()),
+                    };
+                };
```
JunTaoLuo: semicolon?

rowanmiller:
@kichalla that test won't run on Linux because it uses a LocalDb instance (separate issue already filed to enable them on another database engine). I'll send a PR to skip it.

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -108,55 +107,75 @@ public HostingLogScope(HttpContext httpContext)
...
+                return _cachedGetValues;
+            }
         }
 
         private class HostingRequestStarting : ILogValues
         {
             internal static readonly Func<object, Exception, string> Callback = (state, exception) => ((HostingRequestStarting)state).ToString();
 
-            private readonly HttpContext _httpContext;
+            private readonly HttpRequest _request;
```
JunTaoLuo: This is nice, but why only in request starting? For example, we can do something similar for `_httpContext.Response` in `HostingRequestFinished`.

kichalla: The line for `ToString` was extremely long where as for others they are ok

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
kichalla: good catch...apparently the compiler seems ok with this :smile: ...will remove

JunTaoLuo:
:shipit: from me. @pranavkm we are branching so this change will need to go into both release and dev pending approval?

rynowak:
:shipit:

kichalla:
2bb2f955327ea52c4b0536732d85b4a42f16e1a5

PR #457, opened 2015-10-28
Get the current runtime path via PlatformServices.Default

halter73:
- This fixes an issue whith ApplicationDeployer on Mono.
  Instead of giving the dnx path, Process.GetCurrentProcess().MainModule.FileName
  would give the path to the Mono executable.


davidfowl:
Yesssss

Tratcher:
Lovely, :shipit:

kichalla:
> This fixes an issue whith ApplicationDeployer on Mono
But this part of code is not used by the deployer on Mono...like there is a MonoDeployer which does not call into this, so I am a bit unclear

halter73:
@kichalla This is still used when the test runner is on mono but the target runtime is CoreClr. In this case ApplicationDeployerFactory.Create selects the SelfHostDeployer.

kichalla:
:shipit:

PR #463, opened 2015-11-03
New IServer design

JunTaoLuo:
#395 I'm not sure if I understand the new design completely but here's what I currently have.

Any thoughts? @davidfowl @Tratcher @troydai @lodejard @rynowak

### DIFF
--- a/src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
+++ b/src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Hosting.Server
+{
+    /// <summary>
+    /// Represents an HttpApplication.
+    /// </summary>
+    public interface IHttpApplication<TContext>
+    {
+        /// <summary>
+        /// Create a TContext given a collection of HTTP features.
+        /// </summary>
+        /// <param name="contextFeatures">A collection of HTTP features to be used for creating the TContext.</param>
+        /// <returns>The created TContext.</returns>
+        TContext CreateContext(IFeatureCollection contextFeatures);
+
+        /// <summary>
+        /// Asynchronously processes an TContext.
+        /// </summary>
+        /// <param name="context">The TContext that the operation will process.</param>
+        Task ProcessRequestAsync(TContext context);
+
+        /// <summary>
+        /// Dispose a given TContext.
+        /// </summary>
+        /// <param name="context">The TContext to be disposed.</param>
+        /// <param name="exception">The Exception thrown when processing did not complete successfully, otherwise null.</param>
+        void DisposeContext(TContext context, Exception exception);
+    }
+}
--- a/src/Microsoft.AspNet.Hosting.Server.Abstractions/IServer.cs
+++ b/src/Microsoft.AspNet.Hosting.Server.Abstractions/IServer.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
 
 namespace Microsoft.AspNet.Hosting.Server
@@ -18,9 +17,9 @@ namespace Microsoft.AspNet.Hosting.Server
         IFeatureCollection Features { get; }
 
         /// <summary>
-        /// Start the server with the given function that processes an HTTP request.
+        /// Start the server with an HttpApplication.
         /// </summary>
-        /// <param name="requestDelegate">A function that processes an HTTP request.</param>
-        void Start(RequestDelegate requestDelegate);
+        /// <param name="application">An instance of <see cref="IHttpApplication"/>.</param>
+        void Start<TContext>(IHttpApplication<TContext> application);
     }
 }
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
@@ -1,0 +1,91 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Hosting.Internal
+{
+    public class HostingApplication : IHttpApplication<HostingApplication.Context>
+    {
+        private readonly RequestDelegate _application;
+        private readonly ILogger _logger;
+        private readonly DiagnosticSource _diagnosticSource;
+        private readonly IHttpContextFactory _httpContextFactory;
+
+        public HostingApplication(
+            RequestDelegate application,
+            ILogger logger,
+            DiagnosticSource diagnosticSource,
+            IHttpContextFactory httpContextFactory)
+        {
+            _application = application;
+            _logger = logger;
+            _diagnosticSource = diagnosticSource;
+            _httpContextFactory = httpContextFactory;
+        }
+
+        public Context CreateContext(IFeatureCollection contextFeatures)
+        {
+            var httpContext = _httpContextFactory.Create(contextFeatures);
+            var startTick = Environment.TickCount;
+
+            var scope = _logger.RequestScope(httpContext);
+            _logger.RequestStarting(httpContext);
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
+            {
+                _diagnosticSource.Write("Microsoft.AspNet.Hosting.BeginRequest", new { httpContext = httpContext, tickCount = startTick });
+            }
+
+            return new Context
+            {
+                HttpContext = httpContext,
+                Scope = scope,
+                StartTick = startTick,
+            };
+        }
+
+        public void DisposeContext(Context context, Exception exception)
+        {
+            var httpContext = context.HttpContext;
+            var currentTick = Environment.TickCount;
+            _logger.RequestFinished(httpContext, context.StartTick, currentTick);
+
+            if (exception == null)
+            {
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.EndRequest"))
+                {
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.EndRequest", new { httpContext = httpContext, tickCount = currentTick });
+                }
+            }
+            else
+            {
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
+                {
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, tickCount = currentTick, exception = exception });
+                }
+            }
+
+            context.Scope.Dispose();
+
+            _httpContextFactory.Dispose(httpContext);
+        }
+
+        public async Task ProcessRequestAsync(Context context)
+        {
+            await _application(context.HttpContext);
+        }
+
+        public struct Context
+        {
+            public HttpContext HttpContext { get; set; }
+            public IDisposable Scope { get; set; }
+            public int StartTick { get; set; }
+        }
+    }
+}
--- a/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
@@ -10,7 +10,6 @@ using Microsoft.AspNet.Hosting.Server;
 using Microsoft.AspNet.Hosting.Startup;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Internal;
-using Microsoft.Extensions.CompilationAbstractions;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
--- a/src/Microsoft.AspNet.TestHost/ClientHandler.cs
+++ b/src/Microsoft.AspNet.TestHost/ClientHandler.cs
@@ -11,9 +11,10 @@ using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Hosting.Server;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
-using Microsoft.AspNet.Http.Internal;
+using Context = Microsoft.AspNet.Hosting.Internal.HostingApplication.Context;
 
 namespace Microsoft.AspNet.TestHost
 {
@@ -23,27 +24,21 @@ namespace Microsoft.AspNet.TestHost
     /// </summary>
     public class ClientHandler : HttpMessageHandler
     {
-        private readonly RequestDelegate _next;
+        private readonly IHttpApplication<Context> _application;
         private readonly PathString _pathBase;
-        private readonly IHttpContextFactory _factory;
 
         /// <summary>
         /// Create a new handler.
         /// </summary>
         /// <param name="next">The pipeline entry point.</param>
-        public ClientHandler(RequestDelegate next, PathString pathBase, IHttpContextFactory httpContextFactory)
+        public ClientHandler(PathString pathBase, IHttpApplication<Context> application)
         {
-            if (next == null)
+            if (application == null)
             {
-                throw new ArgumentNullException(nameof(next));
+                throw new ArgumentNullException(nameof(application));
             }
-            if (httpContextFactory == null)
-            {
-                throw new ArgumentNullException(nameof(httpContextFactory));
-            }
-
-            _next = next;
-            _factory = httpContextFactory;
+            
+            _application = application;
 
             // PathString.StartsWithSegments that we use below requires the base path to not end in a slash.
             if (pathBase.HasValue && pathBase.Value.EndsWith("/"))
@@ -69,7 +64,7 @@ namespace Microsoft.AspNet.TestHost
                 throw new ArgumentNullException(nameof(request));
             }
 
-            var state = new RequestState(request, _pathBase, _factory);
+            var state = new RequestState(request, _pathBase, _application);
             var requestContent = request.Content ?? new StreamContent(Stream.Null);
             var body = await requestContent.ReadAsStreamAsync();
             if (body.CanSeek)
@@ -77,7 +72,7 @@ namespace Microsoft.AspNet.TestHost
                 // This body may have been consumed before, rewind it.
                 body.Seek(0, SeekOrigin.Begin);
             }
-            state.HttpContext.Request.Body = body;
+            state.Context.HttpContext.Request.Body = body;
             var registration = cancellationToken.Register(state.AbortRequest);
 
             // Async offload, don't let the test code block the caller.
@@ -85,16 +80,17 @@ namespace Microsoft.AspNet.TestHost
                 {
                     try
                     {
-                        await _next(state.HttpContext);
+                        await _application.ProcessRequestAsync(state.Context);
                         state.CompleteResponse();
+                        state.ServerCleanup(exception: null);
                     }
                     catch (Exception ex)
                     {
                         state.Abort(ex);
+                        state.ServerCleanup(ex);
                     }
                     finally
                     {
-                        state.ServerCleanup();
                         registration.Dispose();
                     }
                 });
@@ -105,20 +101,20 @@ namespace Microsoft.AspNet.TestHost
         private class RequestState
         {
             private readonly HttpRequestMessage _request;
+            private readonly IHttpApplication<Context> _application;
             private TaskCompletionSource<HttpResponseMessage> _responseTcs;
             private ResponseStream _responseStream;
             private ResponseFeature _responseFeature;
             private CancellationTokenSource _requestAbortedSource;
-            private IHttpContextFactory _factory;
             private bool _pipelineFinished;
 
-            internal RequestState(HttpRequestMessage request, PathString pathBase, IHttpContextFactory factory)
+            internal RequestState(HttpRequestMessage request, PathString pathBase, IHttpApplication<Context> application)
             {
                 _request = request;
+                _application = application;
                 _responseTcs = new TaskCompletionSource<HttpResponseMessage>();
                 _requestAbortedSource = new CancellationTokenSource();
                 _pipelineFinished = false;
-                _factory = factory;
 
                 if (request.RequestUri.IsDefaultPort)
                 {
@@ -129,12 +125,13 @@ namespace Microsoft.AspNet.TestHost
                     request.Headers.Host = request.RequestUri.GetComponents(UriComponents.HostAndPort, UriFormat.UriEscaped);
                 }
 
-                HttpContext = _factory.Create(new FeatureCollection());
-                
-                HttpContext.Features.Set<IHttpRequestFeature>(new RequestFeature());
+                Context = application.CreateContext(new FeatureCollection());
+                var httpContext = Context.HttpContext;
+
+                httpContext.Features.Set<IHttpRequestFeature>(new RequestFeature());
                 _responseFeature = new ResponseFeature();
-                HttpContext.Features.Set<IHttpResponseFeature>(_responseFeature);
-                var serverRequest = HttpContext.Request;
+                httpContext.Features.Set<IHttpResponseFeature>(_responseFeature);
+                var serverRequest = httpContext.Request;
                 serverRequest.Protocol = "HTTP/" + request.Version.ToString(2);
                 serverRequest.Scheme = request.RequestUri.Scheme;
                 serverRequest.Method = request.Method.ToString();
@@ -168,12 +165,12 @@ namespace Microsoft.AspNet.TestHost
                 }
 
                 _responseStream = new ResponseStream(ReturnResponseMessage, AbortRequest);
-                HttpContext.Response.Body = _responseStream;
-                HttpContext.Response.StatusCode = 200;
-                HttpContext.RequestAborted = _requestAbortedSource.Token;
+                httpContext.Response.Body = _responseStream;
+                httpContext.Response.StatusCode = 200;
+                httpContext.RequestAborted = _requestAbortedSource.Token;
             }
 
-            public HttpContext HttpContext { get; private set; }
+            public Context Context { get; private set; }
 
             public Task<HttpResponseMessage> ResponseTask
             {
@@ -212,16 +209,17 @@ namespace Microsoft.AspNet.TestHost
             private HttpResponseMessage GenerateResponse()
             {
                 _responseFeature.FireOnSendingHeaders();
+                var httpContext = Context.HttpContext;
 
                 var response = new HttpResponseMessage();
-                response.StatusCode = (HttpStatusCode)HttpContext.Response.StatusCode;
-                response.ReasonPhrase = HttpContext.Features.Get<IHttpResponseFeature>().ReasonPhrase;
+                response.StatusCode = (HttpStatusCode)httpContext.Response.StatusCode;
+                response.ReasonPhrase = httpContext.Features.Get<IHttpResponseFeature>().ReasonPhrase;
                 response.RequestMessage = _request;
                 // response.Version = owinResponse.Protocol;
 
                 response.Content = new StreamContent(_responseStream);
 
-                foreach (var header in HttpContext.Response.Headers)
+                foreach (var header in httpContext.Response.Headers)
                 {
                     if (!response.Headers.TryAddWithoutValidation(header.Key, (IEnumerable<string>)header.Value))
                     {
@@ -239,12 +237,9 @@ namespace Microsoft.AspNet.TestHost
                 _responseTcs.TrySetException(exception);
             }
 
-            internal void ServerCleanup()
+            internal void ServerCleanup(Exception exception)
             {
-                if (HttpContext != null)
-                {
-                    _factory.Dispose(HttpContext);
-                }
+                _application.DisposeContext(Context, exception);
             }
         }
     }
--- a/src/Microsoft.AspNet.TestHost/TestServer.cs
+++ b/src/Microsoft.AspNet.TestHost/TestServer.cs
@@ -11,6 +11,7 @@ using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Context = Microsoft.AspNet.Hosting.Internal.HostingApplication.Context;
 
 namespace Microsoft.AspNet.TestHost
 {
@@ -18,15 +19,13 @@ namespace Microsoft.AspNet.TestHost
     {
         private const string DefaultEnvironmentName = "Development";
         private const string ServerName = nameof(TestServer);
-        private RequestDelegate _appDelegate;
         private IDisposable _appInstance;
         private bool _disposed = false;
-        private IHttpContextFactory _httpContextFactory;
+        private IHttpApplication<Context> _application;
 
         public TestServer(WebHostBuilder builder)
         {
             var hostingEngine = builder.UseServer(this).Build();
-            _httpContextFactory = hostingEngine.ApplicationServices.GetService<IHttpContextFactory>();
             _appInstance = hostingEngine.Start();
         }
 
@@ -99,7 +98,7 @@ namespace Microsoft.AspNet.TestHost
         public HttpMessageHandler CreateHandler()
         {
             var pathBase = BaseAddress == null ? PathString.Empty : PathString.FromUriComponent(BaseAddress);
-            return new ClientHandler(Invoke, pathBase, _httpContextFactory);
+            return new ClientHandler(pathBase, _application);
         }
 
         public HttpClient CreateClient()
@@ -110,7 +109,7 @@ namespace Microsoft.AspNet.TestHost
         public WebSocketClient CreateWebSocketClient()
         {
             var pathBase = BaseAddress == null ? PathString.Empty : PathString.FromUriComponent(BaseAddress);
-            return new WebSocketClient(Invoke, pathBase, _httpContextFactory);
+            return new WebSocketClient(pathBase, _application);
         }
 
         /// <summary>
@@ -123,24 +122,49 @@ namespace Microsoft.AspNet.TestHost
             return new RequestBuilder(this, path);
         }
 
-        public Task Invoke(HttpContext context)
-        {
-            if (_disposed)
-            {
-                throw new ObjectDisposedException(GetType().FullName);
-            }
-            return _appDelegate(context);
-        }
-
         public void Dispose()
         {
             _disposed = true;
             _appInstance.Dispose();
         }
 
-        void IServer.Start(RequestDelegate requestDelegate)
+        void IServer.Start<TContext>(IHttpApplication<TContext> application)
         {
-            _appDelegate = requestDelegate;
+            _application = new ApplicationWrapper<Context>((IHttpApplication<Context>)application, () =>
+            {
+                if (_disposed)
+                {
+                    throw new ObjectDisposedException(GetType().FullName);
+                }
+            });
+        }
+
+        private class ApplicationWrapper<TContext> : IHttpApplication<TContext>
+        {
+            IHttpApplication<TContext> _application;
+            Action _preProcessRequestAsync;
+
+            public ApplicationWrapper(IHttpApplication<TContext> application, Action preProcessRequestAsync)
+            {
+                _application = application;
+                _preProcessRequestAsync = preProcessRequestAsync;
+            }
+
+            public TContext CreateContext(IFeatureCollection contextFeatures)
+            {
+                return _application.CreateContext(contextFeatures);
+            }
+
+            public void DisposeContext(TContext context, Exception exception)
+            {
+                _application.DisposeContext(context, exception);
+            }
+
+            public Task ProcessRequestAsync(TContext context)
+            {
+                _preProcessRequestAsync();
+                return _application.ProcessRequestAsync(context);
+            }
         }
     }
 }
--- a/src/Microsoft.AspNet.TestHost/WebSocketClient.cs
+++ b/src/Microsoft.AspNet.TestHost/WebSocketClient.cs
@@ -8,31 +8,26 @@ using System.Net.WebSockets;
 using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Hosting.Server;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Features;
-using Microsoft.AspNet.Http.Internal;
+using Context = Microsoft.AspNet.Hosting.Internal.HostingApplication.Context;
 
 namespace Microsoft.AspNet.TestHost
 {
     public class WebSocketClient
     {
-        private readonly RequestDelegate _next;
+        private readonly IHttpApplication<Context> _application;
         private readonly PathString _pathBase;
-        private readonly IHttpContextFactory _httpContextFactory;
 
-        internal WebSocketClient(RequestDelegate next, PathString pathBase, IHttpContextFactory httpContextFactory)
+        internal WebSocketClient(PathString pathBase, IHttpApplication<Context> application)
         {
-            if (next == null)
+            if (application == null)
             {
-                throw new ArgumentNullException(nameof(next));
+                throw new ArgumentNullException(nameof(application));
             }
-            if (httpContextFactory == null)
-            {
-                throw new ArgumentNullException(nameof(httpContextFactory));
-            }
-
-            _next = next;
-            _httpContextFactory = httpContextFactory;
+            
+            _application = application;
 
             // PathString.StartsWithSegments that we use below requires the base path to not end in a slash.
             if (pathBase.HasValue && pathBase.Value.EndsWith("/"))
@@ -58,11 +53,11 @@ namespace Microsoft.AspNet.TestHost
 
         public async Task<WebSocket> ConnectAsync(Uri uri, CancellationToken cancellationToken)
         {
-            var state = new RequestState(uri, _pathBase, cancellationToken, _httpContextFactory);
+            var state = new RequestState(uri, _pathBase, cancellationToken, _application);
 
             if (ConfigureRequest != null)
             {
-                ConfigureRequest(state.HttpContext.Request);
+                ConfigureRequest(state.Context.HttpContext.Request);
             }
 
             // Async offload, don't let the test code block the caller.
@@ -70,12 +65,14 @@ namespace Microsoft.AspNet.TestHost
             {
                 try
                 {
-                    await _next(state.HttpContext);
+                    await _application.ProcessRequestAsync(state.Context);
                     state.PipelineComplete();
+                    state.ServerCleanup(exception: null);
                 }
                 catch (Exception ex)
                 {
                     state.PipelineFailed(ex);
+                    state.ServerCleanup(ex);
                 }
                 finally
                 {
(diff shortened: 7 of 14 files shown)


src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
```diff
@@ -163,6 +166,17 @@ public WebHostBuilder UseEnvironment(string environment)
             return this;
         }
 
+        public WebHostBuilder UseApplication(IHttpApplication application)
```
JunTaoLuo: Not sure if this is useful but since we separated Server and HttpApplication, there maybe cases where we want to specify a non-default application?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,82 @@
...
+            _applicationServices = applicationServices;
+            _application = application;
+            _logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
+            _diagnosticSource = _applicationServices.GetRequiredService<DiagnosticSource>();
+            _httpContextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
+        }
+
+        public HttpContext CreateHttpContext(IFeatureCollection contextFeatures)
+        {
+            return _httpContextFactory.Create(contextFeatures);
```
Tratcher: @lodejard @davidfowl When we add logging here we'll be able to log RequestStarted and RequestFinished, but we still won't have a way to log RequestException. Do we need another method? Or an optional Exception parameter on DisposeHttpContext?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,82 @@
...
+        public async Task InvokeAsync(HttpContext httpContext)
+        {
+            httpContext.ApplicationServices = _applicationServices;
+
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
+            {
+                _diagnosticSource.Write("Microsoft.AspNet.Hosting.BeginRequest", new { httpContext = httpContext });
+            }
+
+            using (_logger.RequestScope(httpContext))
```
Tratcher: If we log RequestStarted in CreateHttpContext, should we also create the scope there so the log can be within the scope? Same for RequestFinished and DisposeHttpContext.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,82 @@
...
+        public void DisposeHttpContext(HttpContext httpContext)
+        {
+            _httpContextFactory.Dispose(httpContext);
+        }
+
+        public async Task InvokeAsync(HttpContext httpContext)
+        {
+            httpContext.ApplicationServices = _applicationServices;
+
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
```
Tratcher: This can move to CreateHttpContext

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,82 @@
...
+                    _logger.RequestFailed(httpContext, startTime);
+
+                    if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
+                    {
+                        _diagnosticSource.Write("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
+                    }
+                    throw;
+                }
+            }
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.EndRequest"))
```
Tratcher: This could move to DisposeHttpContext, but only if we can distinguish between successful responses and unhandled exceptions.

src/Microsoft.AspNet.Hosting/WebHostBuilder.cs
davidfowl: This isn't required

Tratcher: Let's not go there yet. If we were going to do this you'd want to bypass running Startup completely.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -138,9 +138,9 @@ public void Dispose()
             _appInstance.Dispose();
         }
 
-        void IServer.Start(RequestDelegate requestDelegate)
+        void IServer.Start(IHttpApplication app)
```
Tratcher: TestServer should use the full interface.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Hosting.Server
+{
+    /// <summary>
+    /// Represents an HttpApplication.
+    /// </summary>
+    public interface IHttpApplication
```
Tratcher: @davidfowl, isn't one of the points of this to let servers avoid a dependency on Http.Abstractions?

Tratcher:
@davidfowl @lodejard, the design still needs some work... Especially with regards to logging.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
davidfowl: Yes it's supposed to be object instead of http context. The sever shouldn't understand the context object

JunTaoLuo: So cast from object everywhere?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
troydai: Should we add an additional parameter to `DisposeHttpContext` to indicate if the operation is successful?

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
troydai: Then the method shouldn't be named `CreateHttpContext` or `DisposeHttpContext`, right?

JunTaoLuo: I think the naming still reflects what they do. It creates and returns or disposes an `HttpContext`, it just represent them as objects.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
davidfowl: I think we may want to consider doing on error as well /cc @lodejard

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
davidfowl: Is IHttpLifetime a better name

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
lodejard: Yep, the scope should be created/disposed at the same two places the started/finished are logged

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
lodejard: IHttpHandler? :)

IHttpRequestHandler?

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Create an HttpContext given a collection of HTTP features.
+        /// </summary>
+        /// <param name="contextFeatures">A collection of HTTP features to be used for creating the HttpContext.</param>
+        /// <returns>The created HttpContext.</returns>
+        HttpContext CreateHttpContext(IFeatureCollection contextFeatures);
```
lodejard: HttpContext in all 3 places should be object.

@davidfowl What is the reason it couldn't be a `<TContext>` on the interface and Start method?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,82 @@
...
+        private readonly DiagnosticSource _diagnosticSource;
+        private readonly IHttpContextFactory _httpContextFactory;
+
+        public HostingApplication(
+            IServiceProvider applicationServices,
+            RequestDelegate application)
+        {
+            _applicationServices = applicationServices;
+            _application = application;
+            _logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
```
lodejard: Pass all of the services in explicitly - calling GetRequiredService repeatedly is inefficient

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
lodejard: I dunno - if the server calls LogWarning(ex) it'll work perfectly well. Is it that we want to have the same category & eventid on unhandled exception?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,82 @@
...
+                    _logger.RequestStarting(httpContext);
+
+                    startTime = Environment.TickCount;
+                    await _application(httpContext);
+
+                    _logger.RequestFinished(httpContext, startTime);
+                }
+                catch (Exception ex)
+                {
+                    _logger.RequestFailed(httpContext, startTime);
```
lodejard: removed RequestFailed altogether - "log and rethrow" is an antipattern

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                int startTime = 0;
+                try
+                {
+                    _logger.RequestStarting(httpContext);
+
+                    startTime = Environment.TickCount;
+                    await _application(httpContext);
+
+                    _logger.RequestFinished(httpContext, startTime);
```
lodejard: RequestFinished is called both for successful and for failed requests. For log analysis a user should expect to see exactly one starting and finished pair for every request which passes through the server

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -138,9 +138,9 @@ public void Dispose()
             _appInstance.Dispose();
         }
 
-        void IServer.Start(RequestDelegate requestDelegate)
+        public void Start(IHttpApplication app)
         {
-            _appDelegate = requestDelegate;
+            _appDelegate = app.InvokeAsync;
```
lodejard: it seems like the test server should all all three methods, shouldn't it?

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
davidfowl: Bahahah, :smile:

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
davidfowl: Who specifies the T when the server receives the interface?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
davidfowl: We can do the telemetry in hosting and let the server log the exception.

src/Microsoft.AspNet.TestHost/TestServer.cs
JunTaoLuo: Might as well just retain the application and call the three methods?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
JunTaoLuo: Do we still want the diagnostic source message? I'm assuming we do.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
JunTaoLuo: @davidfowl @lodejard Can we finalize a design for exception handling?

JunTaoLuo:
:up: :date: 

Still need to finalize the design for exception handling.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Represents an HttpApplication.
+    /// </summary>
+    public interface IHttpApplication
+    {
+        /// <summary>
+        /// Create an HttpContext given a collection of HTTP features.
+        /// </summary>
+        /// <param name="contextFeatures">A collection of HTTP features to be used for creating the HttpContext.</param>
+        /// <returns>The created HttpContext.</returns>
+        object CreateHttpContext(object contextFeatures);
```
Tratcher: The input should still be IFeatureCollection

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,90 @@
...
+            var httpContext = _httpContextFactory.Create((IFeatureCollection)contextFeatures);
+            
+            httpContext.ApplicationServices = _applicationServices;
+
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
+            {
+                _diagnosticSource.Write("Microsoft.AspNet.Hosting.BeginRequest", new { httpContext = httpContext });
+            }
+
+            _logScope = _logger.RequestScope(httpContext);
```
Tratcher: You can't store per-request state in HostingApplication, all requests go through here. We need to find a way to return this scope with or inside of the httpContext.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
lodejard: Yeah, :)

But is IHttpRequestHandler a better name than IHttpApplication? It's a minor API think of course, nothing to hold up the PR for

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -138,9 +136,9 @@ public void Dispose()
             _appInstance.Dispose();
         }
 
-        void IServer.Start(RequestDelegate requestDelegate)
+        public void Start(IHttpApplication app)
```
Tratcher: Keep this hidden under `IServer.`

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            _httpContextFactory.Dispose((HttpContext)httpContext);
+        }
+
+        public async Task InvokeAsync(object httpContext)
+        {
+            int startTime = 0;
+            try
+            {
+                _logger.RequestStarting((HttpContext)httpContext);
```
lodejard: Move _logger.RequestStarting to CreateHttpContext

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,90 @@
...
+
+                startTime = Environment.TickCount;
+                await _application((HttpContext)httpContext);
+
+                _logger.RequestFinished((HttpContext)httpContext, startTime);
```
lodejard: Move _logger.RequestFinished to DisposeHttpContext before the call to _logScope.Dispose

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+            catch (Exception ex)
+            {
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
+                {
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext = httpContext, exception = ex });
+                }
+                throw;
+            }
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.EndRequest"))
```
lodejard: Move `Microsoft.AspNet.Hosting.EndRequest` to DisposeHttpContext

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
JunTaoLuo: But that will add a dependency on `Microsoft.AspNet.Http.Features` which I thought we were going to merge into abstractions at some point?

Tratcher: Features is acceptable. We're only going to merge it with Http.Abstractions if this PR fails to make the separation useful.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
lodejard: Change the interface definition to 

```cs
public interface IHttpApplication<THttpContext>
{
  THttpContext CreateHttpContext(IFeatureCollection requestFeatures);
  Task InvokeAsync(THttpContext httpContext);
  void DisposeHttpContext(THttpContext httpContext);
}
//and
public interface IServer : IDisposable
{
  IFeatureCollection Features { get; }
  void Start<THttpContext>(IHttpApplication<THttpContext> app);
}
```
Then in the case of this Hosting assembly: create a struct `Q` which has two public fields, `IDisposable LoggerScope` and `HttpContext HttpContext`. So it is a zero-allocation value-type (tuple).

Then `HttpApplication : IHttpApplication<Q>` and you call `server.Start<Q>(app);`. The `Start<Q>` will probably also tell you that the Q can be inferred.

src/Microsoft.AspNet.TestHost/TestServer.cs
JunTaoLuo: So what did you mean by "implementing the full interface"?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
JunTaoLuo: We'll need to store the start time with `HttpContext` to keep track of the elapsed time per request.

lodejard: Calling RequestFinished before returning to the server means that the original bug isn't fixed (for example, the log doesn't have correct status code).

Store the start time as a third field in the `Q` per-request struct.

JunTaoLuo: Going to keep track of the scope and start time in a feature in the `HttpContext`

src/Microsoft.AspNet.Hosting/Internal/HostingApplicationFeature.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Hosting.Internal
+{
+    public class HostingApplicationFeature
+    {
+        public HostingApplicationFeature(IDisposable logScope, int startTime)
```
JunTaoLuo: This probably needs a better name...

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
HaoK: I think HttpRequestHandler is better, since its not really clear what exactly an HttpApplication is doing :)  Instead of, well duh, this handles the request!

HaoK: HostingRequestHandler sounds a lot better than HostingApplication too...

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// Asynchronously processes an HttpContext.
+        /// </summary>
+        /// <param name="httpContext">The HttpContext that the operation will process.</param>
+        Task InvokeAsync(object httpContext);
+
+        /// <summary>
+        /// Dispose a given HttpContext.
+        /// </summary>
+        /// <param name="httpContext">The HttpContext to be disposed.</param>
+        void DisposeHttpContext(object httpContext);
```
HaoK: Would it not be sufficiently clear for these to just be Create/DisposeContext?  The Http just looks a bit wordy to me...

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        /// <param name="contextFeatures">A collection of HTTP features to be used for creating the HttpContext.</param>
+        /// <returns>The created HttpContext.</returns>
+        object CreateHttpContext(IFeatureCollection contextFeatures);
+
+        /// <summary>
+        /// Asynchronously processes an HttpContext.
+        /// </summary>
+        /// <param name="httpContext">The HttpContext that the operation will process.</param>
+        Task InvokeAsync(object httpContext);
```
HaoK: Perhaps this should be called HandleRequest... Invoke is very generic sounding :)

src/Microsoft.AspNet.Hosting/Internal/HostingApplicationFeature.cs
HaoK: Is this feature's sole purpose to capture the lifetime of the request?  Maybe RequestLifetimeFeature?

lodejard: Nope - shouldn't be a feature. :) I'll push an example of what I was talking about

lodejard: See https://github.com/aspnet/Hosting/commit/dfae316be07d944433a013a181bc2b01837f55e5 for clarification about what the earlier note meant

Adding a feature to the http context would add more overhead, the approach in the example doesn't allocate any heap objects or add anything to feature dictionaries

JunTaoLuo: Thanks for the clarification, I'll look into this. @davidfowl @Tratcher just to confirm, is this the design we want to go with?

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
Tratcher: ProcessRequest?

src/Microsoft.AspNet.Hosting/Internal/HostingApplicationFeature.cs
Tratcher: Try it, we'll see how it works.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
JunTaoLuo: Agreed, since now we are going to return a generic context.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
HaoK: Yeah that's good :+1:

src/Microsoft.AspNet.Hosting.Server.Abstractions/IServer.cs
```diff
@@ -18,9 +17,9 @@ public interface IServer : IDisposable
         IFeatureCollection Features { get; }
 
         /// <summary>
-        /// Start the server with the given function that processes an HTTP request.
+        /// Start the server with an HttpApplication.
         /// </summary>
-        /// <param name="requestDelegate">A function that processes an HTTP request.</param>
-        void Start(RequestDelegate requestDelegate);
+        /// <param name="app">An instance of <see cref="IHttpApplication"/>.</param>
+        void Start<THttpContext>(IHttpApplication<THttpContext> app);
```
davidfowl: application

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -241,10 +234,6 @@ internal void Abort(Exception exception)
 
             internal void ServerCleanup()
             {
-                if (HttpContext != null)
```
davidfowl: What's happening here?

davidfowl:
@JunTaoLuo can you rebase?

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.Threading.Tasks;
+using Microsoft.AspNet.Http.Features;
+
+namespace Microsoft.AspNet.Hosting.Server
+{
+    /// <summary>
+    /// Represents an HttpApplication.
+    /// </summary>
+    public interface IHttpApplication<THttpContext>
```
davidfowl: Much prefer this design. I'm wondering if we should have BeginRequest/ProcessRequest/EndRequest. If we did that i'm not sure how to couple creation of the context object. Would it be on the side or interleaved? Maybe I'm just being anal :smile: 

```C#
var context = app.BeginRequest(features);

try
{
     await app.ProcessRequest(context);
}
finally
{
    app.EndRequest(context);
}
```

HaoK: Isn't it a bit weird that BeginRequest returns a context as far as naming go?  But maybe something super short wouldn't lose too much meaning?

```
var context = app.Initialize(features)
try {
    await app.Process(context)
}
finally
{
   app.Dispose(context)
}

**Comment by JunTaoLuo** [comment]
Still need consensus on how to differentiate between successful and non-successful requests for logging. I'm thinking something like this:

```c#
var context = app.BeginRequest(features);
try
{
    await app.ProcessRequest(context);
}
catch
{
    app.EndRequest(context, successful: false);
    throw;
}
app.EndRequest(context, successful: true);
```

This isn't as clean, any ideas? @davidfowl @lodejard @Tratcher

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
JunTaoLuo: As per offline discussion with @davidfowl, we are going to keep the current naming with the exception of renaming `ProcessRequest` to `ProcessRequestAsync`

Tratcher:
How about:
```
var context = app.BeginRequest(features);
try
{
    await app.ProcessRequest(context);
}
catch (Exception ex)
{
    app.EndRequest(context, ex);
    throw;
}
app.EndRequest(context);
```

JunTaoLuo:
:up::date:

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+            finally
+            {
+                httpContext.ApplicationServices = null;
+            }
+        }
+
+        public struct Context
+        {
+            public HttpContext HttpContext;
```
Tratcher: `{ get; set; }`?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,102 @@
...
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.EndRequest", new { httpContext });
+                }
+                _logger.RequestFinished(httpContext, elapsed);
+            }
+            else
+            {
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
+                {
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext, exception });
+                }
```
Tratcher: should logger.RequestFailed be called here? Or in ProcessRequestAsync? Or in the server? If the latter then remove it from Hosting.

JunTaoLuo: @lodejard said to remove it completely as it is an antipattern

Tratcher: Ok, so then make sure it's logged in all the servers. Also, remove it from HostingLoggerExtensions.

JunTaoLuo:
:up::date:

davidfowl:
:shipit: assuming travis is failing for a good reason?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        public Context CreateContext(IFeatureCollection contextFeatures)
+        {
+            var httpContext = _httpContextFactory.Create(contextFeatures);
+
+            var scope = _logger.RequestScope(httpContext);
+            _logger.RequestStarting(httpContext);
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
+            {
+                _diagnosticSource.Write("Microsoft.AspNet.Hosting.BeginRequest", new { httpContext });
```
davidfowl: change to httpContext = httpContext

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,102 @@
...
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.EndRequest"))
+                {
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.EndRequest", new { httpContext });
+                }
+            }
+            else
+            {
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.UnhandledException"))
+                {
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.UnhandledException", new { httpContext, exception });
```
davidfowl: httpContext = httpContext, exception = exception

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -138,9 +136,22 @@ public void Dispose()
             _appInstance.Dispose();
         }
 
-        void IServer.Start(RequestDelegate requestDelegate)
+        void IServer.Start<THttpContext>(IHttpApplication<THttpContext> app)
```
davidfowl: application

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
+            {
+                _diagnosticSource.Write("Microsoft.AspNet.Hosting.BeginRequest", new { httpContext });
+            }
+
+            return new Context
+            {
+                HttpContext = httpContext,
+                Scope = scope,
+                TickCount = Environment.TickCount,
```
Tratcher: StartTime?

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            var httpContext = context.HttpContext;
+            var elapsed = new TimeSpan(Environment.TickCount - context.TickCount);
+            _logger.RequestFinished(httpContext, elapsed);
+
+            if (exception == null)
+            {
+                if (_diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.EndRequest"))
+                {
+                    _diagnosticSource.Write("Microsoft.AspNet.Hosting.EndRequest", new { httpContext });
```
Tratcher: @avanderhoorn didn't you want the tick counts for these too?

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -129,7 +121,7 @@ internal RequestState(HttpRequestMessage request, PathString pathBase, IHttpCont
                     request.Headers.Host = request.RequestUri.GetComponents(UriComponents.HostAndPort, UriFormat.UriEscaped);
                 }
 
-                HttpContext = _factory.Create(new FeatureCollection());
+                HttpContext = new DefaultHttpContext();
```
Tratcher: Why is this reverting? Shouldn't it use IHttpApplication?

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -129,7 +127,7 @@ public Task Invoke(HttpContext context)
             {
                 throw new ObjectDisposedException(GetType().FullName);
             }
-            return _appDelegate(context);
+            return _requestHandler(context.Features);
```
Tratcher: It's strange to take an HttpContext here, pull out the features, and then create a new one.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
avanderhoorn: Yes please.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
avanderhoorn: Would be useful.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,102 @@
...
+                HttpContext = httpContext,
+                Scope = scope,
+                TickCount = Environment.TickCount,
+            };
+        }
+
+        public void DisposeContext(Context context, Exception exception = null)
+        {
+            var httpContext = context.HttpContext;
+            var elapsed = new TimeSpan(Environment.TickCount - context.TickCount);
```
PinpointTownes: Isn't this a bit fragile? IIRC, `Environment.TickCount` resets after ~25 days of uptime. If it happens in the middle of a request, the result could be really strange :smile:

benaadams:
Would this design allow for pooling at the connection level with `HttpContextFactory` as well as the server level? (I'm guessing using the correct DI scope would allow this?) For more session isolated pooling...

lodejard:
Probably just pooling at the server level, but because HttpContext implementation is essentially stateless in a way (it only references features, never holds values) having a reset on it should be very straightforward.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
lodejard: Wouldn't you be able to stamp the tick count on every message that arrives? Does having it as a field on this case matter?

src/Microsoft.AspNet.TestHost/ClientHandler.cs
lodejard: It didn't seem like the right layering in this particular place

src/Microsoft.AspNet.TestHost/TestServer.cs
lodejard: Kind of a side effect of having the client code pass in an httpcontext, and the app invocation mediated by any TContext that can be created from a feature collection. Could be an interesting way for a different hosting.dll to plug a server into an owin-dictionary TContext for example.

It also helps re-enforce that the httpcontext itself is just a facade over the feature collection. Any httpcontext that stored primary data in a way other than as feature state wouldn't work through the test server.

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
benaadams: Swap for something like?
```csharp
var elapsedTicks = Environment.TickCount - context.TickCount;
if (elapsedTicks < 0)
{
    elapsedTicks = (Int32.MaxValue - context.TickCount) + (Environment.TickCount - Int32.MinValue);
}
var elapsed = new TimeSpan(elapsedTicks);
```

JunTaoLuo: Nice catch.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
JunTaoLuo: This should use IHttpApplication. Here's another case where we would probably want to keep application as an `object` since it will need to be passed to `ClientHandler` and `WebSocketClient` to handle create/processes/dispose.

src/Microsoft.AspNet.TestHost/TestServer.cs
JunTaoLuo: Combined with the comment above, we should determine what the test server should be responsible for. Originally the `ClientHandler` and `WebSocketClient` handles context creation and dispose and Invoke only handles processing. Now Invoke handles all of that and as a side effect recreates a new default HttpContext every time from the features inside the given httpcontext.

davidfowl:
:shipit:

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        /// <param name="httpContext">The HttpContext that the operation will process.</param>
+        Task ProcessRequestAsync(THttpContext httpContext);
+
+        /// <summary>
+        /// Dispose a given HttpContext.
+        /// </summary>
+        /// <param name="httpContext">The HttpContext to be disposed.</param>
+        /// <param name="exception">The Exception thrown when processing did not complete successfully, otherwise null.</param>
+        void DisposeContext(THttpContext httpContext, Exception exception = null);
```
davidfowl: Remove the optional parameter

JunTaoLuo:
more :up::date:s

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// Asynchronously processes an HttpContext.
+        /// </summary>
+        /// <param name="context">The HttpContext that the operation will process.</param>
+        Task ProcessRequestAsync(object context);
+
+        /// <summary>
+        /// Dispose a given HttpContext.
+        /// </summary>
+        /// <param name="context">The HttpContext to be disposed.</param>
+        void DisposeContext(object context);
```
davidfowl: Remove this

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,106 @@
...
+                Scope = scope,
+                StartTick = startTick,
+            };
+        }
+
+        public void DisposeContext(Context context, Exception exception)
+        {
+            var httpContext = context.HttpContext;
+            var currentTick = Environment.TickCount;
+            var elapsed = new TimeSpan(currentTick < context.StartTick ?
```
JunTaoLuo: These calculations should move into RequestFinished so it's not run for every request when logging is disabled.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.AspNet.Http.Features;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Context = Microsoft.AspNet.Hosting.Internal.HostingApplication.Context;
```
JunTaoLuo: This is awkward since hard-coding the generic type seems to defeat the purpose of generics int he first place but this is the cleanest way of implementing the TestServer.

JunTaoLuo: Need to revisit the design of the clients (`ClientHandler` and `WebSocketClient`) to see why we need access to the internal fields of `Context`

JunTaoLuo:
:up::date:

src/Microsoft.AspNet.Hosting/Internal/HostingApplication.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Features;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Hosting.Internal
+{
+    public class HostingApplication : IHttpApplication<HostingApplication.Context>
+    {
+        private readonly IServiceProvider _applicationServices;
```
JunTaoLuo: This is no longer needed.

Tratcher:
@JunTaoLuo don't bother with an `object` based IHttpApplication branch or perf comparison, we've wasted too much time on that debate at this point.  This PR has reached the commit or delete deadline.

JunTaoLuo:
@Tratcher @davidfowl @lodejard let's see if there is anymore we need to do for this design and related reactions in other repos.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.AspNet.Hosting.Server
+{
+    /// <summary>
+    /// Represents an HttpApplication.
+    /// </summary>
+    public interface IHttpApplication<THttpContext>
+    {
+        /// <summary>
+        /// Create an HttpContext given a collection of HTTP features.
```
Tratcher: These comments are out of date, the result is not usually an HttpContext. TRequestState?

Tratcher: Or just TContext?

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -85,16 +86,17 @@ public ClientHandler(RequestDelegate next, PathString pathBase, IHttpContextFact
                 {
                     try
                     {
-                        await _next(state.HttpContext);
+                        await _next(state.Context);
+                        state.ServerCleanup(null);
```
Tratcher: use named parameters when passing in constants like null.

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -239,11 +243,11 @@ internal void Abort(Exception exception)
                 _responseTcs.TrySetException(exception);
             }
 
-            internal void ServerCleanup()
+            internal void ServerCleanup(Exception exception)
             {
-                if (HttpContext != null)
+                if (Context.HttpContext != null)
```
Tratcher: You shouldn't need this if check anymore.

src/Microsoft.AspNet.Hosting.Server.Abstractions/IHttpApplication.cs
JunTaoLuo: The reason I left it as that was because `THttpContext` sounds highly correlated to `HttpContext` and `IHttpApplication`. I imagine that it will always stay as a wrapper around `HttpContext` anyway so I didn't update the comments. I could change it to say `THttpContext` instead but I think `TContext` would be more appropriate if we also change the generic's name.

Tratcher: I was recommending to change the generic name to TContext.

src/Microsoft.AspNet.TestHost/TestServer.cs
```diff
@@ -123,13 +122,13 @@ public RequestBuilder CreateRequest(string path)
             return new RequestBuilder(this, path);
         }
 
-        public Task Invoke(HttpContext context)
+        public Task Invoke(Context context)
```
Tratcher: Is this still used anywhere? If not, get rid of it.

src/Microsoft.AspNet.TestHost/WebSocketClient.cs
```diff
@@ -153,16 +157,20 @@ public void PipelineFailed(Exception ex)
...
-                }
                 if (_serverWebSocket != null)
                 {
                     _serverWebSocket.Dispose();
                 }
             }
 
+            internal void ServerCleanup(Exception exception)
+            {
+                if (Context.HttpContext != null)
```
Tratcher: I don't think you need this null check

src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -23,27 +24,27 @@ namespace Microsoft.AspNet.TestHost
...
+        private readonly IHttpApplication<Context> _application;
         private readonly PathString _pathBase;
-        private readonly IHttpContextFactory _factory;
 
         /// <summary>
         /// Create a new handler.
         /// </summary>
         /// <param name="next">The pipeline entry point.</param>
-        public ClientHandler(RequestDelegate next, PathString pathBase, IHttpContextFactory httpContextFactory)
+        public ClientHandler(Func<Context, Task> next, PathString pathBase, IHttpApplication<Context> application)
```
Tratcher: next is redundant with application.ProcessRequest

JunTaoLuo:
:up::date:

test/Microsoft.AspNet.TestHost.Tests/ClientHandlerTests.cs
```diff
@@ -234,5 +243,33 @@ public Task ExceptionBeforeFirstWriteIsReported()
             var ex = await Assert.ThrowsAsync<HttpRequestException>(() => response.Content.ReadAsStringAsync());
             Assert.IsType<InvalidOperationException>(ex.GetBaseException());
         }
+
+        private class DummyApplication : IHttpApplication<Context>
+        {
+            Func<Context, Task> _application;
+
+            public DummyApplication(Func<Context, Task> application)
```
Tratcher: Is the full Context ever needed? It looks like all the tests just use HttpContext. Change this to `Func<HttpContext, Task>` / RequestDelegate?

test/Microsoft.AspNet.TestHost.Tests/TestServerTests.cs
```diff
@@ -393,6 +392,9 @@ public void CancelAborts()
             diagnosticListener.SubscribeWithAdapter(listener);
             var result = await server.CreateClient().GetStringAsync("/path");
 
+            // This ensures that all diagnostics are completely written to the diagnostic listener
+            Thread.Sleep(500);
```
Tratcher: 1000, some of the CI machines are slow.

test/Microsoft.AspNet.TestHost.Tests/ClientHandlerTests.cs
JunTaoLuo: `ClientHandler` is hardcoded to accept only `IHttpApplication<Context>` so the `DummyApplication` here must be of that type.

Tratcher: Yes, DummyApplication needs to be `IHttpApplication<Context>`, but the constructor can take `Func<HttpContext, Task>` to simplify the tests. The only difference is that `return _application(context);` changes to `return _application(context.HttpContext);`

Issue #453, opened 2015-10-28
Make hosting configuration consistency


davidfowl:
Right now we have a mix of config keys supported by the Hosting library. These grew organically and are inconsistent:

- Hosting:DetailedErrors - Show detailed errors for startup failures
- server or Hosting:Server - Specify which servers
- app or Hosting:App - Application where the startup is
- ASPNET_ENV or Hosting:Environment - The environment to run as (Development, Production etc)
- webroot - Webserver root, used by iis and static file servers

These values can be provided via `Microsoft.AspNet.Hosting.json` (soon to be `hosting.json`), command line options or environment variables.

Linux only supports `_` as the separator for environment key separator so Hosting:Foo=1 turns needs to be specified as `export Hosting__Foo=1`.

Proposal:

### Config file (Microsoft.AspNet.Hosting.json) 

```
{
   "webroot": "wwwroot",
   "server" : "Microsoft.AspNet.Server.Kestrel",
   "app": "MyProjectReference",
   "environment": "Development",
   "detailedErrors": true
}
```

### Command Line Options

`Microsoft.AspNet.Hosting --server Microsoft.AspNet.Server.Kestrel --webroot "wwwroot" --app "MyProjectReference" --environment "Development" --detailedErrors true`

The casing of `--detailedErrors` should really be `detailed-errors`, it's more unixy.

### Environment variables

These are global enough that it makes sense to make them prefixed:

Prefix=ASPNET

```
ASPNET_WEBROOT=wwwroot
ASPNET_SERVER=Microsoft.AspNet.Server.Kestrel
ASPNET_APP=MyProjectReference
ASPNET_ENVIRONMENT=Development
ASPNET_DETAILED_ERRORS=true
```

Prefix=HOSTING

```
HOSTING_WEBROOT=wwwroot
HOSTING_SERVER=Microsoft.AspNet.Server.Kestrel
HOSTING_APP=MyProjectReference
HOSTING_ENVIRONMENT=Development
HOSTING_DETAILED_ERRORS=true
```

Problems, the low level hosting API (WebHostBuilder) lets users pass in IConfiguration directly, what values are we reading from this IConfiguration? Is it prefixed? Or should it assume the names are all top level? 

I'd vote for top level names. Meaning that the above env variables would need to be transformed.

Tratcher:
What about server keys? e.g. server.urls?

[...]

PR #469, opened 2015-11-09
Make hosting configuration consistent

pakrym:
https://github.com/aspnet/Hosting/issues/453

@davidfowl @Tratcher @lodejard @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNet.Hosting/HostingEnvironmentExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/HostingEnvironmentExtensions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
 using Microsoft.AspNet.FileProviders;
 using Microsoft.Extensions.Configuration;
@@ -9,14 +10,14 @@ namespace Microsoft.AspNet.Hosting
 {
     public static class HostingEnvironmentExtensions
     {
-        private const string OldEnvironmentKey = "ASPNET_ENV";
-        private const string EnvironmentKey = "Hosting:Environment";
-
-        private const string WebRootKey = "webroot";
-
-        public static void Initialize(this IHostingEnvironment hostingEnvironment, string applicationBasePath, IConfiguration config)
+        public static void Initialize(this IHostingEnvironment hostingEnvironment, string applicationBasePath, WebHostOptions options)
         {
-            var webRoot = config?[WebRootKey];
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            var webRoot = options.WebRoot;
             if (webRoot == null)
             {
                 // Default to /wwwroot if it exists.
@@ -43,7 +44,7 @@ namespace Microsoft.AspNet.Hosting
             }
             hostingEnvironment.WebRootFileProvider = new PhysicalFileProvider(hostingEnvironment.WebRootPath);
 
-            var environmentName = config?[EnvironmentKey] ?? config?[OldEnvironmentKey];
+            var environmentName = options.Environment;
             hostingEnvironment.EnvironmentName = environmentName ?? hostingEnvironment.EnvironmentName;
         }
     }
--- a/src/Microsoft.AspNet.Hosting/WebApplication.cs
+++ b/src/Microsoft.AspNet.Hosting/WebApplication.cs
@@ -12,6 +12,7 @@ namespace Microsoft.AspNet.Hosting
     public class WebApplication
     {
         private const string HostingJsonFile = "hosting.json";
+        private const string EnvironmentVariablesPrefix = "ASPNET_";
         private const string ConfigFileKey = "config";
 
         public static void Run(string[] args)
@@ -40,12 +41,7 @@ namespace Microsoft.AspNet.Hosting
             var tempBuilder = new ConfigurationBuilder().AddCommandLine(args);
             var tempConfig = tempBuilder.Build();
             var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
-
-            var config = new ConfigurationBuilder()
-                .AddJsonFile(configFilePath, optional: true)
-                .AddEnvironmentVariables()
-                .AddCommandLine(args)
-                .Build();
+            var config = LoadHostingConfiguration(configFilePath, args);
 
             var hostBuilder = new WebHostBuilder(config, captureStartupErrors: true);
             if (startupType != null)
@@ -80,6 +76,15 @@ namespace Microsoft.AspNet.Hosting
 
                 appLifetime.ApplicationStopping.WaitHandle.WaitOne();
             }
+        }
+
+        internal static IConfiguration LoadHostingConfiguration(string configJsonPath, string[] args)
+        {
+            return new ConfigurationBuilder()
+                .AddJsonFile(configJsonPath, optional: true)
+                .AddEnvironmentVariables(prefix: EnvironmentVariablesPrefix)
+                .AddCommandLine(args)
+                .Build();
         }
     }
 }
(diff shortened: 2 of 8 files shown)


src/Microsoft.AspNet.Hosting/WebHostConfiguration.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public class WebHostConfiguration
```
HaoK: Perhaps consider making this a POCO and just using ConfigurationBinder to directly bind against config?

pakrym: @HaoK can it handle two keys for one property and 1/0 boolean property value?

Tratcher: Yeah, this isn't a pattern we use anywhere else.

HaoK: You could just have a string property that maps to the key in config, and then expose the current logic just like you do now.

```
public string DetailedErrors { get; set; }

public bool IsDetailed { get { <same logic as you have> }
```

pakrym:
:up: :date:

src/Microsoft.AspNet.Hosting/WebHostOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        private const string ApplicationKey = "app";
+        private const string DetailedErrorsKey = "detailederrors";
+        private const string EnvironmentVariablesPrefix = "ASPNET_";
+        private const string EnvironmentKey = "environment";
+        private const string OldEnvironmentKey = "ENV";
+        private const string ServerKey = "server";
+        private const string WebRootKey = "webroot";
+
+        public WebHostOptions(string application, bool detailedErrors, string environment, string server, string webRoot)
```
Tratcher: no params needed

src/Microsoft.AspNet.Hosting/WebHostOptions.cs
```diff
@@ -0,0 +1,54 @@
...
+        public bool DetailedErrors { get; set; }
+
+        public string Environment { get; set; }
+
+        public string Server { get; set; }
+
+        public string WebRoot { get; set; }
+
+        public static IConfiguration LoadHostingConfiguration(string configJsonPath, string[] args)
+        {
```
Tratcher: Move this back out

src/Microsoft.AspNet.Hosting/HostingEnvironmentExtensions.cs
```diff
@@ -9,14 +9,10 @@ namespace Microsoft.AspNet.Hosting
 {
     public static class HostingEnvironmentExtensions
     {
-        private const string OldEnvironmentKey = "ASPNET_ENV";
-        private const string EnvironmentKey = "Hosting:Environment";
 
-        private const string WebRootKey = "webroot";
-
-        public static void Initialize(this IHostingEnvironment hostingEnvironment, string applicationBasePath, IConfiguration config)
+        public static void Initialize(this IHostingEnvironment hostingEnvironment, string applicationBasePath, WebHostOptions config)
```
Tratcher: options

src/Microsoft.AspNet.Hosting/WebHostOptions.cs
```diff
@@ -0,0 +1,57 @@
...
+            Application = application;
+            DetailedErrors = detailedErrors;
+            Environment = environment;
+            Server = server;
+            WebRoot = webRoot;
+        }
+
+        public WebHostOptions(IConfiguration configuration)
+        {
+            Application = configuration[ApplicationKey];
```
Tratcher: config can be null

src/Microsoft.AspNet.Hosting/WebApplication.cs
```diff
@@ -40,12 +40,7 @@ public static void Run(Type startupType, string[] args)
             var tempBuilder = new ConfigurationBuilder().AddCommandLine(args);
             var tempConfig = tempBuilder.Build();
             var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
-
-            var config = new ConfigurationBuilder()
-                .AddJsonFile(configFilePath, optional: true)
-                .AddEnvironmentVariables()
-                .AddCommandLine(args)
```
Tratcher: revert

pakrym:
:up: :date:

src/Microsoft.AspNet.Hosting/WebHostOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public class WebHostOptions
+    {
+        private const string ApplicationKey = "app";
+        private const string DetailedErrorsKey = "detailederrors";
+        private const string EnvironmentKey = "environment";
+        private const string OldEnvironmentKey = "ENV";
```
davidfowl: nit:can you move this to the bottom

src/Microsoft.AspNet.Hosting/WebHostOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+using System;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public class WebHostOptions
+    {
+        private const string ApplicationKey = "app";
+        private const string DetailedErrorsKey = "detailederrors";
```
davidfowl: detailedErrors

test/Microsoft.AspNet.Hosting.Tests/WebApplicationTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                var config = WebApplication.LoadHostingConfiguration(fakeConfig.FileName, commandLine);
+
+                Assert.Equal("env.wwwroot", config["webroot"]);
+                Assert.Equal("Cmd.Microsoft.AspNet.Server.Kestrel", config["server"]);
+                Assert.Equal("Json.MyProjectReference", config["app"]);
+            }
+        }
+
+        private class FakeEnvironment : IDisposable
```
davidfowl: Nuke these tests

src/Microsoft.AspNet.Hosting/HostingEnvironmentExtensions.cs
```diff
@@ -9,14 +10,15 @@ namespace Microsoft.AspNet.Hosting
 {
     public static class HostingEnvironmentExtensions
     {
-        private const string OldEnvironmentKey = "ASPNET_ENV";
-        private const string EnvironmentKey = "Hosting:Environment";
```
davidfowl: Kill empty space

davidfowl:
Nuke those tests that change the env and :shipit:

src/Microsoft.AspNet.Hosting/WebHostOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            Application = configuration[ApplicationKey];
+            DetailedErrors = string.Equals("true", configuration[DetailedErrorsKey], StringComparison.OrdinalIgnoreCase)
+                || string.Equals("1", configuration[DetailedErrorsKey], StringComparison.OrdinalIgnoreCase);
+            Environment = configuration[EnvironmentKey] ?? configuration[OldEnvironmentKey];
+            Server = configuration[ServerKey];
+            WebRoot = configuration[WebRootKey];
```
HaoK: So if you wanted to use the config binder, it would look something more like this, the only weirdness is with the Environment/DetailedErrors properties which need a method to add the logic.  

```
public class WebHostOptions {
    public string App { get; set; }
    public string Server {get; set; }
   public string WebRoot {get; set; }
   public string DetailedErrors {get; set; }
   public string Environment {get; set; }
   public string Env {get; set; }

   public bool ShowDetailedErrors { get {
      return string.Equals("true", DetailedErrors)
                || string.Equals("1", DetailedErrors) ;
   } }

   public string ResolveEnvironment() {
      return Environment ?? Env;
   }
}

To get it out of config assuming its at the root level:
   config.Get<WebHostOptions>();
```

HaoK: The upside, is you can nuke the keys (the property names for the POCO are automatically the keys.  It would just work except for the weirdness around DetailedErrors and supporting the old Env key...

davidfowl:
@pakrym make an announcement

Issue #470, opened 2015-11-10
Regression: Not detecting HTTP_PLATFORM_PORT in IIS/Express


Tratcher:
Regressed by https://github.com/aspnet/Hosting/issues/453
The configuration is now filtering out environment variables that do not begin with "ASPNET_". `.AddEnvironmentVariables(prefix: EnvironmentVariablesPrefix)`.

Either re-work the config to pick up other environment variables, or read HTTP_PLATFORM_PORT directly from the Environment (not recommended as this makes testing more difficult). 
@davidfowl

davidfowl:
Did we add a test?

pakrym:
@davidfowl nope, it's environment variable and we cannot test them

PR #472, opened 2015-11-10
Fix environment variable loading in hosting #470

pakrym:
@Tratcher


src/Microsoft.AspNet.Hosting/WebApplication.cs
```diff
@@ -80,8 +80,10 @@ public static void Run(Type startupType, string[] args)
 
         internal static IConfiguration LoadHostingConfiguration(string configJsonPath, string[] args)
         {
+            // We are adding all environment variables first and then load ones starting with ASPNET_ stripping prefix
```
Tratcher: `We are adding all environment variables first and then adding the ASPNET_ ones with the prefix removed to unify with the command line and config file formats.` (multi-line)

Tratcher:
:shipit:

Issue #302, opened 2015-07-07
BUG: WebHostBuilder.UseEnvironment being overwritten on WebHostBuilder.Build


giggio:
If you set the environment using `WebHostBuilder.UseEnvironment` and later call `WebHostBuilder.Build` it will initialize using `HostingEnvironmentExtensions`:
https://github.com/aspnet/Hosting/blob/1.0.0-beta5/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs#L101
This will overwrite the setting applied with `UseEnvironment`, making this method useless.
Either remove the method, or rewrite the way it is working so that on `Build` the environment passed in by the user is used.

davidfowl:
@giggio Care to submit a pull request?

/cc @HaoK

giggio:
@davidfowl sure. I will change it in a way that the Build method works.

PR #474, opened 2015-11-11
Fix bug with environment name beeng overwritten

pakrym:
After being set using UseEnvironment

https://github.com/aspnet/Hosting/issues/302

/cc @davidfowl @Tratcher


davidfowl:
This need s a unit test

pakrym:
@davidfowl  done

Tratcher:
:shipit:

Issue #478, opened 2015-11-12
Logged elapsed time is wrong


sebastienros:
This is what I get when I log the time for a request, and then Hosting is also logging the same time:

```
verb: Orchard.Hosting.OrchardContainerMiddleware[0]
      Request took 42ms
info: Microsoft.AspNet.Hosting.Internal.HostingEngine[2]
      Request finished in 0.0047ms 200 text/html; charset=utf-8
```

Something in how this TimeSpan is constructed must be wrong.

JunTaoLuo:
@sebastienros I took a look at our logging and I don't see an obvious error. Do you have a project available for me to repro this and take a look at why there is a discrepancy between the logged times?

sebastienros:
The issue comes from the fact that the tick counts are base on `Environment.TickCount` but displayed using ` Timespan` instance. The tick counts don't represent the same thing for both, with a difference of about 10000. The solution is to use `DateTime.Now.Ticks` instead on `Environment.Ticks` as you can't assume that the system timer will always return milliseconds (which seems to be the case on my system).

Here is some code showing the difference:
```c#
using System;
					
public class Program
{
	public static void Main()
	{
		var startEnvironmentTicks = Environment.TickCount;
		var startDateTimeTicks = DateTime.Now.Ticks;
		
		System.Threading.Thread.Sleep(50);

		Console.WriteLine("Environment: {0}", Environment.TickCount - startEnvironmentTicks);
		Console.WriteLine("DateTime: {0}", DateTime.Now.Ticks - startDateTimeTicks);
	}
}
```

Which results in:
```
Environment: 79
DateTime: 781292
```

It's also mentioned in this help page under the remarks section: https://msdn.microsoft.com/en-us/library/system.environment.tickcount(v=vs.110).aspx

JunTaoLuo:
We will need to find a substitute for timing requests. `DateTime.Now` is too slow since it's a system call.

sebastienros:
This gist shows the perf differences between different techniques:
https://gist.github.com/sebastienros/40981268e057eb9235f3

For 1 million calls

`Environment.TickCount`: 15ms
`DateTime.UtcNow.Ticks`: 41ms
`DateTime.Now.Ticks`: 135ms

Not sure the way I profiled is right but I assume it should help deciding.

tmds:
The constructor used here: accepts 100-ns ticks:
[...]

PR #485, opened 2015-11-17
Fix Logged elapsed time

tmds:
Fixes https://github.com/aspnet/Hosting/issues/478

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
@@ -10,6 +10,8 @@ namespace Microsoft.AspNet.Hosting.Internal
 {
     internal static class HostingLoggerExtensions
     {
+        private const long TicksPerMillisecond = 10000;
+		
         public static IDisposable RequestScope(this ILogger logger, HttpContext httpContext)
         {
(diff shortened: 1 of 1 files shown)


Tratcher:
@sebastienros

JunTaoLuo:
Verified that `Environment.TickCount` is always in milliseconds on Windows/Linux/Mac across clr/coreclr/mono. :shipit: from me. Note that this will mean that our logging of elapsed time will remain at the millisecond resolution. @sebastienros

sebastienros:
It more than the millisecond, the documentation states that it's in the order of 10ms.
But :shipit:

JunTaoLuo:
actually why are we not using `TimeSpan.FromMilliseconds`?

sebastienros:
I think you are confused, or you confused me.

JunTaoLuo:
Actually using `new Timespan(<long>)` 10x times faster than `TimeSpan.FromMilliseconds(<double>)` so the current implementation is good.

JunTaoLuo:
Going to close and reopen to trigger the cla bot

JunTaoLuo:
merged to dev

benaadams:
@JunTaoLuo added PR for sub-millisecond timings: "Higher accuracy in timings" #543

PR #487, opened 2015-11-18
Enable detailed startup errors by default

Tratcher:
@victorhurdugaci This should make music store test failures much easier to debug with just CI logs.


victorhurdugaci:
:shipit: after the build passes

PR #492, opened 2015-11-20
Add support to stop application through other means than Ctrl+C.

easyt:
The current implementation only allows to stopping an application through Ctrl+C. This works only for applications running in the console.

In order to support stopping an application running in a Windows service (through the use of WebApplication), this change returns a function from WebApplication.Run which triggers the application stop (which can be called from ServiceBase.Stop).

There is a sample application that uses this change here: https://github.com/taskmatics/aspnet-windows-service/tree/webApplication-test

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingEngine.cs
@@ -96,8 +96,6 @@ namespace Microsoft.AspNet.Hosting.Internal
             Server.Start(
                 async httpContext =>
                 {
-                    httpContext.ApplicationServices = _applicationServices;
-
                     if (diagnosticSource.IsEnabled("Microsoft.AspNet.Hosting.BeginRequest"))
                     {
                         diagnosticSource.Write("Microsoft.AspNet.Hosting.BeginRequest", new { httpContext = httpContext });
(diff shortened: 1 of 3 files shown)


Tratcher:
:-1: I don't think enough of the contents of Run apply to scenarios like services. You should just take the contents you need from Run for your service bootstrapping.

davidfowl:
This isn't correct as @Tratcher eludes to :smile: . You should be using the WebHostBuilder API in your service host. BTW Do you want to contribute the service host to this repository?

easyt:
I am using the WebHostBuilder in the rc1 tag here: https://github.com/taskmatics/aspnet-windows-service/blob/1.0.0-rc1-final/src/AspNetWindowsService/Program.cs#L40

It would just be easier for people if they could use WebApplication. It's just missing a way to stop the application.

@davidfowl: What do you mean by the service host? What would that look like?

Take a look at the code in the repo from the link above. The code is all in Program. It is short. LMK.

davidfowl:
> It would just be easier for people if they could use WebApplication. It's just missing a way to stop the application.

The intent is to make Run blocking, that's also not the only thing missing. There's tons of other stuff that could be exposed but we chose not to because we intended this API for a very narrow scenario. WebHostBuilder is just as usable and makes more sense to use because the Ctrl + C feature doesn't make any sense when being hosted in `svchost.exe`. 

> @davidfowl: What do you mean by the service host? What would that look like?

https://github.com/taskmatics/aspnet-windows-service/blob/webApplication-test/src/AspNetWindowsService/Program.cs#L9

It would be another project in this repository that does this same thing but uses WebHostBuilder.

> https://github.com/taskmatics/aspnet-windows-service/blob/1.0.0-rc1-final/src/AspNetWindowsService/Program.cs#L44

This code is much cleaner.

easyt:
@davidfowl: Fair enough. I will close this pull request. 

I have another option for usage here: https://github.com/taskmatics/aspnet-windows-service/tree/aspnet-hosting-winsvc/src/AspNetWindowsService

New PR: https://github.com/aspnet/Hosting/pull/494

PR #494, opened 2015-11-20
Add support for hosting in a Windows service.

easyt:
@davidfowl: I think this is more inline with your thinking. LMK.

### DIFF
--- a/src/Microsoft.AspNet.Hosting.WindowsServices/WebApplicationService.cs
+++ b/src/Microsoft.AspNet.Hosting.WindowsServices/WebApplicationService.cs
@@ -1,0 +1,80 @@
+﻿using Microsoft.AspNet.Hosting.Internal;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.ServiceProcess;
+
+namespace Microsoft.AspNet.Hosting.WindowsServices
+{
+    /// <summary>
+    ///     Provides an implementation of a Windows service that hosts ASP.NET.
+    /// </summary>
+    public class WebApplicationService : ServiceBase
+    {
+        private const string HostingJsonFile = "hosting.json";
+        private const string EnvironmentVariablesPrefix = "ASPNET_";
+        private const string ConfigFileKey = "config";
+        private IWebApplication _application;
+        private IDisposable _applicationShutdown;
+        private bool _stopRequestedByWindows;
+
+        /// <summary>
+        /// Creates an instance of <c>WebApplicationService</c> which hosts the specified web application.
+        /// </summary>
+        /// <param name="application">The web application to host in the Windows service.</param>
+        public WebApplicationService(IWebApplication application)
+        {
+            _application = application;
+        }
+
+        protected sealed override void OnStart(string[] args)
+        {
+            OnStarting(args);
+
+            _application
+                .Services
+                .GetRequiredService<IApplicationLifetime>()
+                .ApplicationStopped
+                .Register(() =>
+                {
+                    if (!_stopRequestedByWindows)
+                    {
+                        Stop();
+                    }
+                });
+
+            _applicationShutdown = _application.Start();
+
+            OnStarted();
+        }
+
+        protected sealed override void OnStop()
+        {
+            _stopRequestedByWindows = true;
+            OnStopping();
+            _applicationShutdown?.Dispose();
+            OnStopped();
+        }
+
+        /// <summary>
+        /// Executes before ASP.NET starts.
+        /// </summary>
+        /// <param name="args">The command line arguments passed to the service.</param>
+        protected virtual void OnStarting(string[] args) { }
+
+        /// <summary>
+        /// Executes after ASP.NET starts.
+        /// </summary>
+        protected virtual void OnStarted() { }
+
+        /// <summary>
+        /// Executes before ASP.NET shuts down.
+        /// </summary>
+        protected virtual void OnStopping() { }
+
+        /// <summary>
+        /// Executes after ASP.NET shuts down.
+        /// </summary>
+        protected virtual void OnStopped() { }
+    }
+}
--- a/Hosting.sln
+++ b/Hosting.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+VisualStudioVersion = 14.0.24720.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{E0497F39-AFFB-4819-A116-E39E361915AB}"
 EndProject
@@ -29,6 +29,8 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{9C7520A0-F2EB-411C-8BB2-80B39C937217}"
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "SampleStartups", "samples\SampleStartups\SampleStartups.xproj", "{485B6745-7648-400A-A969-F68FCF194E46}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Hosting.WindowsServices", "src\Microsoft.AspNet.Hosting.WindowsServices\Microsoft.AspNet.Hosting.WindowsServices.xproj", "{03148731-EA95-40A2-BAE8-A12315EA1748}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -128,6 +130,18 @@ Global
 		{485B6745-7648-400A-A969-F68FCF194E46}.Release|Mixed Platforms.Build.0 = Release|Any CPU
 		{485B6745-7648-400A-A969-F68FCF194E46}.Release|x86.ActiveCfg = Release|Any CPU
 		{485B6745-7648-400A-A969-F68FCF194E46}.Release|x86.Build.0 = Release|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Debug|x86.Build.0 = Debug|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Release|Any CPU.Build.0 = Release|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Release|Mixed Platforms.Build.0 = Release|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Release|x86.ActiveCfg = Release|Any CPU
+		{03148731-EA95-40A2-BAE8-A12315EA1748}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -141,5 +155,6 @@ Global
 		{FDBBA081-5248-4FC0-9E08-B46BEF3FA438} = {E0497F39-AFFB-4819-A116-E39E361915AB}
 		{3DA89347-6731-4366-80C4-548F24E8607B} = {E0497F39-AFFB-4819-A116-E39E361915AB}
 		{485B6745-7648-400A-A969-F68FCF194E46} = {9C7520A0-F2EB-411C-8BB2-80B39C937217}
+		{03148731-EA95-40A2-BAE8-A12315EA1748} = {E0497F39-AFFB-4819-A116-E39E361915AB}
 	EndGlobalSection
 EndGlobal
--- a/src/Microsoft.AspNet.Hosting.WindowsServices/Microsoft.AspNet.Hosting.WindowsServices.xproj
+++ b/src/Microsoft.AspNet.Hosting.WindowsServices/Microsoft.AspNet.Hosting.WindowsServices.xproj
@@ -1,0 +1,19 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">14.0</VisualStudioVersion>
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+  </PropertyGroup>
+  <Import Project="$(VSToolsPath)\DNX\Microsoft.DNX.Props" Condition="'$(VSToolsPath)' != ''" />
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>03148731-ea95-40a2-bae8-a12315ea1748</ProjectGuid>
+    <RootNamespace>Microsoft.AspNet.Hosting.WindowsServices</RootNamespace>
+    <BaseIntermediateOutputPath Condition="'$(BaseIntermediateOutputPath)'=='' ">..\..\artifacts\obj\$(MSBuildProjectName)</BaseIntermediateOutputPath>
+    <OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\bin\$(MSBuildProjectName)\</OutputPath>
+  </PropertyGroup>
+  <PropertyGroup>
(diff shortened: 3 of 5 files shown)


Tratcher:
Better, this is how a service should be written. What does it look like to consume it? Can you add a sample project that uses it?

Tratcher:
Oh wait, this repo has no samples because it can't depend on servers.... Maybe push a sample PR to Entropy?

easyt:
Take a look here: https://github.com/taskmatics/aspnet-windows-service/tree/aspnet-hosting-winsvc

Note: There is a batch file (install.cmd) that uses the 'svc' command to run the application using this new service host. In VS you use the 'web' command.

src/Microsoft.AspNet.Hosting.WindowsService/project.json
```diff
@@ -0,0 +1,29 @@
...
+    "dependencies": {
+        "Microsoft.AspNet.Hosting": "1.0.0-*"
+    },
+
+    "commands": {
+        "Microsoft.AspNet.Hosting.WindowsService": "Microsoft.AspNet.Hosting.WindowsService"
+    },
+
+    "frameworks": {
+        "dnx451": {
```
Eilon: If you have `net451`, do you even need `dnx451`?

easyt:
I thought it was necessary to specify `dnx451` if you choose to use it as the entry point of an application.

The 'commands' can for sure be removed. The project is intended to be used as the entry point of an application, similar to Microsoft.AspNet.Hosting. Although, a user can inherit from the Program class and override OnStart and OnStop if they want.

Eilon:
@easyt shouldn't need it.

src/Microsoft.AspNet.Hosting.WindowsService/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+        protected override void OnStop()
+        {
+            _application?.Dispose();
+        }
+
+        protected IConfigurationBuilder CreateHostingConfigurationBuilder(string configJsonPath, string[] args)
+        {
+            // We are adding all environment variables first and then adding the ASPNET_ ones
+            // with the prefix removed to unify with the command line and config file formats
+            return new ConfigurationBuilder()
```
davidfowl: We need a one line helper for this.

src/Microsoft.AspNet.Hosting.WindowsService/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+                .AddJsonFile(configJsonPath, optional: true)
+                .AddEnvironmentVariables()
+                .AddEnvironmentVariables(prefix: EnvironmentVariablesPrefix)
```
davidfowl: Does this work when running in a service?

src/Microsoft.AspNet.Hosting.WindowsService/Program.cs
```diff
@@ -0,0 +1,48 @@
+﻿using Microsoft.AspNet.Hosting.Internal;
+using Microsoft.Extensions.Configuration;
+using System.ServiceProcess;
+
+namespace Microsoft.AspNet.Hosting.WindowsService
+{
+    public class Program : ServiceBase
+    {
```
davidfowl: Not sure this should be called program. It's somethings users should derive from in their apps right?

src/Microsoft.AspNet.Hosting.WindowsService/Program.cs
easyt: Yes. The "binPath=" CLI option will have it's options passed in. So when you register the service the "binPath=" parameter can take arguments after the path to the executable/batch file.

src/Microsoft.AspNet.Hosting.WindowsService/Program.cs
easyt: The user can use this package as the entry point to his app, or he can derive from it. I see what you are saying here. I can pull the meat out the same way it was pulled out of Hosting.Program and placed into Hosting.WebApplication, but what should that class be called? WindowServiceWebApplication? It's still a web app.

easyt:
It would be easier if I could explain the intended usage over a quick screen-share, and you can give me guidelines to follow. It would save a lot of back and forth. Would that be possible?

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,80 @@
...
+            var tempConfigBuilder = new ConfigurationBuilder().AddCommandLine(args);
+            var tempConfig = tempConfigBuilder.Build();
+            var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
+            var config = CreateConfiguration(configFilePath, args);
+            var hostBuilder = new WebHostBuilder(config, captureStartupErrors: true);
+            var host = hostBuilder.Build();
+            _application = host.Start();
+            _application.Services.GetRequiredService<IApplicationLifetime>().ApplicationStopped.Register(() =>
+            {
+                if (!_stopRequestedByWindows)
```
davidfowl: *always* add braces.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            // Allow the location of the json file to be specified via a --config command line arg
+            var tempConfigBuilder = new ConfigurationBuilder().AddCommandLine(args);
+            var tempConfig = tempConfigBuilder.Build();
+            var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
+            var config = CreateConfiguration(configFilePath, args);
+            var hostBuilder = new WebHostBuilder(config, captureStartupErrors: true);
+            var host = hostBuilder.Build();
+            _application = host.Start();
+            _application.Services.GetRequiredService<IApplicationLifetime>().ApplicationStopped.Register(() =>
```
davidfowl: Assign the IApplicationLifetime to a variable this line is too long.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            // Allow the location of the json file to be specified via a --config command line arg
+            var tempConfigBuilder = new ConfigurationBuilder().AddCommandLine(args);
+            var tempConfig = tempConfigBuilder.Build();
+            var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
+            var config = CreateConfiguration(configFilePath, args);
+            var hostBuilder = new WebHostBuilder(config, captureStartupErrors: true);
+            var host = hostBuilder.Build();
+            _application = host.Start();
+            _application
```
davidfowl: Not what i meant but at least it looks less crazy with the long line :smile:

Tratcher:
@davidfowl I think this is ready, no?

muratg:
@davidfowl Do you think we should take this as part of the product or a sample?

Either way, I think we'll need tests with it as well. :)

davidfowl:
Yep this is good.

muratg:
@easyt Thanks for the contribution.  Could you add some tests as well?

easyt:
The only way to test this fully is to perform an end-to-end test by installing a test service in Windows (using sc.exe) and hitting an endpoint. I can do this without a problem. However, the test runner will need to run with elevated permissions in order to call sc.exe. The test will also take a few seconds to complete since we would need to poll an endpoint up until a timeout period.

Thoughts?

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+        private bool _stopRequestedByWindows;
+
+        public static void Run()
+        {
+            Run(new WindowsServiceWebApplication());
+        }
+
+        public static void Run(WindowsServiceWebApplication app)
+        {
+            Run(app);
```
Tratcher: base.Run?

Tratcher:
@davidfowl @muratg Maybe this belongs in https://github.com/aspnet-contrib. It's a fully independent package for a specialized scenario. It also probably won't ever work on CoreCLR.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
easyt: ServiceBase.Run will throw if it is not running within the context of a Windows service. It will fail if you just run it from the command line. You have to install it using sc.exe.

easyt:
It won't work on CoreCLR. It requires the full .NET framework.

I personally, would like this feature to be included, and many others have requested this feature after I posted a blog about how it's possible to run ASP.NET in a Windows service.

This makes it so the user doesn't have to code any of the hosting layer on their own.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
Tratcher: Is this method recursive or is it calling the ServiceBase.Run?

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+        private const string ConfigFileKey = "config";
+        private IApplication _application;
+        private bool _stopRequestedByWindows;
+
+        public static void Run()
+        {
+            Run(new WindowsServiceWebApplication());
+        }
+
+        public static void Run(WindowsServiceWebApplication app)
```
DamianEdwards: What is this overload for?

DamianEdwards: Is there any reason for this to be public? When I would actually call this overload from my app?

src/Microsoft.AspNet.Hosting.WindowsService/Program.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Microsoft.AspNet.Hosting.WindowsService
+{
+    public class Program
+    {
+        public static void Main(string[] args)
```
DamianEdwards: I think part is obsolete. The app will require its own `Program.Main` in RC2.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+        protected sealed override void OnStart(string[] args)
+        {
+            OnStarting(args);
+
+            // Allow the location of the json file to be specified via a --config command line arg
+            var tempConfigBuilder = new ConfigurationBuilder().AddCommandLine(args);
+            var tempConfig = tempConfigBuilder.Build();
+            var configFilePath = tempConfig[ConfigFileKey] ?? HostingJsonFile;
+            var config = CreateConfiguration(configFilePath, args);
+            var hostBuilder = new WebHostBuilder(config, captureStartupErrors: true);
```
DamianEdwards: I think `captureStartupErrors` should be configurable, and **off** by default. I'd rather my Windows service crash on start and have the exception logged via the usual failed service mechanism than report that it started and have the error only be available by hitting the endpoint and getting HTML. I believe if we just throw the exception here, `ServiceBase` will do the right thing and the log the exception as part of the failed to start message in the Windows Event Log.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
+﻿using Microsoft.AspNet.Hosting.Internal;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using System.ServiceProcess;
+
+namespace Microsoft.AspNet.Hosting.WindowsService
+{
+    public class WindowsServiceWebApplication : ServiceBase
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        private const string HostingJsonFile = "hosting.json";
+        private const string EnvironmentVariablesPrefix = "ASPNET_";
+        private const string ConfigFileKey = "config";
+        private IApplication _application;
+        private bool _stopRequestedByWindows;
+
+        public static void Run()
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            Run(new WindowsServiceWebApplication());
+        }
+
+        public static void Run(WindowsServiceWebApplication app)
+        {
+            ServiceBase.Run(app);
+        }
+
+        protected sealed override void OnStart(string[] args)
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+                    if (!_stopRequestedByWindows)
+                    {
+                        Stop();
+                    }
+                });
+
+            OnStarted();
+        }
+
+        protected sealed override void OnStop()
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+                .AddEnvironmentVariables()
+                .AddEnvironmentVariables(prefix: EnvironmentVariablesPrefix)
+                .AddCommandLine(args);
+
+            AddConfiguration(configBuilder);
+
+            return configBuilder.Build();
+        }
+
+        protected virtual void OnStarting(string[] args) { }
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        protected virtual void OnStarted() { }
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        protected virtual void OnStopping(bool stopRequestedByWindows) { }
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        protected virtual void OnStopped() { }
```
DamianEdwards: Please add XML doc comments.

src/Microsoft.AspNet.Hosting.WindowsService/WindowsServiceWebApplication.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        protected virtual void AddConfiguration(IConfigurationBuilder configBuilder) { }
```
DamianEdwards: Please add XML doc comments.

Tratcher:
(If you couldn't tell) @DamianEdwards loved this and wants it merged as soon as it has some sort of readme or doc comments.

Tratcher:
@JunTaoLuo Squash and merge this. Then update it as part of your other PR.

easyt:
I just noticed PR #521 from @davidfowl. I just pulled his changes and I have altered my code to make use of his changes. It actually simplifies things a bit. I will update this PR once I update the xml docs.

src/Microsoft.AspNet.Hosting.WindowsServices/WebApplicationService.cs
```diff
@@ -0,0 +1,184 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using System;
+using System.ServiceProcess;
+
+namespace Microsoft.AspNet.Hosting.WindowsServices
+{
+    /// <summary>
+    ///     Provides an implementation of a Windows service that hosts ASP.NET.
+    /// </summary>
+    /// <remarks>
```
davidfowl: I'd remove this as it'll get stale quickly. Leave this for docs on docs.asp.net

src/Microsoft.AspNet.Hosting.WindowsServices/WebApplicationService.cs
```diff
@@ -0,0 +1,184 @@
...
+        public WebApplicationService(IWebApplication application)
+        {
+            _application = application;
+        }
+
+        protected sealed override void OnStart(string[] args)
+        {
+            OnStarting(args);
+
+            _applicationShutdown = _application.Start();
```
davidfowl: Start after registering the application shutdown

davidfowl:
@easyt Can you rebase on dev? The new API is in.

/cc @Tratcher and @JunTaoLuo make sure this goes in ASAP.

src/Microsoft.AspNet.Hosting.WindowsServices/WebApplicationService.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.ServiceProcess;
+
+namespace Microsoft.AspNet.Hosting.WindowsServices
+{
+    /// <summary>
+    ///     Provides an implementation of a Windows service that hosts ASP.NET.
+    /// </summary>
+    public class WebApplicationService : ServiceBase
+    {
+        private const string HostingJsonFile = "hosting.json";
```
Tratcher: It looks like you don't need any of these constants now.

Tratcher:
:shipit: @JunTaoLuo

JunTaoLuo:
Merged in https://github.com/aspnet/Hosting/commit/83c8816eb79ca42158fe8fc7684f5013c9d529c8. Thanks for the contribution @easyt!

muratg:
Thanks @easyt!

easyt:
You guys are very welcome :) I'm glad to be helping out!

Issue #481, opened 2015-11-16
[test] The server deployer sets wrong PATH


victorhurdugaci:
Here's some repro code that you can use with music store

            var logF = new LoggerFactory();
            var logger = logF.CreateLogger("Test1");

            var deploymentParameters = new DeploymentParameters(
                  @"C:\Users\victorhu\dev\musicstore\src\MusicStore",
                  ServerType.WebListener,
                  RuntimeFlavor.CoreClr,
                  RuntimeArchitecture.x86)
            {
                ApplicationBaseUriHint = "http://localhost:5004",
                EnvironmentName = "SocialTesting",
                PublishWithNoSource = false,
                
                UserAdditionalCleanup = parameters =>
                {
                }
            };

            deploymentParameters.EnvironmentVariables
                    .Add(new KeyValuePair<string, string>(
                        "SQLAZURECONNSTR_DefaultConnection",
                        string.Format("Server=(localdb)\\MSSQLLocalDB;Database={0};Trusted_Connection=True;MultipleActiveResultSets=true",
                        "asgasgasgas1")));

            deploymentParameters.EnvironmentVariables.Add(
                new KeyValuePair<string, string>("ASPNET_ENV", "Development"));

            var deployer = ApplicationDeployerFactory.Create(deploymentParameters, logger);

            var result = deployer.Deploy();

- Run it from a 32 bit dnx.
- Notice it starts a 32 bit server.
[...]

PR #495, opened 2015-11-20
Add the DNX runtime to the path when lanching test processes.

Tratcher:
#481 @victorhurdugaci 
There appear to be DNX native dll search path issues on win7. When there's a different dnx on the path (e.g. the one we're using to run the parent test process), we can't load SNI. Add the correct DNX to the path.

### DIFF
--- a/src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
@@ -2,8 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -137,6 +137,10 @@ protected string PopulateChosenRuntimeInformation()
 
             Logger.LogInformation($"Chosen runtime path is {ChosenRuntimePath}");
 
+            // Work around win7 search path issues.
+            var newPath = ChosenRuntimePath + ";" + Environment.GetEnvironmentVariable("PATH");
```
victorhurdugaci: Use PathSeparator? https://msdn.microsoft.com/en-us/library/system.io.path.pathseparator(v=vs.110).aspx

victorhurdugaci:
Looks good :shipit:

PR #505, opened 2015-11-30
Add tests for map path without wwwroot, and typos

pakrym:
@victorhurdugaci @davidfowl


davidfowl:
Add another test that makes sure there's a 404 when the webroot isn't set.

Tratcher:
You'd need to put that test in StaticFiles.

pakrym:
https://github.com/aspnet/StaticFiles/pull/87

davidfowl:
:shipit:

Issue #501, opened 2015-11-24
Remove some platform abstractions from WebHostBuilder bootup


davidfowl:
As part of the move to allow calling AddPlatformServices, we should remove these services from hosting.

PR #509, opened 2015-12-06
Remove non-essential platform services

davidfowl:
- Remove services that can be registered by the application itself.
- These services use to come from the DNX but now they are stand alone
implementations that can be registered by applications if they choose.

MVC and glimpse will need to be updated to manually add the library manager

#501

/cc @pakrym @avanderhoorn


Tratcher:
:shipit:

Issue #497, opened 2015-11-22
TestHost.Create() has incompatible overloads


joaroyen:
rc1 added a set of "Func<IServiceCollection, IServiceProvider> configureServices" overloads to TestHost.Create() that is not compatible with the existing "Action<IServiceCollection> configureServices" overloads:
```c#
public static TestServer Create(Action<IApplicationBuilder> configureApp, Action<IServiceCollection> configureServices)
public static TestServer Create(Action<IApplicationBuilder> configureApp, Func<IServiceCollection, IServiceProvider> configureServices)
```
After upgrading to rc1 I get this error from the compiler:
Error	CS0407	'void Startup.ConfigureServices(IServiceCollection)' has the wrong return type	MediaStreams.Test.

I've currently implemented a workaround by changing signature from void to System.IServiceProvider and returning services.BuildServiceProvider() at the end of the method.

muratg:
Ugh.  Perhaps we should rename the second overload?

@davidfowl

Radek44:
Quick workaround for now is to cast explicitely: 
```
var startup = new Startup(hostingEnvironment, applicationEnvironment);
            var server = TestServer.Create(startup.Configure, (Action<IServiceCollection>)startup.ConfigureServices);
```

Issue #333, opened 2015-08-16
Is there a way to parameterize and pass an instance of Startup rather than just the type?


Tratcher:
_From @humblelistener on August 14, 2015 5:8_

The intention is to avoid creating a separate TestStartUp class rather - rather inject an instance of Configuration

Sample start up below,

```
public class Startup
{
  public IConfiguration Configuration {get; set;}

  public void Startup()
  {
    Configuration = new ConfigurationBuilder()
              .AddEnvironmentVariables()
              .Build();
   }

  public void ConfigureServices(IServiceCollection services)
  {
     //uses Configuration object to set up everything
  }
}
```

In the above implementation, if only 

`` TestServer.CreateBuilder().UseStartup(instanceOfStartup)  ``
is permitted

I can do this,

```
Startup instanceOfStartup= new Startup();
instanceOfStartup.Configuration = myOwnconfigurationWithKeysRequiredForTest;
//create a test server with 
TestServer.CreateBuilder().UseStartup(instanceOfStartup) 
```

Advantage: I dont have to maintain multiple startup class.

Does it make sense?

_Copied from original issue: aspnet/Home#831_

Tratcher:
How about the `UseStartup(Action<IApplicationBuilder>)` overload? So your sample above becomes: `TestServer.CreateBuilder().UseStartup(instanceOfStartup.Configure)`

humblelistener:
These overloads are limited - It wont work because

1. Configure method can be using just IApplicationBuilder or taking ILoggerFactory and so on as parameters

2. If I then have a configure services method - it complicates further because the overload takes a different signature of configureservices compared to expected one defined for configure

Instead of trying to draw two parallels - using StartUp class and StartUpMethods and having supported UseStartUp extension method, IMHO - it will be simple and easy to just start supporting taking an instance of StartUp thats got environment specific configuration state in it.

humblelistener:
and this is what I have to do to keep it working - right now

```
var startup = new Startup();

//update configuration
startup.Configuration["XYZ_ENV_VAR"] = testEnvValue;

var Action<IApplicationBuilder> configureMethod = (app) => {
  app.UseMvc();
};

var configureServicesStartupMethodInfo = typeof(Startup).GetMethod("ConfigureServices");

var configureServicesMethod = new ConfigureServicesBuilder(configureServicesStartupMethodInfo).Build(startup);

[...]

PR #521, opened 2015-12-12
API changes to hosting and TestServer

davidfowl:
Please take a look

/cc @JunTaoLuo @Tratcher @DamianEdwards @lodejard @HaoK @rynowak


src/Microsoft.AspNet.Hosting/IWebApplication.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// The <see cref="IServiceProvider"/> for the application.
+        /// </summary>
+        IServiceProvider Services { get; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        IApplicationLifetime Lifetime { get; }
```
Tratcher: IApplicationLifetime, IHostingEnvironment, ILoggerFactory seem pretty weird to require here. They're all in services.

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
```diff
@@ -83,11 +80,43 @@ public IServiceProvider ApplicationServices
...
+        {
+            get
+            {
+                return Services.GetRequiredService<ILoggerFactory>();
+            }
+        }
+
+        public virtual IDisposable Start()
         {
             var application = BuildApplication();
```
Tratcher: Wasn't BuildApplication going to move to the constructor? Otherwise ServerFeatures isn't available before Start.

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
```diff
@@ -98,14 +127,14 @@ public virtual IApplication Start()
...
 
-            return new Application(ApplicationServices, Server.Features, new Disposable(() =>
-            {
-                logger.Shutdown();
-                _applicationLifetime.StopApplication();
-                Server.Dispose();
-                _applicationLifetime.NotifyStopped();
-                (_applicationServices as IDisposable)?.Dispose();
-            }));
+            return new Disposable(() =>
```
Tratcher: We said not to return just an IDisposable because we would want to add stuff. If you're just going to do IDisposable then make IWebApplication and return void here.

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNet.Hosting.Internal
+{
+    public class WebHostOptions
+    {
+        private const string OldEnvironmentKey = "ENV";
```
Tratcher: Can remove this yet?

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNet.Hosting.Internal
+{
+    public class WebHostOptions
```
Tratcher: WebApplicationOptions?

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
```diff
@@ -32,111 +33,33 @@ public class WebHostBuilder
...
-            {
-                throw new ArgumentNullException(nameof(config));
-            }
-
             _hostingEnvironment = new HostingEnvironment();
             _loggerFactory = new LoggerFactory();
-            _config = config;
-            _options = new WebHostOptions(config);
-            _captureStartupErrors = captureStartupErrors;
+            _config = WebApplicationConfiguration.GetDefault();
```
Tratcher: Let _config start as null and then lazy init it in Build. Otherwise you will usually build config twice since UseConfiguration will usually be called.

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        private const string OldEnvironmentKey = "ENV";
+
+        public WebHostOptions()
```
JunTaoLuo: What's this for? Default constructors are automatically provided.

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            Application = configuration[WebApplicationConfiguration.ApplicationKey];
+            DetailedErrors = ParseBool(configuration, WebApplicationConfiguration.DetailedErrorsKey);
+            CaptureStartupErrors = ParseBool(configuration, WebApplicationConfiguration.CaptureStartupErrorsKey);
+            Environment = configuration[WebApplicationConfiguration.EnvironmentKey] ?? configuration[OldEnvironmentKey];
+            Server = configuration[WebApplicationConfiguration.ServerKey];
+            WebRoot = configuration[WebApplicationConfiguration.WebRootKey];
+        }
+
+        public string Application { get; set; }
```
JunTaoLuo: Don't need set if the only place you set it is the constructor.

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
```diff
@@ -228,7 +167,7 @@ public WebHostBuilder UseStartup(Action<IApplicationBuilder> configureApp, Func<
             return this;
         }
 
-        public WebHostBuilder UseStartup(Action<IApplicationBuilder> configureApp, Action<IServiceCollection> configureServices)
+        public WebApplicationBuilder UseStartup(Action<IApplicationBuilder> configureApp, Action<IServiceCollection> configureServices)
```
Tratcher: How about https://github.com/aspnet/Hosting/issues/333? UseStartup(object instance)?

Tratcher: Or do we just make them use this overload for that scenario?

Tratcher: Note we still need to get rid of the conflicting overload here between `Action<IServiceCollection>` and `Func<IServiceCollection, IServiceProvider>`

Tratcher: https://github.com/aspnet/Hosting/issues/497

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
```diff
@@ -246,5 +185,78 @@ public WebHostBuilder UseStartup(Action<IApplicationBuilder> configureApp, Actio
...
+            {
+                _hostingEnvironment.EnvironmentName = _environmentName;
+            }
+
+            var application = new WebApplication(hostingServices, startupLoader, _options, _config);
+
+            // Only one of these should be set, but they are used in priority
+            application.Server = _server;
+            application.ServerFactory = _serverFactory;
+            application.ServerFactoryLocation = _options.Server ?? _serverFactoryLocation;
```
Tratcher: You already passed _options into WebApplication, it's weird to also use it to set properties.

src/Microsoft.AspNet.Hosting/WebApplicationConfiguration.cs
```diff
@@ -0,0 +1,40 @@
...
+        public static readonly string ServerKey = "server";
+        public static readonly string WebRootKey = "webroot";
+        public static readonly string CaptureStartupErrorsKey = "captureStartupErrors";
+
+        public static readonly string HostingJsonFile = "hosting.json";
+        public static readonly string EnvironmentVariablesPrefix = "ASPNET_";
+
+        public static IConfiguration GetDefault()
+        {
+            return GetDefault(args: null);
```
Tratcher: Environment.GetCommandLineArgs?

src/Microsoft.AspNet.Hosting/IWebApplication.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        IFeatureCollection ServerFeatures { get; }
+
+        /// <summary>
+        /// The <see cref="IServiceProvider"/> for the application.
+        /// </summary>
+        IServiceProvider Services { get; }
+
+        /// <summary>
+        ///
```
JunTaoLuo: Empty summary?

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
```diff
@@ -18,12 +18,13 @@
 
 namespace Microsoft.AspNet.Hosting
 {
-    public class WebHostBuilder
+    public class WebApplicationBuilder
     {
         private readonly IHostingEnvironment _hostingEnvironment;
+
```
JunTaoLuo: nit: spacing looks odd here

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
```diff
@@ -180,7 +113,13 @@ public WebHostBuilder UseServerFactory(IServerFactory factory)
             return this;
         }
 
-        public WebHostBuilder UseStartup(string startupAssemblyName)
+        public WebApplicationBuilder UseLoggerFactory(Action<ILoggerFactory> loggerFactory)
```
JunTaoLuo: Sounds more like `ConfigureLoggerFactory` here? Since the user specify what to do with the given `_loggerFactory` instead of providing one.

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
JunTaoLuo: I don't think `set` is needed if the only place you are setting this property is the constructor.

test/Microsoft.AspNet.TestHost.Tests/TestClientTests.cs
```diff
@@ -33,7 +34,7 @@ public TestClientTests()
             var expected = "GET Response";
             RequestDelegate appDelegate = ctx =>
                 ctx.Response.WriteAsync(expected);
-            var server = TestServer.Create(app => app.Run(appDelegate));
+            var server = new WebApplicationBuilder().UseStartup(app => app.Run(appDelegate)).BuildTestServer();
```
Tratcher: This is going to get pretty noisy.  How about:
```
TestServer.Create(builder => builder.UseStartup(app => app.Run(appDelegate)));
```

Tratcher:
@davidfowl are you going to finish this or is @JunTaoLuo supposed to take over?

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
davidfowl: Didn't do that part yet

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
davidfowl: I don't like that. Disposable seems fine. I'm not sure why you;d need anything else.

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
davidfowl: Maybe but not in this PR.

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
davidfowl: Sounds good.

src/Microsoft.AspNet.Hosting/WebApplicationExtensions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Microsoft.AspNet.Hosting
+{
+    public static class WebApplicationExtensions
+    {
+        /// <summary>
+        /// Retruns the server addresses the web application is going to listen on.
+        /// </summary>
+        /// <param name="application"></param>
+        /// <returns></returns>
```
JunTaoLuo: empty return doc

src/Microsoft.AspNet.Hosting/IWebApplication.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// <summary>
+    /// Represents a configured web application
+    /// </summary>
+    public interface IWebApplication
+    {
+        /// <summary>
+        /// The <see cref="IFeatureCollection"/> exposed by the configured server.
+        /// </summary>
+        IFeatureCollection ServerFeatures { get; }
```
HaoK: Is Server necessary as a prefix, or is it sufficiently implied, i.e. the interface isn't called IServerFeatureCollection...

Tratcher: It's the same name we use on IApplicationBuilder. https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Abstractions/IApplicationBuilder.cs#L24
I think we should leave it.

src/Microsoft.AspNet.Hosting/Samples_WebApplication_delete_me.cs
```diff
@@ -0,0 +1,141 @@
...
+                .UseLoggerFactory(loggerFactory => // Configure logging
+                {
+                    loggerFactory.AddProvider(new MyHostLoggerProvider());
+                })
+                .UseServices(services =>
+                {
+                    // Configure services that the application can see
+                    services.AddSingleton<IMyCustomService, MyCustomService>();
+                })
+                .UseStartup(app =>
```
HaoK: is UseStartup the most appropriate name when this is basically Startup.Configure() inline right?

Would app.Configure() convey that better than UseStartup?

Tratcher: Would you have a separate overload for ConfigureServices?

HaoK: builder.UseServices is exactly that already.  We just had no equivalent for the pipleine

Tratcher: Not quite, isn't that hosting services verses app services?

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
Tratcher: That's what we thought before we added IApplication... @DamianEdwards had the same feedback.

src/Microsoft.AspNet.Hosting/Samples_WebApplication_delete_me.cs
HaoK: So I don't know how things stand now,  UseServices was directly ending up in the ApplicationServices before from my recollection when I wrote it, it wasn't specific to Hosting at all.

There's only one ServiceCollection I thought (Formerly called ApplicationServices => Services now)

src/Microsoft.AspNet.Hosting/Samples_WebApplication_delete_me.cs
```diff
@@ -0,0 +1,141 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNet.Hosting
+{
+    internal class SampleApplications
```
Tratcher: We should move these to actual sample projects rather than delete them.

src/Microsoft.AspNet.Hosting/Samples_WebApplication_delete_me.cs
Tratcher: Fair, it's probably the same result. We'll have to decide what happens if you mix `UseServices` with `UseStartup<TStartup>()` when your Startup also has ConfigureServices.

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
HaoK: Perhaps ConfigureLogging would sound less clunky if not quite as precise?

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
```diff
@@ -202,12 +141,12 @@ public WebHostBuilder UseStartup(Type startupType)
...
         }
 
-        public WebHostBuilder UseStartup<TStartup>() where TStartup : class
+        public WebApplicationBuilder UseStartup<TStartup>() where TStartup : class
         {
             return UseStartup(typeof(TStartup));
         }
 
-        public WebHostBuilder UseStartup(Action<IApplicationBuilder> configureApp)
+        public WebApplicationBuilder UseStartup(Action<IApplicationBuilder> configureApp)
```
HaoK: Maybe split the names where UseStartup is for Type/TStartup, and Configure for the overloads that take an configureAction.  Underneath the cover we would still wire it via a StartupMethods, but there's no real reason that name needs to be in the Builder API does it?  Maybe we just kill the UseStartup overload that takes both configureApp and conifgureServices, since they can do UseServices/Configure as replacements

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
davidfowl: It's not needed.

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
davidfowl: Not sure it was there before

src/Microsoft.AspNet.Hosting/Samples_WebApplication_delete_me.cs
davidfowl: Sounds good

src/Microsoft.AspNet.Hosting/Samples_WebApplication_delete_me.cs
davidfowl: That sounds like a good change. Configure sounds better. I'll change it to Configure and ConfigureServices for the callbacks to see how it looks.

test/Microsoft.AspNet.TestHost.Tests/TestClientTests.cs
davidfowl: Tried it and didn't like it. Too much nesting.

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
davidfowl: Will do

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
davidfowl: _options doesn't have everything.

src/Microsoft.AspNet.Hosting/WebApplicationConfiguration.cs
davidfowl: Not sure if that's an ok default. Should we just assume the command line args for any process should be used as config? Feels wrong.

src/Microsoft.AspNet.Hosting/IWebApplication.cs
davidfowl: #lazy

davidfowl:
@JunTaoLuo will need to  take over. I'd like to get this in this week. I wanted to flesh out the API enough to see if it feels good. I'm going to make the review changes and rebase

src/Microsoft.AspNet.Hosting/Internal/WebHostOptions.cs
davidfowl: testing

src/Microsoft.AspNet.Hosting/IWebApplication.cs
davidfowl: They are services that are created by the Builder but you're right, it might be weird. But they are so commonly used.

Tratcher: You don't have a lot of implementations to base that on. Right now they're only used in Run.

src/Microsoft.AspNet.Hosting/IWebApplication.cs
HaoK: I agree they should be the same, we should probably rename IApplicationBuilder.ApplicationServices to be Services to match this change right?  Personally I think Features for both wouldn't be any less clear, but not a huge deal

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
HaoK: Does it actually hurt to return the actual HostingApplication instance?

HaoK:
API looks much nicer! looks good to me

src/Microsoft.AspNet.Hosting/Program.cs
```diff
@@ -7,7 +7,13 @@ public class Program
     {
         public static void Main(string[] args)
```
Tratcher: Are we removing Program.Main yet?

JunTaoLuo: I did, just forgot to update this PR.

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
```diff
@@ -83,29 +81,43 @@ public IServiceProvider ApplicationServices
...
-            var logger = _applicationServices.GetRequiredService<ILogger<HostingEngine>>();
+        public void Initialize()
+        {
+            if (_application == null)
+            {
+                _application = BuildApplication();
+            }
+        }
+
+        public virtual IDisposable Start()
```
Tratcher: WebApplication can be created directly so Start needs to call Initialize to make sure it's happened.

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
```diff
@@ -32,126 +33,47 @@ public class WebHostBuilder
         private StartupMethods _startup;
         private Type _startupType;
         private string _startupAssemblyName;
-        private readonly bool _captureStartupErrors;
 
         // Only one of these should be set
         private string _serverFactoryLocation;
         private IServerFactory _serverFactory;
         private IServer _server;
+        private string _webRoot;
```
Tratcher: nit: new line, this is not a server field.

src/Microsoft.AspNet.Hosting/WebApplicationBuilder.cs
Tratcher: The order here is reversed, _serverFactoryLocation should take priority over _options like it does for _webRoot.

src/Microsoft.AspNet.Hosting/Internal/WebApplicationOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public string Application { get; set; }
+
+        public bool DetailedErrors { get; set; }
+
+        public bool CaptureStartupErrors { get; set; }
+
+        public string Environment { get; set; }
+
+        public string Server { get; set; }
```
Tratcher: ServerFactory or ServerFactoryLocaiton. See WebApplicationBuilder.UseServerFactory.

test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
```diff
@@ -136,10 +135,10 @@ public void CanDefaultAddresseIfNotConfigured()
...
                 .AddInMemoryCollection(vals);
             var config = builder.Build();
-            var host = CreateBuilder(config).Build();
-            var app = host.Start();
-            Assert.NotNull(host.ApplicationServices.GetRequiredService<IHostingEnvironment>());
-            Assert.Equal("http://localhost:5000", app.ServerFeatures.Get<IServerAddressesFeature>().Addresses.First());
+            var application = CreateBuilder(config).Build();
+            var app = application.Start();
+            Assert.NotNull(application.Services.GetService<IHostingEnvironment>());
+            Assert.Equal("http://localhost:5000", application.ServerFeatures.Get<IServerAddressesFeature>().Addresses.First());
```
Tratcher: application.GetAddresses()

test/Microsoft.AspNet.Hosting.Tests/HostingEngineTests.cs
```diff
@@ -9,7 +9,6 @@
 using System.Threading.Tasks;
```
Tratcher: Rename the class, file, and tests.

test/Microsoft.AspNet.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -24,7 +24,7 @@ public void Build_uses_application_for_startup_assembly_by_default()
         {
```
Tratcher: Rename file, class, methods.

test/Microsoft.AspNet.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -137,22 +137,27 @@ public void UseEnvironmentIsNotOverriden()
...
             {
                 { "server", "Microsoft.AspNet.Hosting.Tests" },
                 { "DetailedErrors", "true" },
+                { "captureStartupErrors", "true" }
             };
             var builder = new ConfigurationBuilder()
                 .AddInMemoryCollection(vals);
             var config = builder.Build();
-            return new WebHostBuilder(config, captureStartupErrors: true);
+            return new WebApplicationBuilder().UseConfiguration(config).UseServer(new TestServer());
```
Tratcher: Why is UseServer needed here? It should pick it up from the config.

test/Microsoft.AspNet.TestHost.Tests/RequestBuilderTests.cs
```diff
@@ -25,7 +26,7 @@ public void AddRequestHeader()
         [Fact]
         public void AddContentHeaders()
         {
-            var server = TestServer.Create(app => { });
+            var server = new WebApplicationBuilder().Configure(app => { }).BuildTestServer();
```
Tratcher: @davidfowl :-1: this has regressed the usability of the most common test scenario.

Tratcher: I still think we should have
`TestServer.Create(host => host.Configure(app => {}));`

test/Microsoft.AspNet.Hosting.Tests/WebHostBuilderTests.cs
JunTaoLuo: This is very strange, there is no ServerFactory in this assembly, and most of the tests override it manually anyways by setting the TestServer explicitly.

davidfowl:
@JunTaoLuo @Tratcher another change we should make. We should disable the automatic detection of startup based on assembly if nothing was specified. The API should just make you specify it and our template should be explicit as well.

JunTaoLuo:
@davidfowl What do you mean by automatic detection? Do you mean ApplicationName from PlatformHandlers, Configuration, StartupAssemblyName or all of them?https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs#L132

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
lodejard: It was `IDisposable Start()` from a long time ago, but the pattern changed. The builder was changed to return a type that implemented IDisposable, and also had a public void Start method, but that object which was returned no longer supports multiple calls to Start so there's no reason for each call to Start to return a new disposable.

davidfowl:
Just application name. The others are explicit

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
JunTaoLuo:
```
        {
            var config = WebApplicationConfiguration.GetDefault(args);

            var builder = new WebApplicationBuilder()
                .UseConfiguration(config)
                .UseStartup<StartupBlockingOnStart>();

            using (var application = builder.Build())
            {
                application.Start();
                Console.ReadLine();
            }
        }
```

This is what usage looks like for the alternative API where we `Start()` returns `void`. Does this look better than the current usage?

lodejard: Consider adding .Start and .Run as extension to the builder. That way you could have

using (new Builder().UseEtc().Start()) {...}

Or

new Builder().UseEtc().Run();

Sent from mobile device
________________________________

JunTaoLuo: This would not cover the cases where we want to configure things like addresses after the application builds but before the application runs. Also, doesn't `Start()` or `Run()` feel a little out of place as an extension to a `Builder`?

lodejard: Seems fine to me if it makes the usage cleaner, though we could change the types to be `WebApplication` and `IWebApplicationInstance` if the problem is that the first class has the suffix `Builder`.

Consider adding 
```cs
        public WebApplicationBuilder UseConfiguration(
            string hostingJsonFile = null,
            string environmentVariablesPrefix = null,
            string[] commandLineArgs = null)
        {
            var configBuilder = new ConfigurationBuilder();

            if (!string.IsNullOrEmpty(hostingJsonFile))
            {
                configBuilder.AddJsonFile(
                    path: hostingJsonFile,
                    optional: true);
            }

            if (!string.IsNullOrEmpty(environmentVariablesPrefix))
            {
                configBuilder.AddEnvironmentVariables(
                    prefix: environmentVariablesPrefix);
            }

            if (commandLineArgs != null)
            {
                configBuilder.AddCommandLine(
                    args: commandLineArgs);
            }

            return UseConfiguration(configBuilder.Build());
        }
```
to enable the syntax
```cs
    public class StartupBlockingOnStart
    {
        // Entry point for the application.
        {
            using (new WebApplicationBuilder()
                .UseConfiguration(commandLineArgs: args)
                .Start())
            {
                Console.ReadLine();
            }
        }
    }
```

src/Microsoft.AspNet.Hosting/WebApplicationConfiguration.cs
lodejard: We shouldn't assume any of the configuration sources are "hot" by default when self-hosting through the builder. See other remark for adding builder.UseConfiguration where you opt-in and control which json file and environment variables (if any) apply to your program.

```cs
        public WebApplicationBuilder UseConfiguration(
            string hostingJsonFile = null,
            string environmentVariablesPrefix = null,
            string[] commandLineArgs = null)
```

src/Microsoft.AspNet.Hosting/Internal/WebApplication.cs
JunTaoLuo: `WebApplication` and `WebApplicationInstance` doesn't resolve the usage case illustrated by `StartupConfigureAddresses` unless we add some sort of `ConfigureWebApplicationInstance` callback. We need to finish this up so I'll open a new issue tracking the discussion here and a few other items once this PR is done.

JunTaoLuo: As for the configuration, we are using the inlined version `.UseConfiguration(WebApplicationConfiguration.GetDefault(args))` in our samples. Since we want to use the settings in the "hosting.json". For the suggestion above, we would want something like `.UseConfiguration(hostingJsonFile: WebApplicationConfiguration.HostingJsonFile, commandLineArgs: args)`. 

@Tratcher did suggest adding an overload for `.UsingConfiguration(args)` which by default will call `GetDefault(args)`. This does make things more concise but at the cost of the addition of settings in hosting.json and environment variables being non-transparent.

lodejard: For addresses we should expose those as a first-class controllable thing, like environment. Also, when you call .Build() there's waaay too much of the processing being done. Creating the server factory, and getting the features, and populating the application builder should all happen within the Start method.

So that would look like
```
            new WebApplicationBuilder()
                .UseConfiguration(commandLineArgs: args)
                .UseEnvironment("Production")
                .UseAddresses("http://localhost:5000", "https://localhost:5001")
                .Run();
```

Heck, at that point we should probably add extension methods overloads for Start and Run which take addresses. (Kind of like node's final method call of .listen takes the port).

```cs
    public class StartupExternallyControlled
    {
        private readonly List<string> _addresses = new List<string>();
        private IWebApplication _application;

        public StartupExternallyControlled()
        {
        }

        public void Start()
        {
            _application = new WebApplicationBuilder()
                .Start(_addresses.ToArray());
        }

        public void Stop()
        {
            _application.Dispose();
        }

        public void AddUrl(string url)
        {
            _addresses.Add(url);
        }
    }
```

```cs
    public class StartupConfigureAddresses
    {
        // Entry point for the application.
        {
            new WebApplicationBuilder()
                .UseConfiguration(
                    hostingJsonFile: "hosting.json", 
                    commandLineArgs: args)
                .UseEnvironment("Production")
                .Run("http://localhost:5000", "https://localhost:5001");
        }
    }
```

lodejard: We should definitely continue the conversation when we're all in the office. In the case where you are using these APIs as a program of any kind which is self-hosting an http endpoint it's quite likely that automatically picking up the "hosting.json" and ASPNET_ env vars is actually the wrong thing to do. Too much detail to go into why that's the case on a github thread.

Also - the problems you're bumping into about how you have to call Build before being able to change Addresses is a sign that ServerFeatures interface is the wrong way to control that. You should be able to provide the addresses on the Builder.

PR #523, opened 2015-12-14
Fixing deadlock hang in test when running in single threaded environment

JunTaoLuo:
Found the issue for #422. The test hangs because we are mixing async code in TestServer and blocking code in the test. This will cause a deadlock scenario but only when there is a single threaded scenario so it was undetected until now.

### DIFF
--- a/src/Microsoft.AspNet.TestHost/ClientHandler.cs
+++ b/src/Microsoft.AspNet.TestHost/ClientHandler.cs
@@ -94,7 +94,7 @@ namespace Microsoft.AspNet.TestHost
                     }
                 });
 
-            return await state.ResponseTask;
+            return await state.ResponseTask.ConfigureAwait(false);
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.TestHost/ClientHandler.cs
```diff
@@ -94,7 +94,7 @@ public ClientHandler(PathString pathBase, IHttpApplication<Context> application)
                     }
                 });
 
-            return await state.ResponseTask;
+            return await state.ResponseTask.ConfigureAwait(false);
```
JunTaoLuo: Though either the test change or the `ConfigureAwait` here will resolve the deadlock, it is suggested to do both here: http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

Tratcher:
:shipit:

PR #535, opened 2015-12-22
Fast path state machine

benaadams:


src/Microsoft.AspNet.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -43,10 +43,14 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
             // All done if request services is set
             if (existingFeature?.RequestServices != null)
```
davidfowl: 99% of the time this code path will never get hit.

benaadams: Depends if is skippable for benchmark I suppose? Stepping through profile traces line by line atm

davidfowl: You can just remove the whole thing in the bemchmark.

benaadams: That's probably an even faster path :+1:

Issue #547, opened 2016-01-04
Add Options services back to hosting


davidfowl:
This will allow middleware to depend on IOptions<T> without having to expose AddServices extension methods.

PR #548, opened 2016-01-05
Add Options to the default set of services

davidfowl:
#547


src/Microsoft.AspNet.Hosting/project.json
```diff
@@ -15,6 +15,7 @@
     "Microsoft.AspNet.Hosting.Server.Abstractions": "1.0.0-*",
     "Microsoft.AspNet.Http": "1.0.0-*",
     "Microsoft.AspNet.Http.Extensions": "1.0.0-*",
+    "Microsoft.Extensions.Options": "1.0.0-*",
```
Tratcher: L < O < P

Issue #530, opened 2015-12-21
Hosting API followup


JunTaoLuo:
Continuing discussion on items left over from #525

1. Configuration
Currently we use recommend using `builder.UseConfiguration(WebApplicationConfiguration.GetDefault(args))`. This is verbose. One suggestion for this is to add a `UseConfiguration(string[] args)` overload for only command line configs. Also, there is some contention on whether environment variables and hosting.json configurations should be enabled by default.

2. WebApplication void Start() vs IDisposable Start()
There are two suggestions for `WebApplication.Start()`. We currently use the IDisposable pattern which lends itself to a convenient pattern `using(app.Start()){ ... }` where we handle shutdown through `Dispose()`. However, since we do not support an application being started more than once, it is natural to return `void`. This leads to less desirable usage patterns however, such as `using(var app = builder.Build()){app.Start(); ... }`.

3. WebApplicationBuilder.Run()/.Start()
Currently we build an application using a builder, configure that application (such as setting addresses), then run or start the application. There is a suggestion on making address configuration more first class on the `WebApplicationBuilder` and directly calling `Start()` or `Run()` on the builder but this will require shuffling of startup since Addresses are currently ServerFeatures that are only available after the WebApplication is built. It's also suggested to possibly add overloads to `Start()` and `Run()` to take in a list of addresses to listen on.

4. Exception messages
Some of the current exception messages need to be updated. For example, the message at [StartupLoader.cs#L46](https://github.com/aspnet/Hosting/blob/0394987271383c5934395fa299f662991143b4fe/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs#L46) can be cryptic. For this exception in particular, it should be determined if it's the responsibility of the `WebApplication` or the `StartupLoader`, which could be third party, to throw exception to the user.

davidfowl:
> Also, there is some contention on whether environment variables and hosting.json configurations should be enabled by default.

Since we have the helper that creates the right `IConfiguration` with the default behavior, this could go either way. It's easy to turn it on or off by calling UseConfiguration.

> builder.UseConfiguration(WebApplicationConfiguration.GetDefault(args))

It is too verbose, I do worry about having `UseConfiguration(args)` because it's not useful outside of configuring command line args nor does it compose with the other `UseConfiguration` overload. When we're about to make decisions on these APIs we should have a few scenarios in mind:

1. Give me the defaults
2. I want the default and I want to override 1 configuration option thing
3. Give me something without defaults

[...]

PR #553, opened 2016-01-07
More hosting API changes

davidfowl:
- Added IWebApplicationBuilder
- Made methods extension methods where possible
- Added UseUrls

#530


src/Microsoft.AspNet.Hosting.Abstractions/IWebApplicationBuilder.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
```
JunTaoLuo: As promised, you're missing copyright.

src/Microsoft.AspNet.Hosting/Internal/IncludedConfigurationProvider.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
JunTaoLuo: And here too

src/Microsoft.AspNet.Hosting/Internal/IncludedConfigurationProvider.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNet.Hosting.Internal
+{
+    // TODO: Remove this once https://github.com/aspnet/Configuration/pull/349 gets merged
```
Tratcher: Merged

PR #554, opened 2016-01-07
Move default config to Hosting.Abstractions

davidfowl:
This will make it possible to get at the config values without depending on Hosting


src/Microsoft.AspNet.Hosting.Abstractions/WebApplicationDefaults.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNet.Hosting
+{
+    public static class WebApplicationDefaults
```
moozzyk: not sure if `Defaults` is a fortunate name here...

moozzyk:
:ship::it:

PR #556, opened 2016-01-07
Build with dotnet

victorhurdugaci:
cc @Tratcher 

I disabled two tests just to see if travis/appveyor pass. Waiting on @davidfowl to add the `publishInclude` (or whatever the name will be) and then I'll reenable them. The tests fail because the `wwwroot` folder is not published.


Tratcher:
looks fine

PR #560, opened 2016-01-11
Update IISDeployer and its base class

troydai:
In addition to these changes, there are a few minor issues @pakrym will help to resolve:
1. the guid in deploy path is not necessary;
2. host.json creation.

/cc @Tratcher @pakrym


Tratcher:
:+1: @pakrym does this look like what you did?
Please test with the ServerTests repo.

Tratcher:
FYI: I think we can avoid writing the hosting.json file. If we really need to pass in those parameters we can do it via command line or environment variable.

troydai:
Ran with ServerTests, no new test failures.

troydai:
Ping @Tratcher @pakrym I’d like to have this piece check in early tomorrow and then I’ll sit through the CI.

Tratcher:
:shipit:

PR #561, opened 2016-01-11
Making application services available from TestServer #519

JunTaoLuo:
#519 This will allow DBContexts added to services by the builder available from the TestServer in unit tests.


Tratcher:
:shipit:

davidfowl:
Might be better to expose the `IWebApplication`.

JunTaoLuo:
@davidfowl what does exposing IWebApplication achieve?

davidfowl:
@JunTaoLuo the fact that we never have to change the TestServer API to expose things that are already exposed by `IWebApplication`.

JunTaoLuo:
:up::date:

davidfowl:
:shipit:

PR #562, opened 2016-01-12
Fixing logger nullref when context created with no feature #532

### LINKED ISSUE #532: Using an active logger (not NullLogger) with TestHost results in a null ref
From https://github.com/aspnet/Mvc/pull/3796#discussion_r48160786. Here's the stack trace:

```C#
System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
   at Microsoft.AspNet.Http.Internal.DefaultHttpRequest.get_Protocol()
   at Microsoft.AspNet.Hosting.Internal.HostingLoggerExtensions.HostingRequestStarting.ToString()
   at Microsoft.AspNet.Hosting.Internal.HostingLoggerExtensions.HostingRequestStarting.<>c.<.cctor>b__7_0(Object state, Exception exception)
   at Microsoft.Extensions.Logging.Console.ConsoleLogger.Log(LogLevel logLevel, Int32 eventId, Object state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.Logger.Log(LogLevel logLevel, Int32 eventId, Object state, Exception exception, Func`3 formatter)
```

JunTaoLuo:
Fixes #532. The issue there was that we were creating `HttpContext` with no `IHttpRequestFeature`. When creating the context, we try to log certain details about the request which null refs. The null logger added by MVC circumvents this since if the logger is not enabled, we don't try to access the information on the request. I will update MVC to remove the `NullLogger.Instance` after this is merged.

### DIFF
--- a/test/Microsoft.AspNet.TestHost.Tests/ClientHandlerTests.cs
+++ b/test/Microsoft.AspNet.TestHost.Tests/ClientHandlerTests.cs
@@ -5,13 +5,19 @@ using System;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Hosting;
+using Microsoft.AspNet.Hosting.Internal;
 using Microsoft.AspNet.Hosting.Server;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Internal;
 using Microsoft.AspNet.Http.Features;
 using Microsoft.AspNet.Testing.xunit;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Xunit;
 using Context = Microsoft.AspNet.Hosting.Internal.HostingApplication.Context;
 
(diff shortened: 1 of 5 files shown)


davidfowl:
Add a test?

JunTaoLuo:
Testing's a little tricky. We need to verify that at the time of context creation, the request is non-empty. We can either do this via a special IHttpApplication that does the check before creating the context or via a functional test where we add a logger that tries to access those values. I'll try and see what's better.

test/Microsoft.AspNet.TestHost.Tests/ClientHandlerTests.cs
```diff
@@ -269,5 +275,61 @@ public Task ProcessRequestAsync(Context context)
...
+                            {
+                                app.Run(context =>
+                                {
+                                    return Task.FromResult(0);
+                                });
+                            });
+            var server = new TestServer(builder);
+
+            var result = await server.CreateClient().GetStringAsync("/");
+            Assert.Contains("Request starting HTTP/1.1 GET http://", logger.Message);
```
JunTaoLuo: This is actually pretty misleading. We are only using the logger to access all the fields of the request to make sure they are set when creating the context. It's meaningless as we set them to different values after the creation.

davidfowl: Since we don't care about the text can we assert that it's non empty?

JunTaoLuo: We don't even need the assert at all. We can regard the test as successful as long as no exception was thrown during context creation. I added the assert as a sanity check that at least the logger is able to extract some reasonable values. 

Asserting non-empty is not very useful since `HostingLogScope` and `HostingRequestFinished` messages will almost always guarantee that it's non-empty. The test already looks strange as it is, having no assert or simply asserting it's not empty seems to be even more confusing.

Though maybe I should add some comments to at least indicate what the test is actually doing...

davidfowl: I prefer that than this random log.

JunTaoLuo:
:up::date:

JunTaoLuo:
ping @davidfowl @Tratcher

Tratcher:
:shipit:

Issue #559, opened 2016-01-11
The error message when there's no Startup class specified sucks


davidfowl:
```
System.ArgumentException: Value cannot be null or empty.
Parameter name: startupAssemblyName
   at Microsoft.AspNet.Hosting.Startup.StartupLoader.FindStartupType(String startupAssemblyName, IList`1 diagnosticMessages)
   at Microsoft.AspNet.Hosting.Internal.WebApplication.EnsureStartup()
   at Microsoft.AspNet.Hosting.Internal.WebApplication.EnsureApplicationServices()
   at Microsoft.AspNet.Hosting.Internal.WebApplication.BuildApplication()
   at Microsoft.AspNet.Hosting.WebApplicationBuilder.Build()
   at ConsoleApp1.Program.Main(String[] args) in c:\dev\scratch\WebApplication10\src\ConsoleApp1\Program.cs:line 13
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.Dnx.Runtime.Common.EntryPointExecutor.Execute(Assembly assembly, String[] args, IServiceProvider serviceProvider)
   at Microsoft.Dnx.ApplicationHost.Program.<>c__DisplayClass3_0.<ExecuteMain>b__0()
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
```

/cc @Tratcher @JunTaoLuo

PR #564, opened 2016-01-12
Elaborating error message when no startup is specified

JunTaoLuo:
#559

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
+++ b/src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
@@ -43,7 +43,10 @@ namespace Microsoft.AspNet.Hosting.Startup
             var environmentName = _hostingEnv.EnvironmentName;
             if (string.IsNullOrEmpty(startupAssemblyName))
             {
-                throw new ArgumentException("Value cannot be null or empty.", nameof(startupAssemblyName));
+                throw new ArgumentException(
+                    string.Format("A startup method, startup type or startup assembly is required. If specifying an assembly, '{0}' cannot be null or empty.", 
+                    nameof(startupAssemblyName)), 
+                    nameof(startupAssemblyName));
             }
 
             var assembly = Assembly.Load(new AssemblyName(startupAssemblyName));


src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -43,7 +43,10 @@ public Type FindStartupType(string startupAssemblyName, IList<string> diagnostic
             var environmentName = _hostingEnv.EnvironmentName;
             if (string.IsNullOrEmpty(startupAssemblyName))
             {
-                throw new ArgumentException("Value cannot be null or empty.", nameof(startupAssemblyName));
+                throw new ArgumentException(
+                    string.Format("A startup method, startup type or startup assembly is required. If specifying an assembly, '{0}' cannot be null or empty.",
```
davidfowl: We don't have any resources?

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
```diff
@@ -43,7 +43,10 @@ public Type FindStartupType(string startupAssemblyName, IList<string> diagnostic
             var environmentName = _hostingEnv.EnvironmentName;
             if (string.IsNullOrEmpty(startupAssemblyName))
             {
-                throw new ArgumentException("Value cannot be null or empty.", nameof(startupAssemblyName));
+                throw new ArgumentException(
+                    string.Format("A startup method, startup type or startup assembly is required. If specifying an assembly, '{0}' cannot be null or empty.", 
+                    nameof(startupAssemblyName)), 
+                    nameof(startupAssemblyName));
```
cesarbs: Why twice?

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
JunTaoLuo: Not in hosting, I assume that will occur when localization strings are added down the road.

src/Microsoft.AspNet.Hosting/Startup/StartupLoader.cs
JunTaoLuo: The first one is for the format string, the second is for the parameter name of the exception. I don't think the second one is that useful but it was there previously so I didn't remove it.

davidfowl:
:shipit:

PR #569, opened 2016-01-13
Use dynamic ports for all deployers

pakrym:
@Tratcher @troydai

### DIFF
--- a/src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
@@ -93,7 +93,7 @@ namespace Microsoft.AspNet.Server.Testing
 
         protected void DnuPublish(string publishRoot = null)
         {
-            DeploymentParameters.PublishedApplicationRootPath = Path.Combine(publishRoot ?? Path.GetTempPath(), Guid.NewGuid().ToString());
+            DeploymentParameters.PublishedApplicationRootPath = publishRoot ?? Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
 
             var noSource = DeploymentParameters.PublishWithNoSource ? "--no-source" : string.Empty;
             var command = DeploymentParameters.Command ?? "web";
--- a/src/Microsoft.AspNet.Server.Testing/Common/FreePortHelper.cs
+++ b/src/Microsoft.AspNet.Server.Testing/Common/FreePortHelper.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net;
+using System.Net.Sockets;
+
+namespace Microsoft.AspNet.Server.Testing.Common
+{
+    public static class FreePortHelper
+    {
+        public static Uri FindFreeUrl(string urlHint)
+        {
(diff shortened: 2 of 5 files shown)


src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -9,6 +9,8 @@
 using Microsoft.AspNet.Testing;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.PlatformAbstractions;
+using System.Net.Sockets;
+using System.Net;
```
troydai: Sort using statements.

src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -93,7 +95,7 @@ protected void PickRuntime()
 
         protected void DnuPublish(string publishRoot = null)
         {
-            DeploymentParameters.PublishedApplicationRootPath = Path.Combine(publishRoot ?? Path.GetTempPath(), Guid.NewGuid().ToString());
+            DeploymentParameters.PublishedApplicationRootPath = publishRoot ?? Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
```
troydai: :+1:

src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -268,5 +270,31 @@ protected static string GetOSPrefix()
                     throw new InvalidOperationException("Unrecognized operating system");
             }
         }
+
+        protected Uri FindFreeUrl(string urlHint)
```
troydai: What about just move it to a helper class instead of a protected method?

src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -268,5 +270,31 @@ protected static string GetOSPrefix()
...
+        {
+            var uriHint = new Uri(urlHint);
+            var builder = new UriBuilder(uriHint)
+            {
+                Port = FindFreePort(uriHint.Port)
+            };
+            return builder.Uri;
+        }
+
+        protected static int FindFreePort(int initialPort = 0)
```
troydai: What about just move it to a helper class instead of a protected method?

And we are encouraged not to use default parameter value. Instead use methods overload.

src/Microsoft.AspNet.Server.Testing/Deployers/ApplicationDeployer.cs
pakrym: We may, but it's not used by anyone except implementations of ApplicationDeployer

troydai: It's nice to keep the size of base class small.

troydai:
:shipit:  once the comments are addressed.

pakrym:
:up: :date:

Tratcher:
Are you going to update the tests in IISIntegration, ServerTests, and MusicStore to remove the hardcoded urls?

pakrym:
I wasn't going to, current implementation would try to start server on requested url, and only if it's unavailable will assign another port, so these test should be fine.

Tratcher:
What happens if the test doesn't specify one? Is there a default?

PR #573, opened 2016-01-14
Add default url fallback to deployers

pakrym:
Sometimes `urlHint` is null
@troydai @Tratcher @victorhurdugaci


troydai:
Oh, mine is definitely better https://github.com/aspnet/Hosting/pull/574.

troydai:
Thanks @pakrym. Covered in https://github.com/aspnet/Hosting/commit/80bc8bbbe1de75ef4ca322504f7c433ab589148b

PR #574, opened 2016-01-14
Fix test break caused by null ApplicationBaseUrlHint

troydai:
In addtion to the fix here are the changes to improve the helper:
1. Rename the helper to better represent its function;
2. Add overload to generate free port without hint;
3. Add overload to generate local test url;

/cc @pakrym @Tratcher 

I'll merge it soon to unblock performance test break.

### DIFF
--- a/src/Microsoft.AspNet.Server.Testing/Common/TestUriHelper.cs
+++ b/src/Microsoft.AspNet.Server.Testing/Common/TestUriHelper.cs
@@ -7,16 +7,29 @@ using System.Net.Sockets;
 
 namespace Microsoft.AspNet.Server.Testing.Common
 {
-    public static class FreePortHelper
+    public static class TestUriHelper
     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Server.Testing/Common/TestUriHelper.cs
```diff
@@ -7,16 +7,29 @@
 
 namespace Microsoft.AspNet.Server.Testing.Common
 {
-    public static class FreePortHelper
+    public static class TestUriHelper
     {
-        public static Uri FindFreeUrl(string urlHint)
+        public static Uri BuildTestUri()
```
troydai: Rename from `FindFreeUrl` since it didn't actually "find a url".

pakrym:
:sad: :sheep: :it:

PR #575, opened 2016-01-15
Added overload of Run that triggers shutdown on a cancellation token

davidfowl:
/cc @Tratcher @JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNet.Hosting/WebApplicationBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/WebApplicationBuilderExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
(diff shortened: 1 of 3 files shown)


JunTaoLuo:
Changes looks good, but I think we can add a test using the cancellation token overload now to make sure cancellation is triggered?

src/Microsoft.AspNet.Hosting/WebApplicationBuilderExtensions.cs
```diff
@@ -101,11 +102,37 @@ public static IWebApplication Start(this IWebApplicationBuilder applicationBuild
         }
 
         /// <summary>
-        /// Runs a web application and block the calling thread until host shutdown.
+        /// Runs a web application and block the calling thread until token is triggered or shutdown is triggered
```
JunTaoLuo: Shouldn't the description be swapped with the other overload?

davidfowl:
:up: :date:

Issue #567, opened 2016-01-13
Consider changing WebApplication* to WebHost*


davidfowl:
We have `IWebApplicationBuilder`, `WebApplicationBuilder`, `IWebApplication`, and `IApplicationBuilder`.

`IApplicationBuilder` looks related to `IWebApplicationBuilder` but they are not. 

Today:

```C#
var app = new WebApplicationBuilder()
                        .UseServer("Microsoft.AspNet.Server.Kestrel")
                        .UseUrls("http://localhost:5001")
                        .UseStartup<Startup>()
                        .Build();

app.Run();
```

Potentially changed to:

```C#
var host = new WebHostBuilder()
                        .UseServer("Microsoft.AspNet.Server.Kestrel")
                        .UseUrls("http://localhost:5001")
                        .UseStartup<Startup>()
                        .Build();

host.Run();
```

`IWebHostBuilder` builds the `IWebHost`

/cc @lodejard @Tratcher @JunTaoLuo @DamianEdwards  Thoughts? I'm not 100% on the naming though and I also don't think this matters that much but think about it a little :smile: 

This *is* the hosting repository and we are *hosting* ASP.NET so it sorta makes sense.

Tratcher:
I do agree there's ambiguity between IWebApplicationBuilder and IApplicationBuilder.

davidfowl:
Shall we go with host instead of application then?

Tratcher:
You're with @DamianEdwards right? Did he have anything to say on the subject?

I think it's fine, but I don't envy @JunTaoLuo having to touch all the repos again.

davidfowl:
Yes, we spoke about this today and concluded that it looks like WebApplication was wrong.

mikes-gh:
Would that mean also changing WebApplicationConfiguration to WebHostConfiguration?

davidfowl:
@mikes-gh yes re: `WebHostConfiguration`

mikes-gh:
@davidfowl 
```
WebHostBuilder
WebHostConfiguration
```
vs
```
WebHostingBuilder
WebHostingConfiguration
```
The later seems slightly more abstracted to me

davidfowl:
I dunno what you mean by more abstracted. I prefer:

I prefer a `WebHostBuilder` that builds a  `WebHost` using `WebHostConfiguration`

mikes-gh:
WebHost is fine.

davidfowl:
[...]

PR #577, opened 2016-01-17
Renaming WebApplication* to WebHost*

JunTaoLuo:
#567 Did I missed anything? @Tratcher @davidfowl If not, I'll make all the updates to other repos tomorrow directly to their dev.


davidfowl:
Looks good. As part of this change we should add 2 extension methods to clean up the Program.Main a bit more:

```C#
public static IWebHostBuilder UseDefaultConfiguration(this IWebHostBuilder builder)
{
    return builder.UseDefaultConfiguration(args: null);
}

public static IWebHostBuilder UseDefaultConfiguration(this IWebHostBuilder builder, string[] args)
{
    return builder.UseConfiguration(WebHostConfiguration.GetDefault(args));
}
```

src/Microsoft.AspNet.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public interface IWebHostBuilder
```
davidfowl: Add doc comments to the interface while you're at it?

src/Microsoft.AspNet.Hosting.WindowsServices/WebHostExtensions.cs
```diff
@@ -6,37 +6,37 @@
 namespace Microsoft.AspNet.Hosting.WindowsServices
 {
     /// <summary>
-    ///     Extensions to <see cref="IWebApplication" for hosting inside a Windows service. />
+    ///     Extensions to <see cref="IWebHost" for hosting inside a Windows service. />
     /// </summary>
-    public static class WebApplicationExtensions
+    public static class WebHostExtensions
```
Tratcher: WebHostServiceExtensions

Tratcher: WebHostServiceRunnerExtensions
Extension classes can't be directly named ClassFooExtensions because you start running into conflicts when everybody does it.

src/Microsoft.AspNet.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -10,102 +10,102 @@
 
 namespace Microsoft.AspNet.Hosting
 {
-    public static class WebApplicationBuilderExtensions
+    public static class WebHostBuilderExtensions
```
Tratcher: WebHostBuilderBasicExtensions?

Tratcher:
Hosting.json -> WebHost.json wasn't a terrible suggestion. @davidfowl?

davidfowl:
> Hosting.json -> WebHost.json wasn't a terrible suggestion. @davidfowl?

Nah, I dont see a reason to change it.

src/Microsoft.AspNet.Hosting.WindowsServices/WebHostExtensions.cs
JunTaoLuo: What do we conflict with? If third party extensions?

JunTaoLuo:
:up::date:

src/Microsoft.AspNet.Hosting.WindowsServices/WebHostExtensions.cs
davidfowl: Our own Extension classes as well. Having the same namespace in different assemblies makes it hard to search for docs.

src/Microsoft.AspNet.Hosting/WebHostBuilderExtensions.cs
JunTaoLuo: CoreExtensions?

davidfowl: We make 2 files. One for extensions on `IWebHost`, one for extensions on `IWebHostBuilder`:

`WebHostBuilderExtensions`
`WebHostExtensions`

The services one can be called:

`WebHostServicesExtensions`

JunTaoLuo: I think @Tratcher wanted WebHostServiceRunnerExtensions, because of potential conflicts?

davidfowl: Then call it `WebHostWindowsServicesExtensions`

src/Microsoft.AspNet.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNet.Hosting.Server;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.AspNet.Hosting
+{
+    public static class WebHostBuilderExtensions
```
JunTaoLuo: @davidfowl, @Tratcher also wanted to change this to something like `WebHostBuilderBasicExtensions`. I suggest *CoreExtensions if we really need to differentiate.

davidfowl: Basic? Core? Why? This is the original library, everyone else can pick a worse name :smile:

davidfowl:
:shipit:

PR #581, opened 2016-01-20
Remove top level environment variables from default config

davidfowl:
We shouldn't be reading all environment variables as hosing configuration values.

/cc @JunTaoLuo @Tratcher


src/Microsoft.AspNet.Hosting/WebHostConfiguration.cs
```diff
@@ -25,7 +25,6 @@ public static IConfiguration GetDefault(string[] args)
             var configBuilder = new ConfigurationBuilder()
                 .AddInMemoryCollection(defaultSettings)
                 .AddJsonFile(WebHostDefaults.HostingJsonFile, optional: true)
-                .AddEnvironmentVariables()
```
Tratcher: Update the comment?

Tratcher:
:shipit:

PR #583, opened 2016-01-21
Log request and response headers

JunTaoLuo:
Adding logging for headers for https://github.com/aspnet/KestrelHttpServer/issues/4. This is still in progress as I check how https://github.com/aspnet/Logging/pull/338 impacts these changes.

### DIFF
--- a/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
@@ -21,23 +21,24 @@ namespace Microsoft.AspNet.Hosting.Internal
 
         public static void RequestStarting(this ILogger logger, HttpContext httpContext)
         {
+            var requestStarting = new HostingRequestStarting(httpContext);
+            if (logger.IsEnabled(LogLevel.Information))
+            {
+                logger.Log(
+                    logLevel: LogLevel.Information,
+                    eventId: LoggerEventIds.RequestStarting,
+                    state: requestStarting,
+                    exception: null,
+                    formatter: HostingRequestStarting.MessageFormatter);
+            }
             if (logger.IsEnabled(LogLevel.Debug))
             {
                 logger.Log(
                     logLevel: LogLevel.Debug,
                     eventId: LoggerEventIds.RequestStarting,
-                    state: new HostingRequestStarting(httpContext, LogLevel.Debug),
+                    state: requestStarting,
                     exception: null,
-                    formatter: HostingRequestStarting.Callback);
-            }
-            else if (logger.IsEnabled(LogLevel.Information))
-            {
-                logger.Log(
-                    logLevel: LogLevel.Information,
-                    eventId: LoggerEventIds.RequestStarting,
-                    state: new HostingRequestStarting(httpContext, LogLevel.Information),
-                    exception: null,
-                    formatter: HostingRequestStarting.Callback);
+                    formatter: HostingRequestStarting.HeaderFormatter);
             }
         }
 
@@ -47,24 +48,25 @@ namespace Microsoft.AspNet.Hosting.Internal
             if (startTimestamp != 0)
             {
                 var elapsed = new TimeSpan((long)(TimestampToTicks * (currentTimestamp - startTimestamp)));
+                var requestFinished = new HostingRequestFinished(httpContext, elapsed);
 
+                if (logger.IsEnabled(LogLevel.Information))
+                {
+                    logger.Log(
+                        logLevel: LogLevel.Information,
+                        eventId: LoggerEventIds.RequestFinished,
+                        state: requestFinished,
+                        exception: null,
+                        formatter: HostingRequestFinished.MessageFormatter);
+                }
                 if (logger.IsEnabled(LogLevel.Debug))
                 {
                     logger.Log(
                         logLevel: LogLevel.Debug,
                         eventId: LoggerEventIds.RequestFinished,
-                        state: new HostingRequestFinished(httpContext, LogLevel.Debug, elapsed),
+                        state: requestFinished,
                         exception: null,
-                        formatter: HostingRequestFinished.Callback);
-                }
-                else if (logger.IsEnabled(LogLevel.Information))
-                {
-                    logger.Log(
-                        logLevel: LogLevel.Information,
-                        eventId: LoggerEventIds.RequestFinished,
-                        state: new HostingRequestFinished(httpContext, LogLevel.Information, elapsed),
-                        exception: null,
-                        formatter: HostingRequestFinished.Callback);
+                        formatter: HostingRequestFinished.HeaderFormatter);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -189,7 +236,28 @@ public override string ToString()
             {
                 if (_cachedToString == null)
                 {
-                    _cachedToString = $"Request finished in {_elapsed.TotalMilliseconds}ms {_httpContext.Response.StatusCode} {_httpContext.Response.ContentType}";
+                    if (_logLevel <= LogLevel.Debug)
```
Tratcher: @rynowak @lodejard what do you think? This avoids the duplication of data in multiple events, and keeps Information from getting too noisy. But I don't know if it makes sense to have an event that conditionally changes levels.

rynowak: I'd probably tackle this by just making the headers a separate event.

Tratcher: Won't that make the logs harder to follow? You'd have to correlate the two events or duplicate fields in both.

JunTaoLuo: An alternative is to keep the event's `LogLevel` as `Information` but have its content change based on which level the logger is enabled at.

lodejard: Correlation isn't a concern assuming it's done inside the request is scope.

 Make it a separate event, Debug or possibly Verbose, also have one line per header with {HeaderName} and {HeaderValue} as structured message fields. Be sure the loop is inside an IsEnabled check.

That'll help when doing log analysis queries for things like distinct session id counts, user agent frequency, etc.

JunTaoLuo:
Sample logging:
```
info: Microsoft.AspNet.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:5000/
dbug: Microsoft.AspNet.Hosting.Internal.WebHost[1]
      Request Headers:
      {
          Connection: keep-alive;
          Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/we
bp,*/*;q=0.8;
          Accept-Encoding: gzip, deflate, sdch;
          Accept-Language: en-US,en;q=0.8;
          Cookie: lzHxo2Sz_Pc=CfDJ8CDfMeS_rYJBvpIwonzQwkFqC4sxLLcGOI_biYyJL98fRl
SgvH8ebGc0uP9Sf9CH11-b_Lzbf0zUD7oG8u6_eRmUpmWqQ_T8rXeuWj3hlAJHKmYtfmMh2EJdfQDe8l
NjGNIRR_X0F7CbxZoqYBEBGvg; .AspNet.Session=a7164220-9ef9-22fb-8e61-b24280bc7b5d;
 .AspNet.Cookies=CfDJ8CDfMeS_rYJBvpIwonzQwkFCMWwzdWeCny5A0g4RRJntMGPqcwQYQBJmMCp
U2WSPOb5g26tz8vEXy6qbOQ2wnWaCU9-D28_vu3pCYCefUJEbU8-2N1nAiTbNkcxDvva7fAiVFNUKxkX
ARZmrx_cGuqRnDSvuZp22hqh_eCYQh1xKVoa25SHP3heoPFqN33BW6WDe1KBsUkJky779Vq-3bvYkFmC
lAugEGYbVa7LmfQ6pr81ijvM6950XyVaPX_YA7ygEXdrh65yQ7e8SaX-bMXRK21aZj0rV1JuwvZ14j8T
LqGT-YhZWYXJBg-Hm9XaI2rSnHb9c2rpt4uIueFd3XCc; .AspNet.Correlation.Google=QoSCKZ8
PdMta01SEMBT30vtWjy_prqozZXRc6f8v5dM; SH9R3h5WiuA=CfDJ8CDfMeS_rYJBvpIwonzQwkE-T0
opXM8i6_6JOrZfIcD5humtDvCH-F2_qji1B2ffTkPW3U3kLtcjslJmpV4PsHM3fEjbeIWrr0CKn8u44u
11IzwdJtXp0iVPIwevgo3yfpGap6KUDXWqJJ6qZOSS6Bk;
          Host: localhost:5000;
          User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (K
HTML, like Gecko) Chrome/47.0.2526.111 Safari/537.36;
          Upgrade-Insecure-Requests: 1;
      }
```

lodejard:
Create two new event id numbers - don't re-use the existing [1] and [2] request started and finished

Log with a format string like "Request header {HeaderName}: {HeaderValue}"

Call log one time per header - don't call log once with all headers formatted together.
```cs
  if (logger.IsEnabled(LogLevel.Debug))
  {
    foreach(var header in httpContext.Headers)
    {
      foreach(var headerValue in header.Value)
      {
        logger.LogDebug(
          LoggerEventIds.RequestHeader,
          "Request header {HeaderName}: {HeaderValue}",
          header.Key,
          headerValue);
      }
    }
  }
```

Same thing for response. Expected output:
```
dbug: Microsoft.AspNet.Hosting.Internal.WebHost[...]
      Response header Date: Fri, 22 Jan 2016 01:47:05 GMT
dbug: Microsoft.AspNet.Hosting.Internal.WebHost[...]
      Response header Content-Length: 11
dbug: Microsoft.AspNet.Hosting.Internal.WebHost[...]
      Response header Content-Type: text/plain
dbug: Microsoft.AspNet.Hosting.Internal.WebHost[...]
      Response header Server: Kestrel
info: Microsoft.AspNet.Hosting.Internal.WebHost[2]
      Request finished in 12.5267ms 200 text/plain
```

So you don't need the other changes in the request started/finished class. This debug message isn't on by default and shouldn't be hand-coded to be performance critical like the other was.

Note for the future - we should also have a flag which lets you omit the category name and event id, to produce the output:
```
dbug: Response header Date: Fri, 22 Jan 2016 01:47:05 GMT
dbug: Response header Content-Length: 11
dbug: Response header Content-Type: text/plain
dbug: Response header Server: Kestrel
info: Request finished in 12.5267ms 200 text/plain
```

Oh, goodness. Did we move a class that's doing the logging to the internal namespace? We probably don't want the ".Internal." to be part of the category, do we?

Tratcher:
Did you check with @blowdart about omitting the values of specific headers? E.g. auth, cookies, etc?

blowdart:
It certainly doesn't look that way.

We spent a bunch of time with the privacy folks discussing what we should and should not log by default. Where did that work go?

You cannot log auth headers or cookies by default. 

Until this respects privacy settings it should not be merged.

JunTaoLuo:
@blowdart Is there a list of headers to omit? Also, is it okay to log the existence of those headers, for example, `"SensitiveHeader": "omitted"` or should we avoid that too?

blowdart:
You can log the header presence, but not the value. I will try to dig out the discussion again later in t he day

Tratcher:
In the case of cookies it would be helpful to know at least the keys. Would something like this be acceptable?
`Cookie: cookie1=[omitted], cookie2=[omitted]`
This might be a little harder for Set-Cookie.

blowdart:
* Any Authorization headers
* Any cookie which has the default cookie name that cookie auth middleware uses.
* Any cookie which has the default cookie name that session uses
* Any form field with Password in its name (I know you're not doing form fields, but just in case you consider expanding scope)

lodejard:
@blowdart I completely agree - in fact I would say it's impossible to know what headers contain sensitive information. It would need to bake in knowledge of what is safe to log, not what is unsafe to redact.

And now that you mention it - this is **exactly** like the problem introduced by logging EF field values. We added and subsequently removed the extra logger factory filtering because of it. The reason we removed it is because DiagnosticSource is sufficient to expose the information to inprocess observation, does not have security implications, and logging database values means you're dumping arbitrarily large and unique values into log storage.

New recommendation: **cut this feature**. If we think of specific headers that are highly significant we may log them as individual header messages (like `LogInformation("User-Agent: {UserAgent}", (string)httpContext.Request.Headers["User-Agent"]`) but as a rule of thumb our code shouldn't log incoming/outgoing values of unknown informational purpose. In fact the logging for known headers should probably be moved to better layers than hosting.

I updated the original issue to reflect this :)

blowdart:
If you wanted to keep it you could turn it around. Have a list of headers that we know are safe, and then let users add headers to them if they want.

lodejard:
Not worth adding an entire feature for that when logging request/response headers from the startup class is "Hello World" level of code. 

```cs
  app.Use(async (ctx,next) => {_logger.LogInformation(...); await next();});
```

It wasn't until you mentioned sensitive data that I realized this is exactly like the EF case - a key/value collection where the keys are app-specific, the values are user-activity-specific, and their meaning can't be known a priori. 

So I really do think it is the right thing to come to the same conclusion from the same principles - we should send the http context to DiagnosticSource (we already do) - but not send the headers to logger. We should probably also double-check that MVC model binding is not logging "values" as it goes for exactly the same reason. We'll be putting credit card numbers into log files left and right.

JunTaoLuo:
I will close this PR then. I open a new one in WebListener to remove the logging of headers there.

PR #584, opened 2016-01-21
React to Logging API changes

BrennanConroy:
https://github.com/aspnet/Logging/issues/331

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Diagnostics;
+using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 
@@ -52,7 +53,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             logger.LogError(
                 eventId: LoggerEventIds.ApplicationStartupException,
                 message: "Application startup exception",
-                error: exception);
+                exception: exception);
         }
 
         public static void Starting(this ILogger logger)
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -6,6 +6,7 @@
 using System.Collections.Generic;
 using Microsoft.AspNet.Http;
 using Microsoft.Extensions.Logging;
+using System.Collections;
```
BrennanConroy: fix

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -108,29 +133,66 @@ public override string ToString()
...
             private readonly HttpRequest _request;
 
             private string _cachedToString;
-            private IEnumerable<KeyValuePair<string, object>> _cachedGetValues;
+
+            public int Count
+            {
+                get
+                {
+                    return 9;
```
JunTaoLuo: This doesn't look right. Magic numbers and hard coding indexers sounds like a bad idea. Why is this needed? cc @Tratcher @loudej

src/Microsoft.AspNet.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -108,29 +133,66 @@ public override string ToString()
...
 
-                return _cachedGetValues;
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return GetEnumerator();
             }
         }
 
-        private class HostingRequestStarting : ILogValues
+        private class HostingRequestStarting : IReadOnlyList<KeyValuePair<string, object>>
```
JunTaoLuo: Is there a reason to implement `IReadOnlyList` if we provide a message formatter already?

src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -108,29 +133,66 @@ public override string ToString()
...
 
-                return _cachedGetValues;
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return GetEnumerator();
             }
         }
 
-        private class HostingRequestStarting : ILogValues
+        private class HostingRequestStarting : IReadOnlyList<KeyValuePair<string, object>>
```
nblumhardt: Per #592 - `HostingRequestFinished` below requires updating as well.

PR #589, opened 2016-01-25
Print the application base path out

davidfowl:
- This helps to show how the host is configured

/cc @Tratcher @JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Threading;
 using Microsoft.AspNetCore.Server.Features;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.PlatformAbstractions;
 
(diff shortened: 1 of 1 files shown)


JunTaoLuo:
:shipit: Yikes travis is so broken.

davidfowl:
Looking at @NTaylorMullen :imp:. Coherence has been down all weekend...

Tratcher:
I think this crosses from friendly startup information to debug output. Why not log it instead? Loggers should be available at this point.

davidfowl:
@Tratcher it's part of the relevant configuration for a web application. It's relevant for things like the wwwroot, and for mvc the views folder. Not sure why this would be verbose.

Tratcher:
The server addresses and shutdown message are actionable.  ApplicationBasePath is only displayed for diagnostic purposes. Diagnostics should be written to Logging so you can control the signal-to-noise ratio.

davidfowl:
> ApplicationBasePath is only displayed for diagnostic purposes. Diagnostics should be written to Logging so you can control the signal-to-noise ratio.

Just like the environment? I think the argument is weak. Everything printed in for diagnostics, this is no different.

Issue #529, opened 2015-12-19
WebApplicationBuilder.ConfigureServices should be Func<IServiceCollection, IServiceProvider>


pranavkm:
It's currently `Action<IServiceCollection>` and goes on to use `BuildServiceProvider` to construct a service collection.

Slightly related - we could move `ServiceCollection` to DI.Abstractions so you don't have to import `Microsoft.Extensions.DependencyInjection` to instantiate the type.

Tratcher:
:-1: `Action<IServiceCollection>` is far more common. `Func<IServiceCollection, IServiceProvider>` is for a specialized scenario. We can't have both because that causes resolution conflicts. We could have one with a different name like `SetServiceProvider(Func<IServiceCollection, IServiceProvider>)`.

pranavkm:
> Func<IServiceCollection, IServiceProvider> is for a specialized scenario

Huh? It's the only way you can plug in a DI system into the application.

Tratcher:
Yes, specialized

davidfowl:
There's really 2 sets of ConfigureServices method and with the latest API rev, we cut one of them. There's UseStartup<T> where you can override Configure and ConfigureServices like normal and there's ConfigureServices which lets you configure hosting services, it does *NOT* map to configure services in the startup method (though that might be confusing). The latter we chose to expose because we didn't see the value in the host being able to override application services in Startup.

pranavkm:
Should it be renamed to `ConfigureHostingServices`? The current name doesn't help in distinguishing between the two.

davidfowl:
@pranavkm It's bit weird to override the app's `ConfigureServices` and `Configure` separately. It would need to be a UseStartup overload:

```C#
var application = new WebApplicationBuilder()
        .UseStartup(services =>
         {
               services.AddMvc();

               // Use autofac
               var containerBuilder = new ContainerBuilder();
               containerBuilder.Populate(services);
               return containerBuilder.Build().Resolve<IServiceProvider>();
         },
         app => 
         { 
              app.UseMvc();
         })
         .Build();
```

pranavkm:
If you've seen how we use this in Mvc, we want the goodness of Startup discovery (via `UseStartup`) but need to setup enough things for the TestHost to execute it (via `ConfigureServices`).

davidfowl:
Then why do you care about being able to specify the service provider? Where is the use case?

pranavkm:
[...]

PR #591, opened 2016-01-25
Add method to WebHostBuilder to allow configure IServiceProvider

joplaal:
Added a method to allow setting the ConfigureServicesDelegate in StartupMethods, in order to allow using a custom ServiceProvider. 
If not, in scenarios like this one: 

    var testServer = new TestServer(new WebHostBuilder().
      Configure(ConfigureTestServerMethod).
      ConfigureServices(ConfigureTestServerServicesMethod));

it is not possible to return a custom IServiceProvider loaded from your DI framework of choice. That can be done using the "UseStartup" way.

Example of usage (structuremap): 

     var testServer = new TestServer(
       new WebHostBuilder().
       Configure(app => { ...},
                (services, serviceProvider) => 
                           var container = new Container();
                           container.Configure(....);
                           container.Populate(services);
                           return container.GetInstance<IServiceProvider>(); )));

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
@@ -63,6 +63,14 @@ namespace Microsoft.AspNetCore.Hosting
         IWebHostBuilder Configure(Action<IApplicationBuilder> configureApplication);
 
         /// <summary>
+        /// Specify the startup method to be used to configure the web application and the delegate that is used to configure the services of the web application. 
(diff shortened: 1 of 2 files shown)


Tratcher:
I think we already decided against this: https://github.com/aspnet/Hosting/issues/529

davidfowl:
Yes, we're not supporting this.

joplaal:
Sorry, I did not read : https://github.com/aspnet/Hosting/issues/529 before. I left there a comment with my point of view.

muratg:
@joplaal Thanks for the PR. We have decided not to do this

Issue #582, opened 2016-01-21
Settings always override UseConfiguration, no matter what order they were specified in


davidfowl:
```c#

var host = new WebHostBuilder()
                .UseApplicationBasePath(Directory.GetCurrentDirectory())
                .UseDefaultConfiguration(args)
                .UseIISPlatformHandlerUrl()
                .UseStartup<Startup>()
                .Build();

host.Run();
```

In the above example, the application base will *always* be set to the current directory.

Because we still have 2 sources of truth, an `IConfiguration` and a settings dictionary and the settings are always applied after https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/WebHostBuilder.cs#L193-L196 . Ideally things would be order sensitive so that it works the way you'd expect but that would mean need to either expand the `IConfiguration` inline or only use an `IConfiguration` (somehow).

/cc @divega @JunTaoLuo @Tratcher

divega:
I am not sure I understand the problem correctly, but what if `WebHostBuilder` *had* a `ConfigurationBuilder` member? I.e. we move the creation of the `ConfigurationBuilder` to the constructor of the WebHostBuilder and then each UseXyz methods that need to introduce configuration settings can add their own configuration providers to the `ConfiguraitonBuilder` in the order in which they are called.

davidfowl:
@divega Don't like that because the UseXyz() introduce a single setting into the existing configuration. I also explored exposing the configuration builder but ordering is still a problem. It's even worse because you can't get ahead of any other provider set before you. That's why I think it's really just a settings API (key value pairs).

Tratcher:
I don't expect it to cause too much confusion, especially if our samples specify the config first.

davidfowl:
We need the ordering to work in general. Here's what our template will look like https://github.com/davidfowl/dotnetcli-aspnetcore/blob/master/HelloWeb/Program.cs. The ordering is intentional:

Set defaults in code, override with configuration, overwrite with iis platform handler if specified. This allows us to set the application base in VS as an override to the current directory.

Tratcher:
Wait, you want to set the server in code and then override it from config? That's even more confusing. Code should always have the highest precedence. Environment vars, config files, command line, code.

/cc: @DamianEdwards

DamianEdwards:
I agree with @davidfowl, the order is what matters, not where it came from. In the end, it's all code-based config, as you have to have the lines that load from the other places (inc. the defaults).

[...]

PR #596, opened 2016-01-26
Order sensitive configuration

JunTaoLuo:
#582

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
@@ -2,10 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Server;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.Hosting
@@ -19,20 +17,6 @@ namespace Microsoft.AspNetCore.Hosting
         /// Builds an <see cref="IWebHost"/> which hosts a web application.
         /// </summary>
         IWebHost Build();
-        
-        /// <summary>
-        /// Gets the raw settings to be used by the web host. Values specified here will override 
-        /// the configuration set by <see cref="UseConfiguration(IConfiguration)"/>.
-        /// </summary>
-        IDictionary<string, string> Settings { get; }
-
-        /// <summary>
-        /// Specify the <see cref="IConfiguration"/> to be used by the web host. If no configuration is
-        /// provided to the builder, the default configuration will be used. 
-        /// </summary>
-        /// <param name="configuration">The <see cref="IConfiguration"/> to be used.</param>
-        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
-        IWebHostBuilder UseConfiguration(IConfiguration configuration);
 
         /// <summary>
         /// Specify the <see cref="IServerFactory"/> to be used by the web host.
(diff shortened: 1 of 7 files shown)


JunTaoLuo:
cc @Tratcher @davidfowl

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -187,15 +192,11 @@ public IWebHost Build()
...
-            // Apply the configuration settings
-            var configuration = _config ?? WebHostConfiguration.GetDefault();
-
-            var mergedConfiguration = new ConfigurationBuilder()
-                                .Add(new IncludedConfigurationProvider(configuration))
-                                .AddInMemoryCollection(_settings)
-                                .Build();
+            if (_config == null)
+            {
+                _config = WebHostConfiguration.GetDefault();
```
JunTaoLuo: I'm preserving the behaviour that if no configuration is set at all, the default configuration will be used.

src/Microsoft.AspNetCore.TestHost/TestServer.cs
```diff
@@ -22,7 +22,7 @@ public class TestServer : IServer
 
         public TestServer(IWebHostBuilder builder)
         {
-            if (!builder.Settings.ContainsKey(WebHostDefaults.CaptureStartupErrorsKey))
+            if (builder.Configuration[WebHostDefaults.CaptureStartupErrorsKey] == null)
```
Tratcher: string.IsNullOrEmpty

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
Tratcher: This is almost never going to execute now because most of the UseSettingFoo will populate _config. Just do `new ConfigurationBuilder().AddInMemoryCollection().Build();` here to avoid unpleasant surprises.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -147,6 +147,98 @@ public void UseCaptureStartupErrorsHonored()
...
+        {
+            var hostBuilder = new WebHostBuilder()
+                .UseSetting(WebHostDefaults.ServerKey, "ServerA")
+                .UseSetting(WebHostDefaults.ServerKey, "ServerB")
+                .UseServer(new TestServer())
+                .UseStartup<StartupNoServices>();
+            
+            var host = (WebHost)hostBuilder.Build();
+
+            Assert.Equal("ServerB", host.ServerFactoryLocation);
```
Tratcher: You don't actually verify that you end up with TestServer as your server.

JunTaoLuo: Not sure if that's relevant for the test. I just want to make sure that when we build the host, the correct configuration settings are parsed. Whether they are used or not doesn't really matter.

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -24,7 +24,7 @@ public interface IWebHostBuilder
         /// Gets the raw settings to be used by the web host. Values specified here will override 
         /// the configuration set by <see cref="UseConfiguration(IConfiguration)"/>.
         /// </summary>
-        IDictionary<string, string> Settings { get; }
+        IConfiguration Configuration { get; }
```
davidfowl: Nuke this and add GetSetting

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -80,7 +82,10 @@ public IWebHostBuilder UseSetting(string key, string value)
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
         public IWebHostBuilder UseConfiguration(IConfiguration configuration)
```
davidfowl: Move this to an extension method that copies things into settings.

davidfowl:
:shipit:

PR #598, opened 2016-02-01
Resolving relative application base paths

moozzyk:
This allows providing relative paths in ASPNET_APPLICATIONBASE variable

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Runtime.Versioning;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Builder;
@@ -84,7 +85,7 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         /// <summary>
-        /// Specify the startup type to be used by the web host. 
+        /// Specify the startup type to be used by the web host.
         /// </summary>
         /// <param name="startupType">The <see cref="Type"/> to be used.</param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -224,7 +225,7 @@ private class WrappedApplicationEnvironment : IApplicationEnvironment
         {
             public WrappedApplicationEnvironment(string applicationBasePath, IApplicationEnvironment env)
             {
-                ApplicationBasePath = applicationBasePath;
+                ApplicationBasePath = Path.GetFullPath(applicationBasePath);
```
Tratcher: This resolves relative to CurrentDirectory, which is very arbitrary, no? I thought you were going to resolve relative to the exe path?

moozzyk: When we talked the idea was to resolve relative to the .exe but for a deployed app isn't it more useful/natural to resolve it relative to a current directory?

davidfowl: No, it should be resolved based on the exe directory

moozzyk: OK

moozzyk:
:up::date:

moozzyk:
:watch:

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -238,5 +239,22 @@ public WrappedApplicationEnvironment(string applicationBasePath, IApplicationEnv
...
+            }
+
+            var basePath =
+#if NET451
+                (string)AppDomain.CurrentDomain.GetData("APP_CONTEXT_BASE_DIRECTORY") ??
+                AppDomain.CurrentDomain.BaseDirectory;
+#else
+                AppContext.BaseDirectory;
+#endif
+            return Path.Combine(Path.GetFullPath(basePath), applicationBasePath);
```
Tratcher: In which cases is basePath relative such that you need GetFullPath?

moozzyk: Fair point. As per our conversation I will do GetFullPath on the combined result to collapse unneeded traversals.

PR #599, opened 2016-02-03
Fixes #597: Hosting logging needs to use structured logging instead o…

### LINKED ISSUE #597: Hosting logging needs to use structured logging instead of format strings
E.g. here: https://github.com/aspnet/Hosting/blob/3ea44c6f75c4e74b10942033dae982f0490cb65a/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs#L269

It's using a formatted string, which means that structured logging doesn't get to see what's in there. It also means that the formatting is being done in the current locale of the thread, which is unlikely to achieve the desired effect.

The logging calls need to instead use structured logging.

kichalla:
…f format strings

@Eilon @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -126,7 +127,11 @@ public override string ToString()
             {
                 if (_cachedToString == null)
                 {
-                    _cachedToString = $"RequestId:{_httpContext.TraceIdentifier} RequestPath:{_httpContext.Request.Path}";
+                    _cachedToString = string.Format(
```
kichalla: I suppose this doesn't need to be changed? (initially thought TraceIdentifier was a number but its not)

src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -201,7 +206,18 @@ public override string ToString()
...
+                        "Request starting {0} {1} {2}://{3}{4}{5}{6} {7} {8}",
+                        _request.Protocol,
+                        _request.Method,
+                        _request.Scheme,
+                        _request.Host,
+                        _request.PathBase,
+                        _request.Path,
+                        _request.QueryString,
+                        _request.ContentType,
+                        _request.ContentLength);
```
kichalla: is a `long`

Eilon:
:shipit:

PR #606, opened 2016-02-17
Convert TFM to net451 to allow custom test discovery

JunTaoLuo:
This will be required for tests downstream that have a dependency on Microsoft.AspNetCore.Server.Testing.


Tratcher:
:shipit:

PR #608, opened 2016-02-18
Enabled xml doc generation

ajaybhargavb:
@NTaylorMullen

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
@@ -19,7 +19,8 @@ namespace Microsoft.AspNetCore.Hosting.Server
         /// <summary>
         /// Start the server with an HttpApplication.
         /// </summary>
-        /// <param name="application">An instance of <see cref="IHttpApplication"/>.</param>
+        /// <param name="application">An instance of <see cref="IHttpApplication{TContext}"/>.</param>
+        /// <typeparam name="TContext">The context associated with the HttpApplication.</typeparam>
         void Start<TContext>(IHttpApplication<TContext> application);
     }
 }
--- a/NuGetPackageVerifier.json
+++ b/NuGetPackageVerifier.json
@@ -1,29 +1,20 @@
 {
     "adx": { // Packages written by the ADX team and that ship on NuGet.org
         "rules": [
-            "AssemblyHasDocumentFileRule",
-            "AssemblyHasVersionAttributesRule",
-            "AssemblyHasServicingAttributeRule",
-            "AssemblyHasNeutralResourcesLanguageAttributeRule",
-            "SatellitePackageRule",
-            "StrictSemanticVersionValidationRule"
+            "AdxVerificationCompositeRule"
         ],
         "packages": {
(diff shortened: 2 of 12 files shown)


src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
```diff
@@ -19,7 +19,7 @@ public interface IServer : IDisposable
         /// <summary>
         /// Start the server with an HttpApplication.
         /// </summary>
-        /// <param name="application">An instance of <see cref="IHttpApplication"/>.</param>
+        /// <param name="application">An instance of <see cref="IHttpApplication{TContext}"/>.</param>
         void Start<TContext>(IHttpApplication<TContext> application);
```
NTaylorMullen: Hmm, is it a bug that we don't have to require the type param ref?

NTaylorMullen:
:shipit:

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
NTaylorMullen: Talked offline

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
```diff
@@ -19,7 +19,8 @@ public interface IServer : IDisposable
         /// <summary>
         /// Start the server with an HttpApplication.
         /// </summary>
-        /// <param name="application">An instance of <see cref="IHttpApplication"/>.</param>
+        /// <param name="application">An instance of <see cref="IHttpApplication{TContext}"/>.</param>
+        /// <typeparam name="TContext">The context associated with the HttpApplication.</typeparam>
```
Eilon: Not sure where `HttpApplication` came from - is that a type or something? Should it just say `application`?

Eilon: Or is it `IHttpApplication<T>`, in which case link it?

ajaybhargavb: I just used the same wording we have already used https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IHttpApplication.cs#L11 and https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs#L20.
But there is no `HttpApplication` as such.

Eilon: Yuck. I would just say "application" then (lowercase). Can you patch up those two comments as well? I'm guessing they're copied from Katana or something.

Tratcher: Nope, this all has to do with the gross generics.

ajaybhargavb: Done. 19501c03aa5d1973d7e5237b66af063dc21bb902

PR #609, opened 2016-02-19
Use HostingNullLogScope when no logging

### LINKED ISSUE #367: [Perf] High Allocator LoggingScope when no loggers

benaadams:
Drops the `HostingLogScope` from this 
After-QueueUserWorkItemCallback

Resolves #367

/cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
@@ -14,10 +14,11 @@ namespace Microsoft.AspNetCore.Hosting.Internal
     internal static class HostingLoggerExtensions
     {
         private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+        private static readonly object _nullScope = new HostingNullLogScope();
 
         public static IDisposable RequestScope(this ILogger logger, HttpContext httpContext)
         {
-            return logger.BeginScopeImpl(new HostingLogScope(httpContext));
+            return logger.BeginScopeImpl(!logger.IsEnabled(LogLevel.Critical) ? _nullScope : new HostingLogScope(httpContext));
         }
 
         public static void RequestStarting(this ILogger logger, HttpContext httpContext)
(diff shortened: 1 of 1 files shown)


davidfowl:
Checking IsEnabled for the Hosting logger isn't enough. You need to be able to check if there are any logging providers at all and there's no way to do that.

/cc @lodejard @rynowak

benaadams:
DI or lambda \ `Func<object, object>` scope?

e.g.
```csharp
return logger.BeginScopeImpl((state) => new HostingLogScope((HttpContext)state), httpContext);
```

src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -306,6 +307,78 @@ IEnumerator IEnumerable.GetEnumerator()
...
+
+                object IEnumerator.Current
+                {
+                    get
+                    {
+                        return null;
+                    }
+                }
+
+                public void Dispose()
```
SajayAntony: Elide?

benaadams:
Need to see how this goes https://github.com/aspnet/Logging/issues/366

davidfowl:
Closing until the logging issue is resolved.

PR #611, opened 2016-02-19
Replacing KillProcess with taskkill /t which kills entire process tree

moozzyk:


src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -123,6 +122,35 @@ protected void ShutDownIfAnyHostProcess(Process hostProcess)
...
+        private void KillProcess(int? processId)
+        {
+            if (processId == null)
+            {
+                return;
+            }
+
+            ProcessStartInfo startInfo;
+
+            if (PlatformServices.Default.Runtime.OperatingSystem.Equals("Windows", StringComparison.OrdinalIgnoreCase))
```
victorhurdugaci: This will never change during a test run :) Cache it

moozzyk: OK

src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -106,8 +106,7 @@ protected void ShutDownIfAnyHostProcess(Process hostProcess)
             if (hostProcess != null && !hostProcess.HasExited)
             {
                 // Shutdown the host process.
-                hostProcess.Kill();
-                hostProcess.WaitForExit(5 * 1000);
+                KillProcess(hostProcess.Id);
```
victorhurdugaci: This can never be null

victorhurdugaci:
:shipit:

moozzyk:
:up::date:

PR #616, opened 2016-02-25
Update the self host deployer message so published application comman…

JunTaoLuo:
…d is accurate

otherwise we get messages like 
image

cc @BrennanConroy @kichalla


kichalla:
:shipit:

PR #617, opened 2016-02-26
Append .exe to executable when deploying published CLR applications

JunTaoLuo:
I'm looking into Entropy test failures on Linux and I noticed that to deploy CLR projects, the .exe extension is still required. CoreCLR projects don't need the extension

cc @kichalla

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
@@ -53,7 +53,8 @@ namespace Microsoft.AspNetCore.Server.Testing
             string executableArgs = string.Empty;
             if (DeploymentParameters.PublishApplicationBeforeDeployment)
             {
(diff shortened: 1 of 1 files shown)


kichalla:
:shipit:

src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
```diff
@@ -53,7 +53,8 @@ private CancellationToken StartSelfHost(Uri uri)
             string executableArgs = string.Empty;
             if (DeploymentParameters.PublishApplicationBeforeDeployment)
             {
-                executableName = Path.Combine(DeploymentParameters.PublishedApplicationRootPath, new DirectoryInfo(DeploymentParameters.ApplicationPath).Name);
+                var executableExtension = DeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr ? ".exe" : "";
```
kichalla: Should we even check for the OS platform as even in case of Mono on Linux/Mac, we do not want this extension

PR #620, opened 2016-02-26
Fix how processes are killed on Linux

JunTaoLuo:
I'm seeing hangs when tests use deployers. We need to kill the process itself and not just its children. Tested this change with Entropy on Linux.

cc @moozzyk

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
@@ -27,7 +27,6 @@ namespace Microsoft.AspNetCore.Server.Testing
 
         private readonly Stopwatch _stopwatch = new Stopwatch();
 
-
         public ApplicationDeployer(DeploymentParameters deploymentParameters, ILogger logger)
         {
             DeploymentParameters = deploymentParameters;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -141,8 +140,8 @@ private void KillProcess(int processId)
             {
                 startInfo = new ProcessStartInfo
                 {
-                    FileName = "pkill",
-                    Arguments = $"-TERM -P {processId}",
+                    FileName = "kill",
+                    Arguments = $"-TERM {processId}",
                 };
```
moozzyk: Interesting. I just stole this code from dotnet-watch and the idea was that it would kill process tree. Apparently this is not the case - now when I read I see it only seems to kill child processes and not the process itself. The new command however seems to kill only the process but not the child process which won't work if Deployer creates child processes.

kichalla: I was looking at this and from the following link it says we need to have a `-` in front of the process id
http://fibrevillage.com/sysadmin/237-ways-to-kill-parent-and-child-processes-in-one-command

victorhurdugaci:
Does it kill an entire process tree? Does it work on mac?

src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
JunTaoLuo: Strange that this does not work:
image
I will see what else we can do. In the worst case we can combine `pkill -TERM -P {processId}` and `kill -TERM {processId}` to kill the children then kill the parent but I feel that there must be a better way.

victorhurdugaci: You don't know the ID of the child process

kichalla: @JunTaoLuo did you try just `kill  -TERM -9874`

JunTaoLuo: Same error with or without `--`

JunTaoLuo:
Verified this approach on Linux and Mac. Maybe we should check this in to stop hangs on Travis and look for a more elegant solution?

moozzyk:
@JunTaoLuo - If it stops hangs :ship::it:

Issue #613, opened 2016-02-25
WebHostBuilder should have an option to set the IApplicationEnvironment.ApplicationName


pranavkm:
We could default this to the `StartupAssemblyName` \ `WebHostOptions.Application`. In addition, it would be useful to be able to set these properties independently of each other.

pranavkm:
More background: In Mvc, we treat the `IApplicationEnvironment.ApplicationName` as the entry assembly. This is required to make controller discovery work when the actual entry point is the test runner but the test wants Mvc to treat the test assembly as the entry point. It'd be nice to take advantage of the `IApplicationEnvironment` that the builder creates rather than creating and injecting a mock one in the test,

muratg:
Looks like a good change.  Assigning to @JunTaoLuo

JunTaoLuo:
Merged to dev

PR #625, opened 2016-03-03
Configure ApplicationName of ApplicationEnvironment on WebHostBuilder

JunTaoLuo:
#613 

 cc @Tratcher @pranavkm

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Reflection;
 using System.Runtime.Versioning;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Builder;
@@ -199,10 +200,9 @@ namespace Microsoft.AspNetCore.Hosting
                 if (defaultPlatformServices.Application != null)
                 {
                     var appEnv = defaultPlatformServices.Application;
-                    if (!string.IsNullOrEmpty(_options.ApplicationBasePath))
-                    {
-                        appEnv = new WrappedApplicationEnvironment(_options.ApplicationBasePath, appEnv);
-                    }
+                    var applicationBasePath = ResolveApplicationBasePath(_options.ApplicationBasePath, appEnv.ApplicationBasePath);
+                    var startupAssemblyName = ResolveStartupAssemblyName() ?? appEnv.ApplicationName;
+                    appEnv = new WrappedApplicationEnvironment(applicationBasePath, startupAssemblyName, defaultPlatformServices.Application);
 
                     services.TryAddSingleton(appEnv);
                 }
@@ -221,12 +221,59 @@ namespace Microsoft.AspNetCore.Hosting
             return services;
         }
 
+        private string ResolveApplicationBasePath(string applicationBasePath, string basePath)
+        {
+            if (_startup != null)
+            {
+                var startupAssemblyLocation = _startup.ConfigureDelegate.Target.GetType().GetTypeInfo().Assembly.Location;
+                if (!string.IsNullOrEmpty(startupAssemblyLocation))
+                {
+                    return Path.GetDirectoryName(startupAssemblyLocation);
+                }
+            }
+            else if (_startupType != null)
+            {
+                var startupAssemblyLocation = _startupType.GetTypeInfo().Assembly.Location;
+                if (!string.IsNullOrEmpty(startupAssemblyLocation)) 
+                {
+                    return Path.GetDirectoryName(startupAssemblyLocation);
+                }
+            }
+
+            if (string.IsNullOrEmpty(applicationBasePath))
+            {
+                return basePath;
+            }
+            if (Path.IsPathRooted(applicationBasePath))
+            {
+                return applicationBasePath;
+            }
+            return Path.Combine(Path.GetFullPath(basePath), applicationBasePath);
+        }
+
+        private string ResolveStartupAssemblyName()
+        {
+            if (_startup != null)
+            {
+                return _startup.ConfigureDelegate.Target.GetType().GetTypeInfo().Assembly.GetName().Name;
+            }
+            if (_startupType != null)
+            {
+                return _startupType.GetTypeInfo().Assembly.GetName().Name;
+            }
+            if (!string.IsNullOrEmpty(_options.Application))
+            {
+                return _options.Application;
+            }
+            return null;
+        }
+
         private class WrappedApplicationEnvironment : IApplicationEnvironment
         {
-            public WrappedApplicationEnvironment(string applicationBasePath, IApplicationEnvironment env)
+            public WrappedApplicationEnvironment(string applicationBasePath, string applicationName, IApplicationEnvironment env)
             {
-                ApplicationBasePath = ResolvePath(applicationBasePath, env.ApplicationBasePath);
-                ApplicationName = env.ApplicationName;
+                ApplicationBasePath = applicationBasePath;
+                ApplicationName = applicationName;
                 ApplicationVersion = env.ApplicationVersion;
                 RuntimeFramework = env.RuntimeFramework;
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
```diff
@@ -32,6 +33,8 @@ public WebHostOptions(IConfiguration configuration)
 
         public string Application { get; set; }
 
+        public string EnvironmentApplication { get; set; }
```
Tratcher: :-1: naming.

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
```diff
@@ -22,6 +22,7 @@ public WebHostOptions(IConfiguration configuration)
             }
 
             Application = configuration[WebHostDefaults.ApplicationKey];
+            EnvironmentApplication = configuration[WebHostDefaults.EnvironmentApplicationKey];
```
pranavkm: What's the difference between `Application` and this?

JunTaoLuo: I wasn't sure if you wanted to configure the two separately. `Application` is used for loading the entry assembly. `EnvironmentApplication` is used to set the `ApplicationName` of the `IApplicationEnvironment` service.

JunTaoLuo: I can also just default the `ApplicationName` of the `IApplicationEnvironment` service to what is set by `Application`. Note that only `UseStartup(string)` or configuration will set this variable not `UseStartup(Type)` or `UseStartup<Type>()`.

pranavkm: Sounds like the two should be the same. I'd run it by @davidfowl since this was originally his suggestion.

JunTaoLuo: How would MVC want to use it? I think this feature is used because MVC needed it?

pranavkm: When running functional tests, we create a new `IApplicationEnvironment` and set the `ApplicationName` and `ApplicationBasePath` to the the website being run. See https://github.com/aspnet/Mvc/blob/dev/test/Microsoft.AspNetCore.Mvc.FunctionalTests/MvcTestFixture.cs#L70

JunTaoLuo: If that is the only use case then we don't need to have the two settings and it is sufficient to just use the `Application` setting to set the `ApplicationName` of the `IApplicationEnvironment` service. Note that you still need to adjust the `UseStartup<Type>()` to `UseStartup(typeof(Type).Assembly.GetName().Name)` in the test fixture.

pranavkm: Yup, we already do that.

src/Microsoft.AspNetCore.Hosting.Abstractions/WebHostDefaults.cs
```diff
@@ -6,6 +6,7 @@ namespace Microsoft.AspNetCore.Hosting
     public static class WebHostDefaults
     {
         public static readonly string ApplicationKey = "application";
+        public static readonly string EnvironmentApplicationKey = "environmentApplication";
```
davidfowl: why is this name so bad?

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
davidfowl: The 2 concepts are:

- Startup assembly name - Where do I look for the startup class
- Application entry point -  What is the name of my application entry point

Are those 2 things always the same? I don't think they are.

src/Microsoft.AspNetCore.Hosting.Abstractions/WebHostDefaults.cs
JunTaoLuo: Well a more descriptive one would be `ApplicationEnvironmentApplicationName` but maybe that's a little too verbose? I'm just prepending what we use for Application currently.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -199,10 +200,9 @@ private IServiceCollection BuildHostingServices()
                 if (defaultPlatformServices.Application != null)
                 {
                     var appEnv = defaultPlatformServices.Application;
-                    if (!string.IsNullOrEmpty(_options.ApplicationBasePath))
-                    {
-                        appEnv = new WrappedApplicationEnvironment(_options.ApplicationBasePath, appEnv);
-                    }
+                    var applicationBasePath = ResolvePath(_options.ApplicationBasePath, appEnv.ApplicationBasePath);
+                    var environmentApplicationName = ResolveStartupAssemblyName() ?? appEnv.ApplicationName;
```
JunTaoLuo: I don't think there will ever be a case where we default to `appEnv.ApplicationName` since we need to specify a startup assembly/type/method to build the webhost. But I guess keeping the fallback doesn't hurt.

JunTaoLuo:
:up::date:

Tratcher:
Better. @pranavkm, is that what you were after?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
pranavkm: Yeah this would work for us. Is @davidfowl ok with this behavior?

davidfowl: Yep!

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -221,12 +221,29 @@ private IServiceCollection BuildHostingServices()
             return services;
         }
 
+        private string ResolveStartupAssemblyName()
+        {
+            if (_startup != null)
```
davidfowl: Add some comments here

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -319,6 +319,42 @@ public void RelativeApplicationBaseAreResolved()
             Assert.EndsWith(Path.DirectorySeparatorChar + "bar", basePath);
         }
 
+        [Fact]
+        public void DefaultApplicationNameToStartupAssemblyName()
```
davidfowl: Do we have something that tests the setting?

JunTaoLuo: What do you mean test the setting? This version of `UseStartup` calls `UseSetting`.

davidfowl:
There are some other changes I'd  like to see after this change is in.

davidfowl:
Why is travis failing /cc @pranavkm

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
davidfowl: Can we add a test that uses the default configuration as well.

JunTaoLuo:
:up::date: cc @Tratcher @davidfowl

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -221,12 +221,57 @@ private IServiceCollection BuildHostingServices()
             return services;
         }
 
+        private string ResolveApplicationBasePath(string applicationBasePath, string basePath)
+        {
+            if (_startup != null)
+            {
+                var startupAssemblyLocation = Path.GetDirectoryName(_startup.ConfigureDelegate.Target.GetType().GetTypeInfo().Assembly.Location);
+                if (startupAssemblyLocation != null) {
```
davidfowl: bad formatting :smile:

JunTaoLuo: It's fine, `Path.GetDirectoryName` returns null if the path is null

davidfowl:
:shipit:

Issue #626, opened 2016-03-04
Deployment targets need to be updated for published apps


JunTaoLuo:
Currently we use the RuntimeFlavor to determine which TFM to publish for but this will no longer be valid since we may have a mix of `dnxcore50` and `netstandardapp1.x`. I propose we add another deployment parameter that tests can set to specify which target framework to actually publish for. An alternative is to parse the project's json file and look for a list of targets that are compatible with clr/coreclr but I feel that is not robust.

JunTaoLuo:
cc @BrennanConroy

JunTaoLuo:
Merged to dev, closing. Working on reactions to this change:
- [x] ServerTest
- [x] IISIntegration
- [x] MusicStore
- [ ] Entropy
- [ ] Performance
- [x] Localization, no need for update since none of the tests requires publishing before deployment

PR #627, opened 2016-03-04
Add a parameter for tests to specify TFM for publish before deploymen…

JunTaoLuo:
…t. #626 

Currently this affects IISIntegration and possibly more repos.

cc @BrennanConroy @victorhurdugaci @Tratcher


victorhurdugaci:
:shipit:  but this will break some tests (server tests, music store)

Issue #612, opened 2016-02-23
Treat logger.ApplicationError as Critical


cwe1ss:
https://github.com/aspnet/Hosting/blob/e62ceb852827d386b492abc9cc6801b1a8c70263/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs#L52

Wouldn't it be better to treat startup exceptions as LogLevel `Critical`? 

```csharp
/// <summary>
/// Logs that describe an unrecoverable application or system crash, 
/// or a catastrophic failure that requires immediate attention.
/// </summary>
Critical = 5,
```

If you have a web application that doesn't start up, it's as bad as it can get in my opinion. Using Critical would make it easier to raise alerts etc.

muratg:
This makes sense to me. @Tratcher let me know what you think.

Tratcher:
Sounds good.

PR #628, opened 2016-03-04
Elevate application starup errors to be logged as critical #612

JunTaoLuo:
cc @Tratcher


Tratcher:
:shipit:

moozzyk:
Should we have a regression test for this?

PR #640, opened 2016-03-11
Rename environment variables to ASPNETCORE_. 

Tratcher:
Also remove the legacy ENV abbreviation.
@DamianEdwards @davidfowl @muratg @JunTaoLuo @BillHiebert


BillHiebert:
@Tratcher - do you still support the old variables? ie. will we be broken when we pick up this change?

davidfowl:
Yes you'll be broken.

Make sure you write an announcement

Tratcher:
Correct, we're not supporting the old variables this time.

muratg:
:shipit:

muratg:
@Tratcher Announcement issue?

cwe1ss:
Could you please also update and/or add a comment to https://github.com/aspnet/Announcements/issues/108 which contains the table with all new names.

Tratcher:
Commented.

PR #643, opened 2016-03-13
Fixed Bug. (Port In Use)

AlirezaP:


AlirezaP:
fixed bug. (default port was in use).

AlirezaP:
fixed bug. (default port was in use).

src/Microsoft.AspNetCore.Hosting/Helper/Network.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Hosting.Helper
+{
+    public class Network
+    {
+        public static bool PortValidity(int port)
```
Tratcher: We have test infrastructure that does something similar except it checks by binding to the port as if it were going to be the server. I think it even tries to bind to 0 which causes it to generate and return a random port.

Tratcher: Method name should be a verb.

src/Microsoft.AspNetCore.Hosting/Helper/Network.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Hosting.Helper
```
Tratcher: Hosting.Internal

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -223,7 +223,22 @@ private void EnsureServer()
                 if (addresses != null && !addresses.IsReadOnly && addresses.Count == 0)
                 {
                     // Provide a default address if there aren't any configured.
-                    addresses.Add("http://localhost:5000");
+                    int port = 5000;
+
+                    //The range 49152�65535 (215+214 to 216?1) contains dynamic or private ports that cannot be registered with IANA
+                    for (int i = 49151; i < 65534; i++)
```
Tratcher: This kind of loop should be in the helper method. e.g. `FindFreePort`

Tratcher:
@DamianEdwards do we want dynamic default port selection? It makes some assumptions about the server, e.g. it may have false negatives for WebListener that can share ports. I also think the most common scenario for conflicts is when you've left a prior instance of your server running. Changing the port dynamically there could break your flow if you just went back to the browser and just hit refresh.

davidfowl:
Not sure this should be built in. It also brings in some other dependencies...

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
AlirezaP: yes.
you're right.

DamianEdwards:
Yeah I don't think this should be in the box. It's also inherently prone to race-conditions leading to random failures anyway. Much prefer to just default to the single port like we do now. The idea of having this as some sort of helper isn't bad though (although it will still be racy).

davidfowl:
Closing this out. @AlirezaP feel free to provide a different package with this functionality as an extension method on `IWebHostBuilder`

AlirezaP:
ok
thanks

Issue #641, opened 2016-03-11
ApplicationBasePath set to temporary folder


JunTaoLuo:
Using `Assembly.Location` to set the application base path does not work for tests launched with xunit console runner. 

It seems that the location of the test assembly is set to somewhere in the temporary folder such as `AppData\Local\Temp\7e915e1e-22f9-4172-8ade-491b924bf79c\7e915e1e-22f9-4172-8ade-491b924bf79c\assembly\dl3\815768c5\09292b41_dd7bd101\Microsoft.AspNetCore.Hosting.Tests.dll`. We expected the path instead to be `\test\Microsoft.AspNetCore.Hosting.Tests\bin\Debug\net451\win7-x64\Microsoft.AspNetCore.Hosting.Tests.DLL`. 

This currently breaks tests that contains content files (such as StaticFiles).

cc @davidfowl

JunTaoLuo:
`Path.Get(new Uri(Assembly.CodeBase).LocalPath)` seems to do the trick.

davidfowl:
Ohhhh shadow copying hmmmm maybe we should hair stick to the original application base path

PR #646, opened 2016-03-14
Revert ApplicationBasePath default resolution #641

JunTaoLuo:
#641 This will revert to the previous behaviour which sets the ApplicationBasePath to the default ApplicationEnvironment provided by PlatformServices and the user configured options.


JunTaoLuo:
cc @Tratcher @davidfowl

davidfowl:
:shipit:

PR #652, opened 2016-03-15
Add `ObjectPool<StringBuilder>` to default services

dougbu:
- can now use pooled `StringBuilder`s to address aspnet/HttpAbstractions#561 and similar issues

nit: Fix incomplete test noted when adding new one


test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -92,8 +93,27 @@ public void Build_honors_UseStartup_with_string()
...
+                var services = host.Services.GetServices<IApplicationLifetime>();
+                Assert.NotNull(services);
+                Assert.NotEmpty(services);
+
+                await AssertResponseContains(server.RequestDelegate, "Exception from constructor");
+            }
+        }
+
+        [Fact]
+        public async Task ObjectPoolOfStringBuilderRegistered_EvenWhenStartupCtorThrows()
```
Tratcher: We don't actually care about ObjectPool in the StartupThrows scenario. You can find a simpler test to copy.

Tratcher:
Do not merge this until we see how aspnet/HttpAbstractions#561 looks.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
dougbu: Suggestions? I don't see much that just validates services are registered.

Tratcher: `.Configure(app => { })` would be an adequate replacement for UseStartup in this test.

dougbu:
:up::date: and aspnet/HttpAbstractions#587 is out.

Tratcher:
@davidfowl Gag check?

davidfowl:
This is odd. I'm not sure this makes sense.

rynowak:
@davidfowl - what do you propose? Use a static pool? That might make the most sense if we're talking about HttpAbstractions/WebUtilities

davidfowl:
Yes, I was going to suggest a static pool for something like this with optional overloads to pass in a pool into whatever method is there (low level method).

dougbu:
>  I was going to suggest a static pool

Let's discuss this offline. We already have at least one other pattern for `ObjectPool<T>` use in other repos. Would rather move ahead w/ the general guidance on when to use what (and what to use when) in place.

dougbu:
Will move instantiation of the pool to Kestrel as decided offline.

Issue #653, opened 2016-03-15
Make RequestServicesContainerMiddleware the first middleware in the pipeline


weidazhao:
When we add a middleware through IStartupFilter, it will be added before RequestServicesContainerMiddleware in the pipeline (see the impl at https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L163) Therefore, the middleware won't be able to get any scoped service instance from HttpContext.RequestServices since RequestServices is not constructed by RequestServicesContainerMiddleware yet at the moment.

This could be fixed by ensuring RequestServicesContainerMiddleware always to be the first middleware in the pipeline. A possible fix is to move https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L202 to https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L218.

For more context and details, check out https://github.com/weidazhao/Hosting/pull/8.

weidazhao:
Took a further look at the code when I was trying to make a pull request. 

I think I understand the reason that RequestServicesContainerMiddleware is currently added after the middlewares registered by IStartupFilter so that those 3rd party middlewares can plug in their own implementations of IServiceProvidersFeature. However, IServiceProvidersFeature is now defined under Microsoft.AspNetCore.Http.Features.Internal. Is it really expected for 3rd party to provide custom implementation of IServiceProvidersFeature?

davidfowl:
> I think I understand the reason that RequestServicesContainerMiddleware is currently added after the middlewares registered by IStartupFilter so that those 3rd party middlewares.

I think it's reasonable to pick an order that allows hosts that plug into the system the ability to run after the built in middleware. The main reason we add ours to the collection first is because we want configure services to see all of the registered services so it can add/remove or replace them. If we add them after the callback then it's impossible to observe/change that specific hosting service.

> However, IServiceProvidersFeature is now defined under Microsoft.AspNetCore.Http.Features.Internal. Is it really expected for 3rd party to provide custom implementation of IServiceProvidersFeature?

@dougbu is fixing this very soon. It's going to be moved to Microsoft.AspNetCore.Http.Features. /cc @Tratcher

weidazhao:
Thanks. Fair enough.

[...]

PR #657, opened 2016-03-17
Configure IStartupFilters in FIFO order #653

JunTaoLuo:
#653 

cc @Tratcher @davidfowl

### DIFF
--- a/test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
+++ b/test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
@@ -275,7 +275,7 @@ namespace Microsoft.AspNetCore.TestHost
             {
                 app.Run(context =>
                 {
-                    Assert.NotNull(context.RequestServices);
+                    Assert.Null(context.RequestServices);
                     return context.Response.WriteAsync("Success");
                 });
             })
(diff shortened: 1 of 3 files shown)


Tratcher:
@weidazhao

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
```diff
@@ -275,7 +275,7 @@ public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
             {
                 app.Run(context =>
                 {
-                    Assert.NotNull(context.RequestServices);
+                    Assert.Null(context.RequestServices);
```
muratg: Why did this change?

davidfowl:
:shipit:

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
JunTaoLuo: Because the test was wrong previously.

muratg: :+1: :shipit:

weidazhao:
Thanks guys for the quick turnaround! I've verified on v1.0.0-rc2-20260. @davidfowl  @JunTaoLuo @muratg @Tratcher

Issue #648, opened 2016-03-14
Rename UseCaptureStartupErrors to CaptureStartupErrors


danroth27:
Consistency is not an excuse for nonsensical English :smile:

Tratcher:
How about `UseStartupErrorPage`?

danroth27:
Sure

Issue #636, opened 2016-03-10
Rename UseDefaultConfiguration extension method to UseDefaultHostingConfiguration


muratg:
[Came up in the Templates review]

rustd:
Add XML doc comments for the API as well.

rustd:
@JunTaoLuo can you please send a PR to templates with the changes.

Issue #622, opened 2016-03-02
Log startup exceptions when CaptureStartupErrors is not set


cwe1ss:
Hi! Right now, if there is a startup exception the process keeps running and prints out the exception. This makes it really hard for orchestration/monitoring tools like Azure Service Fabric to find out whether the startup succeeded or not. 

I know that this is a feature ( #370, #486 ) and this is helpful in development scenarios. However, I think that there should be an option that allows us to disable this behavior. In a production environment, these errors must be addressed with proper logging techniques anyway (ETW, ILogger, Windows Event Log, ...) to trigger alerts and so on. Crashing the process would allow Azure Service Fabric to automatically retry a few times.

I would suggest the following behavior:

* It should *not* be bound to the `Environment` variable. We might want to have this behavior locally or in our test environment as well.
* It should be a separate parameter (e.g. like `--server.urls` or a different environment variable). The default could be the current behavior.
* There should be a best-effort to catch the exception and send it to the ILogger pipeline. This way we could send the exception to our logging system - of course, only in case the logger was already initialized. AFAIK this is already the current implementation. 
* After the attempt to log the exception, the process should exit with a failure code. If possible, the exception should be sent to the .NET exception EventSource-Provider (but I think this will happen automatically!?)

davidfowl:
This already exists in rc2 as an option that can be set in the same ways as server.urls.

cwe1ss:
Hi David, thank you for your response! Looks like a lot has changed since RC1 (which is a good thing!! :-) )

Regarding the logging: [Are we talking about this code?](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L187) 

Right now, if `CaptureStartupErrors` is `false`, exceptions never enter the catch block and therefore, `logger.ApplicationError(ex);` is never called. Is there some other location that logs the error?

If not, I think the code should be changed to something like this:

```csharp
catch (Exception ex) {

  // log and crash the process
  logger.ApplicationError(ex);
  if (!_options.CaptureStartupErrors) {
    throw;
    
  // keep the server running and output the error
}
```

davidfowl:
@Eilon @lodejard @rynowak @Tratcher What did we say about logging and re-throwing? Wasn't that one of the logging no nos?

cwe1ss:
That's not great, yes. Maybe something like this would be better:

```csharp
catch (Exception ex) {

  // log and crash the process
[...]

PR #660, opened 2016-03-18
WebHostBuilder extensions rename and documentation. #636 #648

JunTaoLuo:
#636 #648

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
@@ -12,96 +12,72 @@ namespace Microsoft.AspNetCore.Hosting
     {
         private static readonly string ServerUrlsSeparator = ";";
 
-        public static IWebHostBuilder UseDefaultConfiguration(this IWebHostBuilder builder)
+        /// <summary>
+        /// Use the default hosting configuration settings on the web host.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder UseDefaultHostingConfiguration(this IWebHostBuilder hostBuilder)
         {
-            return builder.UseDefaultConfiguration(args: null);
+            return hostBuilder.UseDefaultHostingConfiguration(args: null);
         }
 
-        public static IWebHostBuilder UseDefaultConfiguration(this IWebHostBuilder builder, string[] args)
+        /// <summary>
+        /// Use the default hosting configuration settings on the web host and allow for override by command line arguments.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="args">The command line arguments used to override default settings.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder UseDefaultHostingConfiguration(this IWebHostBuilder hostBuilder, string[] args)
         {
-            return builder.UseConfiguration(WebHostConfiguration.GetDefault(args));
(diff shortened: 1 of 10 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -12,36 +12,87 @@ public static class WebHostBuilderExtensions
...
+        /// <summary>
+        /// Set whether the startup error page should be used in the configuration settings of the web host.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="startupErrorPage"><c>true</c> to use startup error page; otherwise <c>false</c>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder UseStartupErrorPage(this IWebHostBuilder hostBuilder, bool startupErrorPage)
         {
-            return hostBuilder.UseSetting(WebHostDefaults.CaptureStartupErrorsKey, captureStartupError ? "true" : "false");
+            return hostBuilder.UseSetting(WebHostDefaults.CaptureStartupErrorsKey, startupErrorPage ? "true" : "false");
```
Tratcher: Should we also rename the key?

JunTaoLuo: Yup, oops.

davidfowl:
Now I think we should turn the error page off by default and remove the bool. We can update our templates to specify `UseStartupErrorPage()`.

Thoughts? /cc @danroth27 @Tratcher

danroth27:
@davidfowl How often would you expect people to **not** want the startup error page? Don't you pretty much always want it?

cwe1ss:
Azure Service Fabric is a common example were the startup error page is not helpful. Letting the application crash allows Service Fabric to retry a few times. (see #622 for further information)

davidfowl:
:shipit:

Tratcher:
And add `CaptureStartupErrors(true)` to UseIIS, correct?

JunTaoLuo:
@Tratcher like https://github.com/aspnet/IISIntegration/commit/81b9faaf58b66a983f279105008fcf278ddf5b9a?

cwe1ss:
That sounds reasonable. I guess almost everyone will want this when they host in IIS. 

Should this also be in the XML docs of UseIIS... to make sure, people are aware of it? The name only says "use the platform handler" - you might not expect something else to happen?!

Tratcher:
Yes, that works.

cwe1ss:
oh, I was talking about UseIISPlatformHandlerUrl(). I didn't know that UseIIS() exists. But having it in the docs there is fine anway :)

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -125,22 +125,22 @@ public void Build_honors_UseStartup_with_string()
         }
 
         [Fact]
-        public void DefaultConfigurationCapturesStartupErrors()
+        public void DefaultConfigurationDoesNotCapturesStartupErrors()
```
cwe1ss: Should be ...CaptureStartup... instead of ...CapturesStartup... right? Should it also be "DefaultHostingConfiguration..."?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostConfigurationsTests.cs
```diff
@@ -11,11 +11,11 @@ namespace Microsoft.AspNetCore.Hosting.Tests
     public class WebHostConfigurationTests
     {
         [Fact]
-        public void DefaultCapturesStartupErrors()
+        public void DefaultDoesNotCapturesStartupErrors()
```
cwe1ss: "CaptureStartup" instead of "CapturesStartup"?

Issue #658, opened 2016-03-17
Pass existing LoggerFactory to WebHostBuilder


cwe1ss:
Initializing the logging system should be one of the first things in an app. Now the `LoggerFactory` is created in the constructor of `WebHostBuilder` and it can be configured by calling `webHostBuilder.ConfigureLogging` or by configuring the logger in your `Startup` class.

However, there are scenarios where you want to initialize your logging system before creating a `WebHostBuilder`:
* your application hosts multiple WebHosts
* some stuff happens before you create the WebHost (e.g. registering with and getting the hosting url from Service Fabric)
* your app might be a GUI app that just happens to host a WebAPI for whatever reason, ...

For this reason, I suggest that it should be possible to optionally pass an existing `LoggerFactory` into the `WebHostBuilder`. This could be done by introducing a second constructor `WebHostBuilder(ILoggerFactory loggerFactory)` or by introducing another method `.UseLoggerFactory(ILoggerFactory loggerFactory)`. 

I would prefer the second constructor because this makes sure that there's always just one instance of the factory and you can't mix up the calls to `UseLoggerFactory` and `ConfigureLogging`.

If you agree to this, I'd be happy to create a pull request!

cwe1ss:
Of course, we could also just use the existing constructor with an optional parameter:

```csharp
public WebHostBuilder(ILoggerFactory loggerFactory = null)
{
     _hostingEnvironment = new HostingEnvironment();
     _loggerFactory = loggerFactory ?? new LoggerFactory();
     _configureServicesDelegates = new List<Action<IServiceCollection>>();
}
```

cwe1ss:
Closed to continue discussion directly in the pull request. I hope that's ok - I haven't figured out yet, whether it's better to have discussions in an issue or PR.

JunTaoLuo:
@cwe1ss Don't close the issue until you have merged the PR. We use issues to track progress. Discussions can happen in both places though it's more design oriented in issues and more implementation oriented in PRs

PR #661, opened 2016-03-19
Pass existing LoggerFactory into WebHostBuilder

cwe1ss:
As described in #658 

I'm motivated so I couldn't wait for your answer. :) Feel free to reject it if you think this is not correct or if you want something changed.

It allows to pass an existing ILoggerFactory into the WebHostBuilder constructor. I think this is better than a `UseLoggerFactory` method because the LoggerFactory is already created in the constructor right now.

I also added two tests. One checks if the host can actually resolve an ILoggerFactory and the second checks if the LoggerFactory passed in gets resolved correctly.
I added the tests at the top of the file because they are related to the constructor - I hope that's ok?!

BTW: Supporting a way to pass a LoggerFactor into the WebHostBuilder is also discussed here: 
https://github.com/aspnet/Logging/issues/346 ( @davidfowl @lodejard )

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
@@ -5,6 +5,7 @@ using System;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Hosting
 {
@@ -17,6 +18,13 @@ namespace Microsoft.AspNetCore.Hosting
         /// Builds an <see cref="IWebHost"/> which hosts a web application.
         /// </summary>
         IWebHost Build();
+
+        /// <summary>
+        /// Specify the <see cref="ILoggerFactory"/> to be used by the web host.
+        /// </summary>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> to be used.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder UseLoggerFactory(ILoggerFactory loggerFactory);
 
         /// <summary>
         /// Specify the <see cref="IServerFactory"/> to be used by the web host.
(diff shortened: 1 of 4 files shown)


cwe1ss:
Do you think that `ConfigureLogging()` should still be available if you accept this PR. Since you can now configure the logger factory before you create the WebHostBuilder, there's not really a reason to do call this method.

davidfowl:
I think we should support this but not via the constructor

cwe1ss:
Ok. How about this:

* `_loggerFactory` will no longer be initialized in the constructor.
* A new method `UseLoggerFactory(ILoggerFactory loggerFactory)` allows people to set `_loggerFactory`. This way it's similar to other methods like UseStartup etc.
* `BuildHostingServices` will only add `_loggerFactory` if it is not null. If no loggerFactory was configured, the call to `services.AddLogging()` will add a new LoggerFactory automatically:
```csharp
if (_loggerFactory != null) 
{
  services.AddSingleton(_loggerFactory);
}

services.AddLogging();
```

The existing method `ConfigureLogging(Action<ILoggerFactory> configureLogging)` is problematic because people could call it in the wrong order. I would remove it since the LoggerFactory can be configured before it is passed to UseLoggerFactory. 

If you want to keep it, we would have to do the following:
* `ConfigureLogging()` no longer directly executes the action. Instead it stores it in a new variable called `_configureLoggingDelegate`.
* `BuildHostingServices()` would now always create the LoggerFactory. It would look like this:
```csharp
var loggerFactory = _loggerFactory ?? new LoggerFactory();
if (_configureLoggingDelegate != null) 
{
  _configureLoggingDelegate(loggerFactory);
}

services.AddSingleton(_loggerFactory);
services.AddLogging();
```

If you disagree, what would you suggest?

cwe1ss:
I changed the code. It now has a `UseLoggerFactory` method. I also kept `ConfigureLogging` and made it similar to `ConfigureServices` to make sure that it still can be called multiple times.

Should it be possible to call `Build()` multiple times without side effects? Right now, I'm setting `_loggerFactory` in the Build method and I'm executing the delegates afterwards. If someone would call `Build` a second time, these delegates would be executed again on the same variable. The alternative would be to create a local LoggerFactory variable.

Tratcher:
Don't worry about Build being called more than once.

davidfowl:
While you're at it. add ConfigureLogging back to `IWebHostBuilder`.

cwe1ss:
I added `ConfigureLogging` and `UseLoggerFactory` to the interface. Note that this results in a new dependency on `Microsoft.Extensions.Logging.Abstractions` for the Abstractions package.

davidfowl:
This needs a rebase

cwe1ss:
hmm... I haven't done much merging/rebasing yet so I probably screwed it up. After trying a few things I ended up with duplicate commits. For that reason I did a complete rebase with `git rebase -i dev` and `git push -f` (because the push was denied otherwise). 
The commit history looks good now, however the PR still says all checks have failed. Probably because it didn't execute it again?!?

Is it ok like this or should I do something else?

JunTaoLuo:
merged in https://github.com/aspnet/Hosting/commit/4b16e83a1fa36ec03c417dbd8b47e4687dc19e89. Tested locally. Thanks @cwe1ss!

cwe1ss:
thank you very much!

PR #665, opened 2016-03-20
Use ConfigureDelegate instead of Action<IApplicationBuilder>

cwe1ss:
This fixes the open TODO `replace all Action<IApplicationBuilder> eventually with this` in ConfigureDelegate.cs. Since I'm trying to learn the codebase I thought I could give it a try.

* I moved the `ConfigureDelegate` to the Abstractions project because the signature was used there as well. I thought this was ok because it is similar to e.g. `RequestDelegate` from the HttpAbstractions package.
* I changed the namespace from `Microsoft.AspNetCore.Hosting.Startup` to `Microsoft.AspNetCore.Hosting` because that's the namespace of all other types in that package. This would be a small breaking change but it seems like `ConfigureDelegate` wasn't used so far. (I checked it on google with this query: "ConfigureDelegate site:github.com/aspnet")
* I did not rename the existing ConfigureDelegate.cs file to ConfigureBuilder.cs because there is a separate pull request for this.
* Some classes still need the namespace Microsoft.AspNetCore.Builder because some xml docs refer to it. I didn't want to update the comments because I'm not a native english speaker.

All in all, I'm not 100% sure that these delegate types are a good thing because you loose help from IntelliSense. `Action<IApplicationBuilder>` was very easy to understand and IntelliSense made clear how to use it and how many parameters there are. With `ConfigureDelegate` however (or other similar types like `RequestDelegate`), you might have to F12 into it to see the signature or do some trial & error.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/ConfigureDelegate.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/ConfigureDelegate.cs
@@ -1,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Builder;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    public delegate void ConfigureDelegate(IApplicationBuilder builder);
+}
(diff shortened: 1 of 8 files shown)


davidfowl:
All of the custom delegates should be gone actually.

davidfowl:
So the ConfigureDelegate should be removed.

cwe1ss:
Ok, that's good news. I'll close this one.

PR #668, opened 2016-03-21
Fix relative content root path test for non-windows systems

JunTaoLuo:
FYI @BrennanConroy @davidfowl 

Tested on my mac, still need to fix for net451 on windows.


JunTaoLuo:
Tested on Mac, Ubuntu, Windows, merging now.

PR #671, opened 2016-03-22
Remove the ConfigureDelegate delegate

cwe1ss:
@davidfowl said in #665 that the ConfigureDelegate should be removed. This would be a PR for it. 

[Seems like it is not used anywhere](https://www.google.at/#q=%22ConfigureDelegate%22+site:github.com%2Faspnet) - although that's probably not the best way to find out ;)


davidfowl:
:shipit:

JunTaoLuo:
merged in https://github.com/aspnet/Hosting/commit/bfcaa00e1b60bffd04aa866d19ec1d7f6da9f458

PR #673, opened 2016-03-23
Ensure an `ObjectPoolProvider` is registered

dougbu:
- e.g. take advantage of aspnet/HttpAbstractions#561 fix wherever cookies are used

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -17,6 +17,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.ObjectPool;
 using Microsoft.Extensions.PlatformAbstractions;
 
 namespace Microsoft.AspNetCore.Hosting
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -16,6 +16,7 @@
 using Microsoft.Extensions.PlatformAbstractions;
 using System.Reflection;
 using Xunit;
+using Microsoft.Extensions.ObjectPool;
```
Tratcher: sort

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -201,6 +202,9 @@ private IServiceCollection BuildHostingServices()
             // Conjure up a RequestServices
             services.AddTransient<IStartupFilter, AutoRequestServicesStartupFilter>();
 
+            // Ensure object pooling is available everywhere.
+            services.TryAddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
```
Tratcher: Try?

dougbu: Why _not_ allow users to override the provider? In any case, this method uses both Try and unconditional; I went with Try.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
dougbu: Ooops. (Likely wasn't placed correctly because the list was unsorted before.)

dougbu:
:up::date:

davidfowl:
:shipit:

dougbu:
b06a844

PR #677, opened 2016-03-25
Fix rare deadlock in TestHost

BrennanConroy:
@Tratcher @pakrym @muratg @davidfowl


JunTaoLuo:
Looks reasonable. Is it possible to write a test for this?

PR #681, opened 2016-03-30
React to HttpAbstractions namespace changes

dougbu:
- aspnet/HttpAbstractions#549 and aspnet/HttpAbstractions#592
- clean up `using`s


Tratcher:
:shipit: As this is non-breaking cleanup do not check it in until the original changes make it through coherence.

dougbu:
6e87b51

PR #682, opened 2016-03-30
Remove IServerLoader and add the IServerFactory to the DI container

davidfowl:
- This change removes the indirection and between the IServerLoader and
the IServerFactory. We now add the IServerFactory directly to the DI
container and resolve it from there.

/cc @lodejard @Tratcher @JunTaoLuo


src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -101,7 +104,46 @@ public static IWebHostBuilder UseServer(this IWebHostBuilder hostBuilder, string
...
+                .FirstOrDefault();
+
+            if (serverTypeInfo == null)
+            {
+                throw new InvalidOperationException($"No server type found that implements IServerFactory in assembly: {assemblyName}.");
+            }
+
+            return hostBuilder.ConfigureServices(services =>
+            {
+                services.AddSingleton(typeof(IServerFactory), serverTypeInfo.AsType());
```
Tratcher: Transient?

Tratcher: `UseServer<TServerFactory>() where TServerFactory is IServerFactory`?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -248,6 +222,12 @@ private IServiceCollection BuildHostingServices()
             services.AddSingleton(defaultPlatformServices.Application);
             services.AddSingleton(defaultPlatformServices.Runtime);
 
+            if (!string.IsNullOrEmpty(_options.ServerFactoryAssembly))
+            {
+                // Add the server factory
+                this.UseServer(_options.ServerFactoryAssembly);
```
Tratcher: It's a little strange to go back and call one of the extensions here. Revert the changes to `UseServer(string)` and just have a private `ResolveServerAssembly(string, IServiceCollection)`.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -237,99 +237,7 @@ public void ConfigureServices_CanBeCalledMultipleTimes()
             Assert.NotNull(host.Services.GetRequiredService<ServiceA>());
             Assert.NotNull(host.Services.GetRequiredService<ServiceB>());
         }
-
-        [Fact]
-        public void CodeBasedSettingsCodeBasedOverride()
-        {
-            var hostBuilder = new WebHostBuilder()
```
Tratcher: These were real config tests (not server tests). How about converting it to test a different field like CaptureStartupErrors?

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
davidfowl: Can't be transient. See the other place where we register this. Unless we remove the instance overloads it's bad to use different lifetimes for the same interface.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
davidfowl: Hmm, will check that.

davidfowl:
:up::date:

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -101,7 +104,33 @@ public static IWebHostBuilder UseServer(this IWebHostBuilder hostBuilder, string
                 throw new ArgumentNullException(nameof(assemblyName));
             }
 
-            return hostBuilder.UseSetting(WebHostDefaults.ServerKey, assemblyName);
```
Tratcher: Not just going to revert this change and let it happen in Build?

Tratcher:
:shipit:

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
davidfowl: Ah, will do that.

Issue #680, opened 2016-03-29
RequestServicesFeature.get_RequestServices is too expensive


rynowak:
Sampling data from HelloWorldMvc (TechEmpowerish) https://github.com/aspnet/Performance/tree/dev/testapp/HelloWorldMvc

![image](https://cloud.githubusercontent.com/assets/1430011/14116527/fe9f91e0-f594-11e5-98d1-b480de470ffd.png)

After adjusting for GC, creating the request services is pretty expensive. Would it be possible to cache the `IServiceScopeFactory`?

davidfowl:
Wow, @rynowak can you paste the get_RequestServices subtree?

rynowak:
There's a bit of strangeness going on here

![image](https://cloud.githubusercontent.com/assets/1430011/14117250/afdceedc-f598-11e5-9b95-673c9ec7b903.png)

There's nothing really **in** the method that's showing up as hot
```
.method public hidebysig static 
	object GetRequiredService (
		class [System.ComponentModel]System.IServiceProvider provider,
		class [System.Runtime]System.Type serviceType
	) cil managed 
{
	.custom instance void [System.Runtime]System.Runtime.CompilerServices.ExtensionAttribute::.ctor() = (
		01 00 00 00
	)
	// Method begins at RVA 0x284c
	// Code size 69 (0x45)
	.maxstack 2
	.locals init (
		[0] class Microsoft.Extensions.DependencyInjection.ISupportRequiredService
	)

	IL_0000: ldarg.0
	IL_0001: brtrue.s IL_000e

	IL_0003: ldstr "provider"
	IL_0008: newobj instance void [System.Runtime]System.ArgumentNullException::.ctor(string)
	IL_000d: throw

	IL_000e: ldarg.1
	IL_000f: brtrue.s IL_001c

	IL_0011: ldstr "serviceType"
	IL_0016: newobj instance void [System.Runtime]System.ArgumentNullException::.ctor(string)
	IL_001b: throw

	IL_001c: ldarg.0
	IL_001d: isinst Microsoft.Extensions.DependencyInjection.ISupportRequiredService
	IL_0022: stloc.0
	IL_0023: ldloc.0
	IL_0024: brfalse.s IL_002e

	IL_0026: ldloc.0
	IL_0027: ldarg.1
	IL_0028: callvirt instance object Microsoft.Extensions.DependencyInjection.ISupportRequiredService::GetRequiredService(class [System.Runtime]System.Type)
[...]

PR #686, opened 2016-04-01
Allocate the IServiceScopeFactory lazily and reuse it

davidfowl:
- Since we resolve the IServiceProvider from the application container, it's should be safe to reuse it across requests without the extra lookup everytime.

#680

/cc @rynowak @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
@@ -5,26 +5,27 @@ using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class RequestServicesContainerMiddleware
     {
         private readonly RequestDelegate _next;
-        private readonly IServiceProvider _services;
+        private IServiceScopeFactory _scopeFactory;
 
-        public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider services)
+        public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFactory scopeFactory)
         {
             if (next == null)
             {
                 throw new ArgumentNullException(nameof(next));
             }
-            if (services == null)
+            if (scopeFactory == null)
             {
-                throw new ArgumentNullException(nameof(services));
+                throw new ArgumentNullException(nameof(scopeFactory));
             }
 
-            _services = services;
+            _scopeFactory = scopeFactory;
             _next = next;
         }
 
@@ -44,7 +45,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 return;
             }
 
-            using (var feature = new RequestServicesFeature(_services))
+            using (var feature = new RequestServicesFeature(_scopeFactory))
             {
                 try
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -27,6 +30,18 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
             _services = services;
             _next = next;
         }
+        
+        public IServiceScopeFactory ScopeFactory
+        {
+            get
+            {
+                if (_scopeFactory == null)
```
davidfowl: Didn't bother locking, worst case scenario, we call into the DI container twice (we were doing it every request before).

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -27,6 +30,18 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider
...
+                {
+                    _scopeFactory = _services.GetRequiredService<IServiceScopeFactory>();
```
davidfowl: We should probably verify this is safe to do with 3rd party containers. It's supposed to be but I'm not sure if it's part of our spec tests.

/cc @pranavkm

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesFeature.cs
```diff
@@ -9,19 +9,19 @@ namespace Microsoft.AspNetCore.Hosting.Internal
...
     public class RequestServicesFeature : IServiceProvidersFeature, IDisposable
     {
-        private IServiceProvider _appServices;
+        private RequestServicesContainerMiddleware _middleware;
         private IServiceProvider _requestServices;
         private IServiceScope _scope;
         private bool _requestServicesSet;
 
-        public RequestServicesFeature(IServiceProvider applicationServices)
+        public RequestServicesFeature(RequestServicesContainerMiddleware middleware)
```
davidfowl: Thought about passing a `Func<IServiceScopeFactory>` here but we already have the middleware object so we allocate another :smile:

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -13,6 +14,8 @@ public class RequestServicesContainerMiddleware
         private readonly RequestDelegate _next;
         private readonly IServiceProvider _services;
 
+        private IServiceScopeFactory _scopeFactory;
+
         public RequestServicesContainerMiddleware(RequestDelegate next, IServiceProvider services)
```
rynowak: Why not just pass it in here? It's eager sure, but it's just one resolve. You get to drop the rest of the complexity and just pass it in a parameter to the feature.

davidfowl: Don't want to pay the cost. We'd have to measure it to see how it affects scenarios where you didn't ask for any services. If it was negligible then sure.

davidfowl: Lets try it.

davidfowl:
:up: :date:

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
pranavkm: https://github.com/aspnet/DependencyInjection/issues/397

rynowak:
:shipit: - this fixes teh problem

**Before**
image


**After**
image

davidfowl:
Wow that difference is kinda crazy

PR #687, opened 2016-04-02
Eagerly initialize hosting environment

javiercn:
Eagerly initialize hosting environment to prevent an issue with IHostingEnvironment.ApplicationName being null when calling ConfigureServices with a lambda expression during startup.


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -162,12 +162,6 @@ public IWebHost Build()
             var appEnvironment = hostingContainer.GetRequiredService<IApplicationEnvironment>();
```
davidfowl: remove this?

davidfowl:
:shipit:

PR #691, opened 2016-04-06
React to moving web.config

JunTaoLuo:
We need to update the IISExpress deployer to not  after moving `web.config` to the application root so that our sample apps can be tested.

cc @muratg @Eilon @DamianEdwards @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
@@ -79,7 +79,7 @@ namespace Microsoft.AspNetCore.Server.Testing
             DeploymentParameters.ApplicationPath =
                 (DeploymentParameters.ServerType == ServerType.IISExpress ||
                  DeploymentParameters.ServerType == ServerType.IIS) ?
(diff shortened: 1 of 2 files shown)


muratg:
:shipit:

src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -79,7 +79,7 @@ protected void DotnetPublish(string publishRoot = null)
             DeploymentParameters.ApplicationPath =
                 (DeploymentParameters.ServerType == ServerType.IISExpress ||
                  DeploymentParameters.ServerType == ServerType.IIS) ?
-                Path.Combine(DeploymentParameters.PublishedApplicationRootPath, "wwwroot") :
+                DeploymentParameters.PublishedApplicationRootPath :
```
moozzyk: nit: formatting is a bit off

JunTaoLuo: I think that was intentional. The line I modified lines up with the tabs. The line above it aligns with the boolean expression.

PR #692, opened 2016-04-07
Removed PublishApplicationBeforeDeployment from DeploymentParameters

kichalla:
@JunTaoLuo @Tratcher


Tratcher:
Why?

kichalla:
> Why?

Do we plan to support this in future? Currently we set to `true` everywhere this is used, no?

Tratcher:
It wasn't true by default for self-host, only IIS/Express.

kichalla:
:+1: Seems like in MusicStore its true for all types of server, but not in case of ServerTests.I will leave it as it is then (but I would prefer the publish approach as it seems more like a production scenario)

PR #696, opened 2016-04-12
Fix #695

### LINKED ISSUE #695: Startup failure with rc2-20466 and dotnet rc2-002339
_From @nil4 on April 10, 2016 22:23_

When an exception is thrown in the constructor of a service, and that service is injected in the `Configure` method, the application crashes on startup, but does not include any trace of the actual exception.

I encountered the issue in a larger project, and reduced the repro [to this gist](https://gist.github.com/nil4/dfb742b35fe1e7b87de80a62be773371). Versions of Windows, `dotnet`, `dnvm` and the content of `project.lock.json` are also included there. Here is the code:

```csharp
sealed class Startup {
  public static void Main(string[] args) => new WebHostBuilder().CaptureStartupErrors(false)
    .UseKestrel().UseStartup<Startup>().Build().Run();

  public void ConfigureServices(IServiceCollection services) => services.AddSingleton<Crasher>();

  public void Configure(Crasher crasher) {}
}

sealed class Crasher {
  public Crasher() { throw new Exception("Crash in the constructor"); }
}
```

Running the repro application with `-f net461` prints the following to the console before the Windows error dialog *(Repro.exe has stopped working)* shows up:

```cmd
> dotnet run -f net461
Unhandled Exception: System.Exception: Could not resolve a service of type 'Repro.Crasher' for the parameter 'crasher' of method 'Configure' on type 'Repro.Startup'.
   at Microsoft.AspNetCore.Hosting.Startup.ConfigureBuilder.Invoke(Object instance, IApplicationBuilder builder)
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()
   at Repro.Startup.Main(String[] args)
```

I expected the actual exception message *(Crash in the constructor)* and source (`Repro.Crasher`) to be shown. In the real application, it took a long time to figure out the cause of the crash.

Changing to `CaptureStartupErrors(true)` prints the same error message, but the application does not crash; it starts listening on http://localhost:5000. However, the startup error page also does not include any details of the actual exception; the only thing it shows is:

```
Oops. 500 Internal Server Error
An error occurred while starting the application.
```

Trying to run the same repro with `-f netcoreapp1.0` fails with `Could not resolve coreclr path`. I know things are still in flux with the new TFM, so I thought I should mention this also, just in case I missed something obvious in my `project.json`.

_Copied from original issue: aspnet/Home#1384_

nil4:
Capture the original exception in `ConfigureBuilder.Invoke`

When resolving services from the service provider, exceptions thrown from the service constructor were lost. Now they are included, addressing #695.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Startup/ConfigureBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Startup/ConfigureBuilder.cs
@@ -37,14 +37,14 @@ namespace Microsoft.AspNetCore.Hosting.Startup
                     {
                         parameters[index] = serviceProvider.GetRequiredService(parameterInfo.ParameterType);
                     }
-                    catch (Exception)
+                    catch (Exception ex)
                     {
                         throw new Exception(string.Format(
                             "Could not resolve a service of type '{0}' for the parameter '{1}' of method '{2}' on type '{3}'.",
                             parameterInfo.ParameterType.FullName,
                             parameterInfo.Name,
                             MethodInfo.Name,
-                            MethodInfo.DeclaringType.FullName));
+                            MethodInfo.DeclaringType.FullName), ex);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


davidfowl:
@nil4 Add a unit test? Verifying the inner exception is set?

nil4:
@davidfowl I added a test, care to have another look?

Tratcher:
:shipit: @JunTaoLuo

JunTaoLuo:
Merged in https://github.com/aspnet/Hosting/commit/eb8c0a263400e467da417fd2fc7e4039c6b5427c. Thanks @nil4!

Issue #698, opened 2016-04-13
Remove IServerFactory from Hosting.Server.Abstractions


davidfowl:
- Remove `IServerFactory` and instead resolve the `IServer` instance directly from the container
- Set the `IServerAddressesFeature` in Hosting prepopulated from the hosting specific configuration.
- Remove `UseServer` overloads

/cc @Tratcher @DamianEdwards

PR #702, opened 2016-04-14
Remove IServerFactory

Tratcher:
#698 @davidfowl @DamianEdwards @JunTaoLuo @muratg @mikeharder 

Hosting now configures the addresses via IServerAddressesFeature, just like it was doing with defaults before.

One oddity: The server gets disposed twice. Once to stop the server, and once when we dispose the container. This is harmless.


Tratcher:
Updated.

mikeharder:
:shipit: from me, you may want to check with others.

PR #706, opened 2016-04-15
Make contentRootPath error give failed path

### LINKED ISSUE #707: Include the content root path as part of exception message
UseContentRoot throws an `ArgumentException` when the directory does not exist. We should modify the error message to include the actual path. 

https://github.com/aspnet/Hosting/blob/b7bdc9c40494f8e0f0eac22db91b8d5c58811ee2/src/Microsoft.AspNetCore.Hosting/Internal/HostingEnvironmentExtensions.cs#L28 line to change.

ryanbrandenburg:
Include the content root that doesn't exist in the error message for a missing content root to make debugging easier.

Fixes #707.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingEnvironmentExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingEnvironmentExtensions.cs
@@ -25,7 +25,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             }
             if (!Directory.Exists(contentRootPath))
             {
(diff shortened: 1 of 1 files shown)


Tratcher:
:shipit:

Tratcher:
https://github.com/aspnet/Hosting/issues/707

ryanbrandenburg:
Ha, I didn't even see that bug @Tratcher, I just got annoyed at it.

Tratcher:
Check the timestamps. you fixed it before there was a bug.

pranavkm:
Can we fix this in release? Seems like a poor user experience not having this parameter

ryanbrandenburg:
I'll ask @Eilon.

PR #709, opened 2016-04-17
Remove diagnostic messages from IStartupLoader API

davidfowl:
- This is left over from katana where the hosting layer had more of a pipeline. It's not required of used right now.

/cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -129,23 +129,17 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
             if (StartupType == null)
             {
-                var diagnosticTypeMessages = new List<string>();
-                StartupType = _startupLoader.FindStartupType(StartupAssemblyName, diagnosticTypeMessages);
+                StartupType = _startupLoader.FindStartupType(StartupAssemblyName);
                 if (StartupType == null)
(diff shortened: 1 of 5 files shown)


Tratcher:
:shipit:

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -129,23 +129,17 @@ private void EnsureStartup()
...
             {
-                var diagnosticTypeMessages = new List<string>();
-                StartupType = _startupLoader.FindStartupType(StartupAssemblyName, diagnosticTypeMessages);
+                StartupType = _startupLoader.FindStartupType(StartupAssemblyName);
                 if (StartupType == null)
                 {
-                    throw new ArgumentException(
-                        diagnosticTypeMessages.Aggregate("Failed to find a startup type for the web application.", (a, b) => a + "\r\n" + b),
-                        StartupAssemblyName);
+                    throw new ArgumentException("Failed to find a startup type for the web application.", StartupAssemblyName);
```
moozzyk: Shouldn't this be nameof(StartupAssemblyName)? I am pretty sure the second parameter is the name of the parameter causing the issue. Currently you will get something like "Parameter Name: MyAssembly"
(This also tells me that there are no tests for this code path)

davidfowl:
Superseded by #714

Issue #712, opened 2016-04-18
Startup fails when static method is passed to WebHostBuilder.Configure


nil4:
```csharp
using System;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;

static class Program {
    static void Main(string[] args) => new WebHostBuilder()
        .UseKestrel()
        .Configure(ConfigureApplication)
        .Build().Run();

    static void ConfigureApplication(IApplicationBuilder app) {}
}
```

Expected: the application starts up and `ConfigureApplication` is called.

Actual: the application crashes on startup with stack trace:

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Object.GetType()
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.ResolveApplicationName()
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.BuildHostingServices()
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.Build()
   at Program.Main(String[] args) in D:\projects\Repro\Program.cs:line 6
```

In [`WebHostBuilder.ResolveApplicationName`](https://github.com/aspnet/Hosting/blob/4e5fd908fd5eb33f55cf90f852f3328468e64e2a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L257), `_startup.ConfigureDelegate.Target` is assumed to be non-null. This assumption is invalid for static methods.

nil4:
Pull request #713 resolves this issue by using `_startup.ConfigureDelegate.GetMethodInfo().DeclaringType` instead.

khellang:
Doesn't #714 get rind of this code completely?

davidfowl:
Nope, we still determine the application name https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs#L82

nil4:
PR #716 resolves this issue. /cc @davidfowl @Tratcher

EricABC:
[...]

PR #713, opened 2016-04-18
Resolve aspnet/Hosting#712

nil4:
[`WebHostBuilder.ResolveApplicationName`](https://github.com/aspnet/Hosting/blob/4e5fd908fd5eb33f55cf90f852f3328468e64e2a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L257) assumes `_startup.ConfigureDelegate.Target` is non-null. This assumption is invalid for static methods, resulting in applications crashing on startup. 

Instead, use `_startup.ConfigureDelegate.Method.DeclaringType` which is non-null for both static and instance methods.


davidfowl:
Needs a test

davidfowl:
@nil4 can you rebase on dev and add a unit test?

PR #714, opened 2016-04-18
Introduce IStartup interface

davidfowl:
This change introduces a new service `IStartup` that can be registered in the hosting container to override any startup logic. `UseStartup` overloads have been changed to detect `IStartup` and directly put it in the container, or to wrapping it with a `ConventionBasedStartup` implementation to preserve the existing behavior.

- Remove IStartupLoader and add `IStartup` instead that matches the signature hosting cares about
- Moved `UseStartup` to extension methods
- Move existing logic into `ConventionBasedStartup` class

/cc @Tratcher @JunTaoLuo


src/Microsoft.AspNetCore.Hosting/Startup/Startup.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    public abstract class Startup : IStartup
```
Tratcher: Having `Startup` as the base name is going to be an obnoxious conflict with app's `Startup`.

Tratcher: `StartupBase`?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -225,8 +180,50 @@ private IServiceCollection BuildHostingServices()
...
+                        return null;
+                    });
+                }
+            }
+
+            if (!string.IsNullOrEmpty(_options.StartupAssembly))
+            {
+                try
+                {
+                    var startupType = StartupLoader.FindStartupType(_options.StartupAssembly, _hostingEnvironment.EnvironmentName);
```
Tratcher: `ResolveServerType` vs `FindStartupType`?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -225,8 +180,50 @@ private IServiceCollection BuildHostingServices()
...
+                    
+                    if (typeof(IStartup).GetTypeInfo().IsAssignableFrom(startupType.GetTypeInfo()))
+                    {
+                        services.AddSingleton(typeof(IStartup), startupType);
```
Tratcher: Hmm, this keeps the startup instance alive for the duration of the app. It also would call Dispose at the end if implemented.

Tratcher: `services.AddSingleton(typeof(IStartup), startupType);` vs `services.AddSingleton<IStartup>(startupType);`?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -225,8 +180,50 @@ private IServiceCollection BuildHostingServices()
...
+                            return new ConventionBasedStartup(StartupLoader.LoadMethods(sp, startupType, hostingEnvironment.EnvironmentName));
+                        });
+                    }
+                }
+                catch (Exception ex)
+                {
+                    var capture = ExceptionDispatchInfo.Capture(ex);
+                    services.AddSingleton<IStartup>(_ =>
+                    {
+                        capture.Throw();
```
Tratcher: Why this pattern?

Tratcher:
How about these?
https://github.com/aspnet/Hosting/issues/712
https://github.com/aspnet/Hosting/issues/333

Tratcher:
The code is simpler in a few places, but you've added multiple concepts that we now have to explain to people. E.g. The interface, base class, wrapper classes, incompatible Configuration variations, etc. All told I think this is worse for consumers because it will be much more confusing. We need to choose one pattern or the other.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl:
```
services.AddSingleton<IStartup>(startupType); 
```

That doesn't work. That adds an instance of the Type object.

davidfowl:
> The code is simpler in a few places, but you've added multiple concepts that we now have to explain to people. E.g. The interface, base class, wrapper classes, incompatible Configuration variations, etc. All told I think this is worse for consumers because it will be much more confusing. We need to choose one pattern or the other.

That's inaccurate, the main change is simplifying the hosting interface between the hosting and the application layer. The new contract is `IStartup`. Now we can talk about preserving the behavior of the `Startup` class we have today and determine if we want to keep that separately.

The new classes are:

- `IStartup` - Contract for calling the application bootstrapper
- `StartupBase` - Convenience base class for end users
- `ConventionBasedStartup` - Logic we have today
- `DelegateStartup` - Direct delegate to invoke the configure method 

I'm going to commit this because it cleans up the hosting code very nicely and doesn't break the existing mechanisms around the Startup pattern. We can figure out how to change the `ConventionBasedStartup` to do more things like #712 #333 if we choose to later.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: So the Startup error page works. Ideally though, we would just remove this capability completely.

src/Microsoft.AspNetCore.Hosting/Startup/Startup.cs
davidfowl: I like that better

davidfowl:
@Tratcher Also:

https://github.com/aspnet/Hosting/issues/333 - We get this for free now, you can call `AddSingleton<IStartup>(new MyStartup());`
https://github.com/aspnet/Hosting/issues/712 - Just a bug, needs to be rebased.

davidfowl:
TBH, we could get rid of the base class and force user code to implement the interface since we already have the convention based startup.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: Kept the same name but I don't have a strong opinion here.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: > Hmm, this keeps the startup instance alive for the duration of the app. It also would call Dispose at the end if implemented.

Seems fine.

divega:
@davidfowl do we get real public (i.e. Internal namespace doesn't count) to resolve the `IStartup` of the app? A well defined contract should help but w shouldn't be in the business of repliacting this logic in design time tools like EF migrations and scaffolding.

davidfowl:
@divega If you use the hosting logic then you have everything in the container, otherwise it's no different than what you have today. Today we still support all of the injection etc via ConventionBasedStartup. It's another discussion if we want to stick to the interface only. 

Also, injection is still supported in the ctor of the `IStartup` so you still need to deal with that.

PR #716, opened 2016-04-19
Fixes #712

### LINKED ISSUE #712: Startup fails when static method is passed to WebHostBuilder.Configure

nil4:
Rebased per https://github.com/aspnet/Hosting/pull/713#issuecomment-211726427. /cc @davidfowl @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
@@ -79,7 +79,7 @@ namespace Microsoft.AspNetCore.Hosting
                 throw new ArgumentNullException(nameof(configureApp));
             }
             
-            var startupAssemblyName = configureApp.Target.GetType().GetTypeInfo().Assembly.GetName().Name;
+            var startupAssemblyName = configureApp.GetMethodInfo().DeclaringType.GetTypeInfo().Assembly.GetName().Name;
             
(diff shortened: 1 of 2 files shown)


muratg:
LGTM

davidfowl:
@nil4 We're going to make this change part of RC2 as well.

Issue #718, opened 2016-04-20
Update deployers for shared runtime


JunTaoLuo:
Currently our deployers are not handling portable apps correctly with the logic in https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs#L53-L72. We need to run `dotnet *.dll` instead. This is required for testing converted MusicStore, ServerTests, and IISIntegration and Entropy.

PR #719, opened 2016-04-20
Updating deployers for apps using shared runtime

JunTaoLuo:
#718

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
@@ -74,6 +74,8 @@ namespace Microsoft.AspNetCore.Server.Testing
         /// </summary>
         public bool PublishApplicationBeforeDeployment { get; set; }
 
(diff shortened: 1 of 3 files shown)


JunTaoLuo:
Oops I mean to merge this into release not dev

src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
```diff
@@ -74,6 +74,8 @@ public class DeploymentParameters
         /// </summary>
         public bool PublishApplicationBeforeDeployment { get; set; }
 
+        public bool UseSharedRuntime { get; set; }
```
Tratcher: Maybe an enum? It would better describe the else case.

Tratcher: Default to shared? That's what we changed all of our samples to, correct?

JunTaoLuo:
:up::date:

Tratcher:
:shipit:

PR #720, opened 2016-04-20
Change default environment to "Development"

cwe1ss:
I believe it's extremely dangerous to set "Production" as the default environment. It's just way to easy to screw this up on developer's or test machines! 

E.g. if you're using environment variables for detecting the environment, this basically means that whenever you are starting an app on a new machine, it will run against the production environment by default! This can't be good!

Instead, if we would change the default to "Development" this would mean that it always runs in a "safe" mode unless you actively configure a system to be the production system. (e.g. by setting an Environment variable). This effort should be ok for everyone because it is a one-time thing and a conscious decision to set something as the production environment.

It's also a lot easier to notice that a production system points to a development system than the opposite.

PS: I know this is temporary but just look at the current situation due to all the name-changes in this repository. If people forget to change this when they move to RC2, they will suddenly end up targeting their production environment.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingEnvironment.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingEnvironment.cs
@@ -7,7 +7,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class HostingEnvironment : IHostingEnvironment
     {
-        public string EnvironmentName { get; set; } = Hosting.EnvironmentName.Production;
+        public string EnvironmentName { get; set; } = Hosting.EnvironmentName.Development;
 
         public string ApplicationName { get; set; }
 


Tratcher:
This was an explicit design decision. @blowdart care to explain?

cwe1ss:
yes, please explain. this is extremely dangerous. You're not worried about this? How do you ensure people don't screw this up?

blowdart:
Basically we have always had a bunch of problems with people pushing "dev" configurations into production and ending up with information disclosure due to error messages, or miss-configured logging.

As part of our push to be secure by default it was a conscious decision to default to production. Developer mode is the more dangerous configuration, so you have to specifically configure it (or start from VS).

This won't change.

cwe1ss:
In terms of security settings, you are correct - in terms of connection strings and so on, it's the opposite IMO.

I would have liked to see more discussion about this. But thanks for taking the time to answer. I really hope you are right.

I just want to say that it happened at least 10 times to us already that people ended up pointing to production (which luckily doesn't exist yet) during our chase of RC2 daily builds.

blowdart:
Well we don't want you using config outside of dev mode for things like connection strings, that's what environment settings are for. That's why `builder.AddEnvironmentVariables();` comes after reading the config files. They take precedence.

cwe1ss:
Last comment (sorry): I'd much rather end up with miss-configured logging, disclosed exceptions, ... than accidentally having someone overwrite some production data. The first problem is a one-time fix, the latter is an always-present problem.

blowdart:
But if you move config to environment that won't happen anyway, and of course, you'd never give your live site credentials to developers would you? :)

cwe1ss:
Of course, we don't store plain connection strings in our files but we use Azure Key Vault for everything and we e.g. have the Azure AD client ID in the file

blowdart:
Also you should remember that launching from VS will [launch in development](https://docs.asp.net/en/latest/fundamentals/environments.html) by default.

cwe1ss:
I understand your point. But having all production variables in environment variables isn't great either. It's really hard to manage them and you would need process specific environment variables if you have more than one app on a server. There's also stuff like Azure Service Fabric which moves your services between servers all the time and so on.

Some more context about our current solution:
We use Azure Key Vault for *everything* that's sensitive. Every app is setup in Azure AD and gets a client ID. We use certificates instead of passwords for access to the key vault.
This allowed us to have all configs in our repository - we would just check-in the client-id and the thumbprint. The certificate will only be installed on the corresponding environments so developers should never have the production cert on their machine (however... you might have to debug against production one day so it's not completely safe to say that it will never end up on one). On deployment, our CI system makes sure that only the corresponding *.ENVIRONMENT.* files are copied to the system. 
So this system is pretty safe but I guess I still have to reconsider it and move the client ID and the thumbprint away anyway...

blowdart:
In all honesty you're way ahead of the majority of devs in your setup, and I'm seriously impressed. I take your point, but the change protected more developers than it harms, so the needs of the many outweighs you I'm afraid,

cwe1ss:
could we at least colorize the environment name on console startup and show "Production" with a red text? :-) 

https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs#L52

blowdart:
That I wouldn't object to, feel free to file a separate issue. It's a shame we can't do blinking :)

cwe1ss:
yeah - blinking would be great. I will research if that's possible somehow :D. thank you very much for your time!

khellang:
There are ANSI escape codes for blinking. https://en.m.wikipedia.org/wiki/ANSI_escape_code :wink: Question is how widely supported they are.

blowdart:
I'm going to cry underneath my desk now.

cwe1ss:
great - I will create a PR! :dancers: 

I would like to add 2 last things to this issue after having slept on it:

* Visual Studio may start with Development as the default, however if you use "dotnet run" it does not (at least not for me on the latest RC2 bits). This basically means, everyone who's not using Visual Studio will not have this behavior. There shouldn't be such a big difference just because you use a different editor IMO.

* There are also lots of non-security settings which you also wouldn't want to use from the wrong environment: URLs for external HttpClient calls, ... - basically everything that you would have had in your appSettings entry in web.config

I'm definitely on your side that security settings should default to a secure state, however since the `Configuration`-system will be used for security AND regular settings, I'm not convinced that this is a good decision for the majority of people.

But thanks for clarifying, I will change a few things on our side to make sure we don't get bitten by this.

cwe1ss:
fyi - I finally figured out why my applications suddenly pointed to Production after an upgrade to the latest RC2 bits. I already had a system-wide environment variable for ASPNET_ENV but not yet for the new ASPNETCORE_ENVIRONMENT.

Proofs my point about it being weird that Visual Studio does stuff differently than the regular dotnet cli.

blowdart:
Ah. Darned RC changes. Yea, until tooling arrives for RC2 VS is going to be very out of sync.

cwe1ss:
I found an easy fix for this whole dilemma - we will just not use the word "Production" for our production system! Instead we will use e.g. "Live". 

Funny that I haven't thought of this in the first place!

There isn't any performance optimization or so that only happens if the environment is "Production", is there?

blowdart:
Not that I know of. That's pretty sneaky :)

Issue #722, opened 2016-04-21
DNX branding shows up in the ErrorPage


muratg:
_From @rustd on April 21, 2016 19:21_

**Repro Steps:**

- Create an ASP.NET Core Web App
- Introduce a compilation error
- Run the app to see the error page

**Expected:**
DNX branding should not show up in the error page. It does at the bottom as shown in this screenshot

![errorpage](https://cloud.githubusercontent.com/assets/1132652/14721455/4cbca222-07bb-11e6-85a7-5fd8fd891cc9.PNG)

_Copied from original issue: aspnet/Diagnostics#271_

muratg:
_From @Eilon on April 21, 2016 21:11_

Hmm I don't see this in the latest packages. I ran from source code and there's no footer at all. I also don't see this being generated by searching the repo. I also looked over the last month or two of changes and didn't see where the might have been removed.

@Tratcher @ryanbrandenburg do you guys recall removing this? Or is it coming from somewhere else?

muratg:
_From @Tratcher on April 21, 2016 21:24_

That's Hosting's startup exception page, not the developer error page:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Startup/StartupExceptionPage.cs
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/compiler/resources/GenericError_Footer.html

muratg:
_From @rustd on April 21, 2016 21:36_

I am seeing this with this version  1.0.0-rc2-20581

PR #726, opened 2016-04-25
Remove DNX branding from footer

JunTaoLuo:
#722 I also did a quick search for other occurrences of the word dnx. Still need to investigate the use of https://github.com/aspnet/Hosting/blob/12fff1833fdcaa36b6d43bbdeb0465a11d9cf365/src/Microsoft.AspNetCore.Server.Testing/Deployers/IISExpressDeployer.cs#L88-L95.

The new error footer looks like:
image

Review: @Tratcher 
Approval: @Eilon @DamianEdwards

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Startup/StartupExceptionPage.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Startup/StartupExceptionPage.cs
@@ -234,6 +234,7 @@ namespace Microsoft.AspNetCore.Hosting.Startup
         {
             var environment = PlatformServices.Default.Runtime;
             var runtimeType = HtmlEncodeAndReplaceLineBreaks(environment.RuntimeType);
+            var runtimeDisplayName = runtimeType == "CoreCLR" ? ".NET Core" : runtimeType == "CLR" ? ".NET Framework" : "Mono";
 #if NETCOREAPP1_0 || NETSTANDARD1_3
             var systemRuntimeAssembly = typeof(System.ComponentModel.DefaultValueAttribute).GetTypeInfo().Assembly;
             var assemblyVersion = new AssemblyName(systemRuntimeAssembly.FullName).Version.ToString();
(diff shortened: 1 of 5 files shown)


davidfowl:
The version is wrong but it's the "rightest" thing we can do right now 😄 

:shipit: !

DamianEdwards:
Shouldn't be ".NET Framework CoreCLR", that's combining two things that shouldn't be. It's either ".NET Framework version 4.*" or ".NET Core version *.*".

Also, the "x64 Runtime version 4.0.0.0" is confusing. What is that referring to exactly? We don't have DNX anymore so I don't think it's even needed. Move the bitness string into the first string and remove that section completely.

For the OS string, please include the bitness there too, as it may be different to the runtime bitness.

JunTaoLuo:
@DamianEdwards Just want to clarify a few things here. 

1. ".NET Framework" and ".NET Core" makes sense but what about Mono? Since we differentiate it in `PlatformAbstractions`, we originally had ".NET Framework Mono" but maybe we can drop the "Mono" part of it. 
2. The runtime version is what we get from `PlatformAbstractions`. There it uses the version of the assembly containing the type `object` as the "Runtime" version. I'm guessing this is what was meant by it's wrong but it's the rightest thing for now.
3. I don't think hosting is the right place to figure out the OS bitness. `PlatformAbstractions` feels like the more logical place to have this logic. There is a work item to update to using CoreFX's `RuntimeInformation` which will have this information but that's for RTM. Unless we are going to pull that back to rc2, I'll look into some other options.

DamianEdwards:
Let's just fix the first one for RC2 then. Happy to keep "Mono" where applicable, so ".NET Framework", ".NET Core", and ".NET Framework (Mono)"?

JunTaoLuo:
@DamianEdwards I also removed the runtime section as well and moved the bitness string to the clr type section. The updated footer looks like 
image
image
 
Do you like this better?

muratg:
I think that's fine, only part that's a little weird is ".NET Framework x64 version 4.0....".  @DamianEdwards thoughts?

src/Microsoft.AspNetCore.Hosting/Startup/StartupExceptionPage.cs
```diff
@@ -234,6 +234,7 @@ private static string GenerateFooterEncoded()
         {
             var environment = PlatformServices.Default.Runtime;
             var runtimeType = HtmlEncodeAndReplaceLineBreaks(environment.RuntimeType);
+            var runtimeDisplayName = runtimeType == "CoreCLR" ? ".NET Core" : runtimeType == "CLR" ? ".NET Framework" : ".NET Framework Mono";
```
davidfowl: Change .NET Framework Mono -> Mono?

JunTaoLuo:
:up::date: as per offline discussion with @DamianEdwards 

The updated footer looks like:
new clr exception
new coreclr exception
 Will create a followup issue to convert the full CLR version to something more reasonable like 4.5.1, 4.6...

PR #729, opened 2016-04-26
[Fixes #721]Test server should fire onResponseCompleted after the actual response has completed

### LINKED ISSUE #721: Test server should fire onResponseCompleted after the actual response has completed instead of just the headers
Currently TestServer fires OnResponseCompleted here:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs#L202
At this point, the response stream might have not been completely written to the response which might cause the response get cut in multiple and bizarre ways.

Instead, the event needs to be triggered after the response has been completed here:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs#L195

We should also add more logging to the different pieces on TestHost to ensure that its easy to troubleshoot these types of issues if necessary, one such way is to allow providing a logger that can be used to trace important events like the response finishing, etc

javiercn:


Tratcher:
:shipit:

davidfowl:
@javiercn this was the bug? Nice!

javiercn:
Checked-in to release and merged into dev as this was approved for RC2

Issue #727, opened 2016-04-26
RC2 Templates Review follow-up: Remove UseDefaultHostingConfiguration


rustd:
From the templates review, it was brought up that the following line of code (https://github.com/aspnet/Templates/blob/dev/src/BaseTemplates/StarterWeb/Program.cs#L18) should be removed from the templates. Please make the corresponding FX changes.
```c#
               // ToDo: Open a bug
                .UseDefaultHostingConfiguration(args)
```
@muratg FYI

muratg:
@JunTaoLuo This is from the templates review today. Please have a look, and send a mail to @Eilon and @DamianEdwards for approval.

cc @Tratcher

muratg:
https://github.com/aspnet/Hosting/issues/700

khellang:
@muratg During the template review, did you discuss IStartup/StartupBase/ConventionBasedStartup? :smile:

davidfowl:
We didn't but I don't see us using that interface in the templates.

khellang:
So that means ConventionBasedStartup implicitly, like today?

DamianEdwards:
It did come up as I thought we were making that change (to use `IStartup`). For now it appears we're staying with the convention based approach in the templates by default.

khellang:
Alright. As long as I can opt-out (of the conventions), I'm all good :smile:

JunTaoLuo:
Need to track the updates of our samples.

Issue #700, opened 2016-04-14
Add ASPNETCORE_* environment variables by default to hosting configuration


davidfowl:
Today you have to call `UseDefaultConfiguration` to turn on the ASPNETCORE_ environment variables. This means that `ASPNETCORE_` settings aren't applied unless the application code is setup to do so. While this seems like a pure choice, it makes it hard to diagnose why things aren't working. An example of this is the IISIntegration middleware. If you don't call UseDefaultConfiguration then it won't bind to IIS. 

Instead, we should just setup the the env vars by default.

/cc @DamianEdwards @Tratcher

Tratcher:
How do you turn it off?

LeandroPT:
Hello all... 

This should be reflected in all examples with upper cases :), specially hosting in IIS...

I have been banging my head with this.. Thanks for the post :+1:

muratg:
@JunTaoLuo We think this is required for #727. 

cc @Tratcher

Issue #693, opened 2016-04-08
Need a way to inform user of Hosting-supported command-line arguments`


dougbu:
Command-line arguments such as `--contentRoot {root}` are `<sarcasm>`:100:% discoverable`</sarcasm>` -- if you've received the right email from @NTaylorMullen or are reading the code. Users are not even informed that all supported options are name / value pairs, let alone that they map to configuration options. Then, there's nothing available to list the configuration options that Hosting (in particular) supports.

Part of this may be an `aspnet/Configuration` bug. But the end-to-end experience isn't great either way.

/cc @danroth27 @glennc

muratg:
@dougbu Were you thinking documentation or cmd line help?

dougbu:
Either but command-line help would likely be more discoverable.

Tratcher:
Command line help is problematic as hosting does not own Program.Main, the app does.

dougbu:
@Tratcher command-line help would depend on a bit o' magic in `UseDefaultHostingConfiguration()`.

Tratcher:
UseDefaultHostingConfiguration doesn't control the flow of the application. It can't terminate the app if there's a "--help" in the config, it doesn't control when Main exists.

davidfowl:
Maybe we should just remove support for args and support the config file and env variables by default.

Tratcher:
How does that help?  Note we need the environment variable support for IIS.

dougbu:
> Maybe we should just remove support for args and support the config file and env variables by default.

And then document the configuration options Hosting supports for all ASP.NET Core web applications? How would you make that information discoverable?

davidfowl:
> And then document the configuration options Hosting supports for all ASP.NET Core web applications? How would you make that information discoverable?

Docs, like the other things of the same nature.

> How does that help? Note we need the environment variable support for IIS.

Not sure what you mean. The entire issue is about obscure command line arguments. We don't own the program so showing help is kinda odd. Env variables and config would still be supported...

Tratcher:
Oh, nevermind, I read that as "Maybe we should just remove support for the config file and env variables by default."

Args are useful during development and maybe even in production. Removing them does not address the undiscovered nature of configuring Hosting, it just makes configuring Hosting even more difficult.

davidfowl:
I think the big difference between args and the other things it that it collides with the real application's args. That's why it's messy and sure you can say it doesn't address the problem but it makes it more difficult to reason about. 

[...]

PR #730, opened 2016-04-27
Removing DefaultHostingConfiguration

JunTaoLuo:
This issue addresses #727, #700 and render #693 not applicable

As discussed offline with @davidfowl @Tratcher , we are removing default hosting configurations and the user is expected to construct their own configurations. The only exception is that we fallback to environment variables for the `Environment` option. I'll make an update to IISIntegration in a later PR to see what the changes there look like.

We will also need to update all of our samples again.


samples/SampleStartups/StartupHelloWorld.cs
```diff
@@ -27,8 +28,10 @@ public void Configure(IApplicationBuilder app)
         // Entry point for the application.
         public static void Main(string[] args)
         {
+            var config = new ConfigurationBuilder().AddCommandLine(args).Build();
```
davidfowl: Remove config from this sample.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -353,6 +352,41 @@ public void UseEnvironmentIsNotOverriden()
...
+            // Default to Production
+            var host = new WebHostBuilder()
+                .UseServer(testServer)
+                .UseStartup("Microsoft.AspNetCore.Hosting.Tests")
+                .Build();
+
+            Assert.Equal(EnvironmentName.Production, host.Services.GetService<IHostingEnvironment>().EnvironmentName);
+
+            // Use environment variables if available
+            Environment.SetEnvironmentVariable(WebHostDefaults.EnvironmentVariablesPrefix + WebHostDefaults.EnvironmentKey, "EnvironmentA");
```
davidfowl: Scary, let's *not* have a test for this. Global state like this affects tests running in parallel.

JunTaoLuo: Maybe we can launch the test in a subprocess with environment variables set so it doesn't affect the global state of the current process? I can also just remove it if that's too much trouble.

Tratcher: The test deployers already set the environment this way, correct? That should be adequate coverage.

Tratcher:
Side note: The IISIntegration changes need to be checked in first, and they don't rely on any of these changes.

samples/SampleStartups/StartupFullControl.cs
```diff
@@ -13,7 +14,14 @@ public class StartupFullControl
     {
         public static void Main(string[] args)
         {
+            var config = new ConfigurationBuilder()
+                .AddCommandLine(args)
+                .AddEnvironmentVariables(prefix: WebHostDefaults.EnvironmentVariablesPrefix)
+                .AddJsonFile(WebHostDefaults.HostingJsonFile, optional: true)
```
Tratcher: Remove the WebHostDefaults.HostingJsonFile constant, it's no longer used in src.

I'm tempted to remove WebHostDefaults.EnvironmentVariablesPrefix as well, as it's only used once in src.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
JunTaoLuo: They do, though they will only fail if the deployed test tests something that's environment specific and the failure can be hard to trace back to this issue? Also, I don't think anything tests the hierarchy of the fallbacks, but I assume that it's not deemed important.

Tratcher: They use the environment to select different Startups for different tests, so yes, they will fail if it's wrong.

JunTaoLuo: Fair.

samples/SampleStartups/StartupHelloWorld.cs
```diff
@@ -28,7 +28,6 @@ public void Configure(IApplicationBuilder app)
         public static void Main(string[] args)
         {
             var host = new WebHostBuilder()
-                .UseDefaultHostingConfiguration(args)
                 .UseStartup<StartupHelloWorld>()
```
Tratcher: UseServer?

Tratcher:
:shipit:

samples/SampleStartups/StartupHelloWorld.cs
JunTaoLuo: good catch, here and in `StartupExternallyControlled`.

muratg:
:shipit:

PR #735, opened 2016-04-28
Pre-populate the environment rather than fall back to it.

Tratcher:
@davidfowl This is followup to https://github.com/aspnet/Hosting/pull/730/files#diff-b41ef5dfd9823ae114dd7c9ea12d1676.

When updating the downstream projects we found it's better to pre-populate the environment so we know what it is while configuring WebHostBuilder rather than afterwards.
https://github.com/aspnet/MusicStore/blob/cf35795db6449949d5a297e15a9cc649df5c9266/src/MusicStore.Standalone/Program.cs#L24-L27

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -47,7 +47,19 @@ namespace Microsoft.AspNetCore.Hosting
             _configureLoggingDelegates = new List<Action<ILoggerFactory>>();
 
             // This may end up storing null, but that's indistinguishable from not adding it.
-            UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"));
+            UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")
+                // Legacy keys, never remove these.
+                ?? Environment.GetEnvironmentVariable("Hosting:Environment")
+                ?? Environment.GetEnvironmentVariable("ASPNET_ENV"));
+
+            if (Environment.GetEnvironmentVariable("Hosting:Environment") != null)
+            {
+                Console.WriteLine("The environment variable 'Hosting:Environment' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
+            }
+            if (Environment.GetEnvironmentVariable("ASPNET_ENV") != null)
+            {
+                Console.WriteLine("The environment variable 'ASPNET_ENV' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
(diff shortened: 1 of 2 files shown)


Tratcher:
Added legacy fallbacks (#738).
@blowdart @DamianEdwards @Eilon

Tratcher:
Unfortunately no logging is available this early in the application, so it's difficult to notify users that they're using outdated keys.

davidfowl:
:shipit:

Tratcher:
What if I add an obnoxious Console.WriteLine if I detect legacy environment variables?
`The environment variable 'ASPNET_ENV' is obsolete and has been replaced by 'ASPNETCORE_ENVIRONMENT'`

davidfowl:
@Tratcher I think that would be good because without it, it's as bad as breaking from RC1->RC2. We need something.

Tratcher:
Updated with console warnings

Tratcher:
Updated after offline discussion with @Eilon and @blowdart. We're keeping the fallbacks indefinitely, but we still have the warnings.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -45,6 +45,21 @@ public WebHostBuilder()
...
+
+            // This may end up storing null, but that's indistinguishable from not adding it.
+            UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")
+                // Legacy keys, never remove these.
+                ?? Environment.GetEnvironmentVariable("Hosting:Environment")
+                ?? Environment.GetEnvironmentVariable("ASPNET_ENV"));
+
+            if (Environment.GetEnvironmentVariable("Hosting:Environment") != null)
+            {
+                Console.WriteLine("The environment variable 'Hosting:Environment' is obsolete and has been replaced by 'ASPNETCORE_ENVIRONMENT'");
```
Eilon: I would favor `replaced *with* XYZ`.

Eilon:
:shipit:

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -45,6 +45,21 @@ public WebHostBuilder()
             _hostingEnvironment = new HostingEnvironment();
             _configureServicesDelegates = new List<Action<IServiceCollection>>();
             _configureLoggingDelegates = new List<Action<ILoggerFactory>>();
+
+            // This may end up storing null, but that's indistinguishable from not adding it.
+            UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")
+                // Legacy keys, never remove these.
+                ?? Environment.GetEnvironmentVariable("Hosting:Environment")
```
davidfowl: These should be removed based on our latest discussion.

Issue #728, opened 2016-04-26
Self host deployers not testing on full CLR if test app is not published before deployment


JunTaoLuo:
Currently if the app is not published before deployment the command executed by the self host deployer looks like `dotnet run -p "C:\gh\Localization\samples\LocalizationSample" -- --server.urls http://localhost:5081/  --server Microsoft.AspNetCore.Server.Kestrel`. This is missing the framework specification `--framework net451/netcoreapp1.0` and will always run the coreclr version by default. As a result, some of our full CLR functional tests currently are tested with CoreCLR instead. We should update the deployer so we don't have this hole in our test coverage.

Eilon:
@muratg do we want to do this for RC2? If we're missing important test coverage we should add this ASAP.

muratg:
Yes, we should do this now. @JunTaoLuo I assume the change is not that hard (although if tests fail, fixing those may take some time.)

PR #743, opened 2016-05-06
Update self host deployer to target specific framework when lauching without publish

JunTaoLuo:
#728 As titled.

I'm renaming the property to reflect the fact that it is no longer only used for publishing. This will require changes in Entropy, MusicStore, IISIntegration, ServerTests, Localization.
I have tested this locally, both default targets and explicit targets, to verify that apps are launched in the correct framework but I'll wait for some feedback before seeking RC2 approval.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
@@ -74,8 +74,10 @@ namespace Microsoft.AspNetCore.Server.Testing
             }
             else
             {
+                var targetFramework = DeploymentParameters.TargetFramework ?? (DeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr ? "net451" : "netcoreapp1.0");
+
                 executableName = DotnetCommandName;
-                executableArgs = $"run -p \"{DeploymentParameters.ApplicationPath}\" {DotnetArgumentSeparator}";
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
```diff
@@ -74,8 +74,10 @@ protected CancellationToken StartSelfHost(Uri uri)
             }
             else
             {
+                var targetFramework = DeploymentParameters.TargetFramework ?? (DeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr ? "net451" : "netcoreapp1.0");
```
JunTaoLuo: I'm adding these default targets since I don't think they should change anymore and individual deployers can still override the default by specifying it explicitly. These defaults can also be added to `ApplicationDeployer`. The reason we had this option previously is because some apps contained multiple .NET Core targets (`dnxcore50` and `netstandardapp1.5` for example) making default targets less applicable but that is no longer the case.

Tratcher:
:shipit:

JunTaoLuo:
RC2 approval cc @Eilon @DamianEdwards. I will merge this into release instead of dev on which I accidentally opened this PR.

Eilon:
@JunTaoLuo go for it, and please keep an eye on the CI to make sure that affected tests pass (or disable them and log RC2 bugs to investigate).

Issue #731, opened 2016-04-27
Remove UseServer overloads


davidfowl:
We're moving away from generic hosting understanding how to wire up servers to servers adding themselves to the DI container via first class extension methods. This also avoids issues where the default extension methods setup things by default (like a default set of options) so that it's possible to configure server settings in the application.

muratg:
This would be a breaking change, right? So in RC2? Could you start a thread for approval?

cc @Eilon @DamianEdwards

Eilon:
What exactly do we want to remove/change? And are these APIs already frequently used?

muratg:
Let's revisit after RC2.

Tratcher:
Fowler's latest ask: Remove the extension, the config value, and the server factory that read the server value. Require servers be registered via DI in Main.

PR #756, opened 2016-05-20
Remove UseServer(string) overload

davidfowl:
- Removed the overload that takes a string because it's broken

#731

/cc @Tratcher @JunTaoLuo @DavidObando  (to make sure the stress repository doesn't use this thing)

### DIFF
--- a/test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
+++ b/test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
@@ -71,13 +71,12 @@ namespace Microsoft.AspNetCore.Hosting
         {
             var vals = new Dictionary<string, string>
             {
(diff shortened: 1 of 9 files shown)


test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -71,13 +71,12 @@ public void CanStartWithOldServerConfig()
         {
```
Tratcher: Theses tests were for the feature you removed, remove them.

Tratcher:
Some usages to replace before merging this:
https://github.com/search?q=org%3Aaspnet+UseServer&type=Code
:shipit:

JunTaoLuo:
Looks good to me after you remove the redundant tests. I think we still use this in some places for testing. We'll need to update them. :shipit:

Issue #690, opened 2016-04-06
Microsoft.AspNetCore.Server.Testing inconsistency between shipping status and version number


JunTaoLuo:
This package is used by our functional tests and we need to decide whether it should be shipped or not. If we decide to ship it, we will probably need to clean up the code and update `packages.csv` to `ship`. If not, we should change the version number to `0.1.0` from `1.0.0`  to be consistent with other noship packages and remove `1.0.0` versions from our feeds.

cc @davidfowl @Eilon @muratg

Eilon:
We can't avoid doing this in RC2 and then do it in RTM. This is either done in RC2 or never.

@muratg thoughts?

muratg:
Let me clarify one point:
 - **ship/noship/ext** -- this distinction is for the packages that we own and ship on nuget, packages we own but not push to nuget and partner packages
 - **Prerelease/RTM** -- this distinction is ONLY valid for "ship" packages (i.e. the ones we push to nuget.org).  Prerelease versions would have 0.1.0 version, and RTM ones would have 1.0.0.

This package is currently [noship](https://github.com/aspnet/Coherence/blob/dev/packages/packages.csv#L97) category (among other Testing packages)

So if we decide to ship this one (or more of the Testing packages), we'll need to do the version changes now. If not, we can skip the work.

Tratcher:
We can change the version for no-ship packages at any time, it's the same amount of work to purge them from our myget feeds. That said, the sooner the better.

Is there any reason not to change all of our no-ship package versions to v0.1?

Eilon:
@muratg ah ok so not urgent.

@Tratcher I suppose we could.

pakrym:
https://github.com/aspnet/MusicStore/commit/87e3544fda2f5cf89897ea6aefdcab683f4aebfd
 https://github.com/aspnet/ServerTests/commit/ed687d08b6f5bdd6e7c5a082c31856fcd26dccec
 https://github.com/aspnet/Localization/commit/0283099928382da520f958269bb714b3480680cb
 https://github.com/aspnet/Entropy/commit/0a0130560e52e1ce0df3601bb0ffe676ca675a2d
 https://github.com/aspnet/ServerTests/commit/ed687d08b6f5bdd6e7c5a082c31856fcd26dccec

PR #758, opened 2016-05-20
Downgrade Server.Testing version to 0.1

pakrym:
As per: https://github.com/aspnet/Hosting/issues/690

@pranavkm


Tratcher:
Update:
MusicStore
ServerTests
Localization
Entropy
IISIntegration
:shipit:

pranavkm:
We'll have to drop the package from the feeds too.

Issue #755, opened 2016-05-19
Add a second environment variable for Urls


glennc:
Add `ASPNETCORE_URLS` in addition to `server.urls`.

JunTaoLuo:
Curious why we want to add another environment variable that does the same thing? Is it because environment variables containing dots are hard to set in certain environments? If that's the case, maybe we should consider renaming the environment variable instead simply adding another one?

davidfowl:
> Curious why we want to add another environment variable that does the same thing? Is it because environment variables containing dots are hard to set in certain environments? 

> If that's the case, maybe we should consider renaming the environment variable instead simply adding another one?

We don't want to break anybody so we're going to add a new one and deprecate the old one. Eventually we can remove it.

PR #760, opened 2016-05-21
Support ASPNETCORE_URLS to set server urls

davidfowl:
/cc @Tratcher @glennc

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -9,7 +9,6 @@ using System.Reflection;
 using System.Runtime.ExceptionServices;
 using Microsoft.AspNetCore.Hosting.Builder;
 using Microsoft.AspNetCore.Hosting.Internal;
-using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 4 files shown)


Tratcher:
#755

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -47,14 +46,8 @@ public WebHostBuilder()
...
-            if (Environment.GetEnvironmentVariable("Hosting:Environment") != null)
-            {
-                Console.WriteLine("The environment variable 'Hosting:Environment' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
-            }
-            if (Environment.GetEnvironmentVariable("ASPNET_ENV") != null)
-            {
-                Console.WriteLine("The environment variable 'ASPNET_ENV' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
-            }
+            // Add the default server.urls key
+            UseSetting(WebHostDefaults.ServerUrlsKey, Environment.GetEnvironmentVariable("ASPNETCORE_URLS"));
```
Tratcher: I thought `urls` was going to become the primary key over here: https://github.com/aspnet/Hosting/blob/abe6dd5692c0d8bda573a25cdeb708a908dce517/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L191
And then `server.urls` would become the obsolete fallback.

davidfowl: That would break existing `hosting.json` files with server.urls.

Tratcher: You can read both in WebHost

davidfowl: Why? That's just more complicated

davidfowl: Made the change

PR #774, opened 2016-05-27
Fixed RemoteWindowsDeployer to enable copying  dontet runtime to target server f…

kichalla:
…or enabling portable apps scenario

@JunTaoLuo The changes here are for our CI scenario where we run tests against the latest dotnet runtime. Scenario is (i will send out a PR in music store too) that tests do the setup of dotnet runtime on target server upfront and then run the tests.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/RemoteWindowsDeployer/RemotePSSessionHelper.ps1
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/RemoteWindowsDeployer/RemotePSSessionHelper.ps1
@@ -10,6 +10,9 @@ param(
 	[string]$accountPassword,
 
 	[Parameter(Mandatory=$true)]
+	[string]$dotnetRuntimePath,
+
+	[Parameter(Mandatory=$true)]
 	[string]$executablePath,
 
 	[Parameter(Mandatory=$false)]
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Server.Testing/Deployers/RemoteWindowsDeployer/RemotePSSessionHelper.ps1
```diff
@@ -10,6 +10,9 @@ param(
 	[string]$accountPassword,
 
 	[Parameter(Mandatory=$true)]
+	[string]$dotnetRuntimePath,
```
JunTaoLuo: I don't think this should be a mandatory parameter. We don't need this for standalone apps.

src/Microsoft.AspNetCore.Server.Testing/Deployers/RemoteWindowsDeployer/RemoteWindowsDeployer.cs
```diff
@@ -53,6 +53,12 @@ public RemoteWindowsDeployer(RemoteWindowsDeploymentParameters deploymentParamet
                     " Account credentials are required to enable creating a powershell session to the remote server.");
             }
 
+            if (_deploymentParameters.ApplicationType == ApplicationType.Portable
+                && string.IsNullOrWhiteSpace(_deploymentParameters.DotnetRuntimePath))
+            {
+                throw new ArgumentException($"Invalid value '{_deploymentParameters.DotnetRuntimePath}' for {nameof(RemoteWindowsDeploymentParameters.DotnetRuntimePath)}. ");
```
JunTaoLuo: Ambiguous error message. Maybe something like "Invalid value for ... DotnetRuntimePath must be non-empty for portable apps."

JunTaoLuo:
:shipit:

kichalla:
bb3555c3dc48cad365277dc7c045be1abf9c5ed6

PR #780, opened 2016-06-03
Use Microsoft.Extensions.StackTrace.Sources

pranavkm:


pranavkm:
cc @Tratcher \ @davidfowl

src/Microsoft.AspNetCore.Hosting/project.json
```diff
@@ -30,11 +30,13 @@
     "Microsoft.Extensions.DependencyInjection": "1.0.0-*",
     "Microsoft.Extensions.Logging": "1.0.0-*",
     "Microsoft.Extensions.PlatformAbstractions": "1.0.0-*",
+    "Microsoft.Extensions.StackTrace.Sources": "1.0.0-*",
```
davidfowl: type build

src/Microsoft.AspNetCore.Hosting/project.json
```diff
@@ -50,7 +52,7 @@
         }
       }
     },
-    "netstandard1.3": {
+    "netstandard1.5": {
```
davidfowl: that sucks

cwe1ss: that would basically push the whole ecosystem to .net 4.6.2 wouldn't it? looks like a no-go?!

davidfowl: No we cross compile for net451 as well.

cwe1ss: I understand that netstandard is a good thing but it's also confusing as hell... isn't net451 here just for backwards-compatibility until all tools understand netstandard? why would it be necessary to have both if netstandard covers the full .net framework anyway? (sorry, not the best place to ask this but since you are the guru, I'm asking it here anyway :) )

pranavkm: @cwe1ss - Asp.Net Core supports .NET451 minimally (which maps to `netstandard1.2`). However `netstandard` exposes a subset of APIs that NET451 supports. In cases where we are unable to target a low enough netstandard to support NET451, we cross compile it.

I wrote some notes when I was converting `dnxcore50` -> `dotnet5.x` here (https://gist.github.com/pranavkm/1f2786f91a2e065ca0bb) which lists some of these APIs that caused us to cross compile. In this case specific case, `Assembly` exists in NETStandard1.3`, but `Assembly.Location` is only available in `NETStandard1.5`

benaadams: :crying_cat_face: 

Can you double compile for `netstandard1.3` and `netstandard1.5`; saying `netstandard1.5` will give better diagnostics and `#if` `Assembly.Location`?

Or would that not work?

pranavkm: That would work. Is there a reason you'd like us to cross-compile this project?

benaadams: Its caused the whole of aspnetcore's xplat to be at `netstandard1.5`/`4.6.2` which is a full framework pre-release level - which while it isn't a huge problem in itself (as coreclr is 1.5).

However, aspnetcore will be used as a primary example for library authors on what pattern to follow so will likely cause up-targeting across the ecosystem rather than encouraging them to target the lowest netstandard they can.

benaadams: Also anyone taking an RC2 dependency on aspnetcore libs will have to bump from `netstandard1.3` to `netstandard1.5` for rtm

benaadams: If previously I wanted a Kestrel upgrade stream/filter support to 4.6+ & core I'd just need to use `netstandard1.3`; however now will need to use both `net46` and `netstandard1.5` as there isn't an upgrade path between them

pranavkm: :+1: I'll have it cross compile to `netstandard1.3`. cc @Eilon for ask mode.

PR #787, opened 2016-06-06
Fixed SelfHostDeployer and IISExpressDeployer to use the correct appl…

kichalla:
…ication path

@Tratcher @JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/IISExpressDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/IISExpressDeployer.cs
@@ -34,15 +34,15 @@ namespace Microsoft.AspNetCore.Server.Testing
                 DotnetPublish();
             }
 
-            var webRootLocation = DeploymentParameters.PublishApplicationBeforeDeployment ? DeploymentParameters.PublishedApplicationRootPath : DeploymentParameters.ApplicationPath;
+            var appRoot = DeploymentParameters.PublishApplicationBeforeDeployment ? DeploymentParameters.PublishedApplicationRootPath : DeploymentParameters.ApplicationPath;
 
             var uri = TestUriHelper.BuildTestUri(DeploymentParameters.ApplicationBaseUriHint);
             // Launch the host process.
-            var hostExitToken = StartIISExpress(uri, webRootLocation);
+            var hostExitToken = StartIISExpress(uri, appRoot);
 
             return new DeploymentResult
             {
-                WebRootLocation = webRootLocation,
+                AppRoot = appRoot,
                 DeploymentParameters = DeploymentParameters,
                 // Right now this works only for urls like http://localhost:5001/. Does not work for http://localhost:5001/subpath.
                 ApplicationBaseUri = uri.ToString(),
@@ -50,7 +50,7 @@ namespace Microsoft.AspNetCore.Server.Testing
             };
         }
 
-        private CancellationToken StartIISExpress(Uri uri, string webRootLocation)
+        private CancellationToken StartIISExpress(Uri uri, string appRoot)
         {
             if (!string.IsNullOrWhiteSpace(DeploymentParameters.ServerConfigTemplateContent))
             {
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.Server.Testing/Deployers/IISExpressDeployer.cs
```diff
@@ -34,21 +34,23 @@ public override DeploymentResult Deploy()
...
+            var webRootLocation = DeploymentParameters.PublishApplicationBeforeDeployment ? DeploymentParameters.PublishedApplicationRootPath : DeploymentParameters.ApplicationPath;
+
             var uri = TestUriHelper.BuildTestUri(DeploymentParameters.ApplicationBaseUriHint);
             // Launch the host process.
-            var hostExitToken = StartIISExpress(uri);
+            var hostExitToken = StartIISExpress(uri, webRootLocation);
 
             return new DeploymentResult
             {
-                WebRootLocation = DeploymentParameters.ApplicationPath,
```
kichalla: BTW, I saw similar issue with Nginx deployer too. If you feel these changes are good to do I will update Nginx deployer too.

kichalla:
@BrennanConroy

src/Microsoft.AspNetCore.Server.Testing/Deployers/IISExpressDeployer.cs
```diff
@@ -57,7 +59,7 @@ private CancellationToken StartIISExpress(Uri uri)
 
                 DeploymentParameters.ServerConfigTemplateContent =
                     DeploymentParameters.ServerConfigTemplateContent
-                        .Replace("[ApplicationPhysicalPath]", DeploymentParameters.ApplicationPath)
+                        .Replace("[ApplicationPhysicalPath]", webRootLocation)
```
Tratcher: This should still be ApplicationPath, not webroot. webroot is for static files.

src/Microsoft.AspNetCore.Server.Testing/Deployers/IISExpressDeployer.cs
```diff
@@ -66,7 +68,7 @@ private CancellationToken StartIISExpress(Uri uri)
             }
 
             var parameters = string.IsNullOrWhiteSpace(DeploymentParameters.ServerConfigLocation) ?
-                            string.Format("/port:{0} /path:\"{1}\" /trace:error", uri.Port, DeploymentParameters.ApplicationPath) :
+                            string.Format("/port:{0} /path:\"{1}\" /trace:error", uri.Port, webRootLocation) :
```
Tratcher: This one too

src/Microsoft.AspNetCore.Server.Testing/Deployers/IISExpressDeployer.cs
```diff
@@ -34,21 +34,23 @@ public override DeploymentResult Deploy()
                 DotnetPublish();
             }
 
+            var webRootLocation = DeploymentParameters.PublishApplicationBeforeDeployment ? DeploymentParameters.PublishedApplicationRootPath : DeploymentParameters.ApplicationPath;
```
Tratcher: Ah, this is just misnamed. webroot is for static files, not app root.

kichalla: ah, you mean this should be named something like `appRoot` instead and rest of the changes are good?

Tratcher: If you make it appRoot then StartIISExpress is correct, but DeploymentResult.WebRootLocation is wrong. What's WebRootLocation used for?

Tratcher: I think WebRootLocation was never renamed when we moved web.config from wwwroot to the parent directory. It looks like it should be renamed to AppRoot

kichalla: - I glanced through the tests and do not see this property being used
- From the documentation here https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentResult.cs#L18-L21, it clearly doesn't indicate about `wwwroot` but i agree `AppRoot` is a better name

JunTaoLuo: To be consistent with what we used as hosting keys, we should call it `contentRoot`

kichalla:
:up: :date:

Tratcher:
:shipit:

JunTaoLuo:
:shipit:

kichalla:
808ee7e96542e0e4c47d532572f4a9ae7febf0a8

Issue #784, opened 2016-06-06
Add ASPNETCORE_DETAILEDERRORS back to default configuration


davidfowl:
We should be reading this value by default.

/cc @Tratcher

Tratcher:
At what point do we just read all environment variables by default?

davidfowl:
We *could* read the ones we have settings for. Everything else is kinda pointless.

Tratcher:
Seed config with `.AddEnvironmentVariables(prefix: "ASPNETCORE_")` rather than adding them all piecemeal.

davidfowl:
Yep! Sounds good to me. That was the original bug I filed.

PR #790, opened 2016-06-07
Read environment variables by default

JunTaoLuo:
#784 

cc @Tratcher @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -27,7 +27,7 @@ namespace Microsoft.AspNetCore.Hosting
         private readonly List<Action<IServiceCollection>> _configureServicesDelegates;
         private readonly List<Action<ILoggerFactory>> _configureLoggingDelegates;
 
-        private IConfiguration _config = new ConfigurationBuilder().AddInMemoryCollection().Build();
+        private IConfiguration _config;
         private ILoggerFactory _loggerFactory;
         private WebHostOptions _options;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -40,14 +40,20 @@ public WebHostBuilder()
...
             // This may end up storing null, but that's indistinguishable from not adding it.
             UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")
                 // Legacy keys, never remove these.
                 ?? Environment.GetEnvironmentVariable("Hosting:Environment")
                 ?? Environment.GetEnvironmentVariable("ASPNET_ENV"));
 
             // Add the default server.urls key
-            UseSetting(WebHostDefaults.ServerUrlsKey, Environment.GetEnvironmentVariable("ASPNETCORE_URLS"));
+            UseSetting(WebHostDefaults.ServerUrlsKey, Environment.GetEnvironmentVariable("ASPNETCORE_URLS")
+                // Lebacy key, never remove this
```
nil4: Did you mean *Legacy*?

JunTaoLuo: Oops, Engrish is hard haha. Will fix.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -40,14 +40,20 @@ public WebHostBuilder()
             _configureServicesDelegates = new List<Action<IServiceCollection>>();
             _configureLoggingDelegates = new List<Action<ILoggerFactory>>();
 
+            _config = new ConfigurationBuilder()
+                .AddEnvironmentVariables(prefix: "ASPNETCORE_")
+                .Build();
+
             // This may end up storing null, but that's indistinguishable from not adding it.
             UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT")
```
Tratcher: It seems strange to read ASPNETCORE_ENVIRONMENT and URLS here when we just added ASPNETCORE_ above. What if we restructured this like:
```
UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("Hosting:Environment") ?? Environment.GetEnvironmentVariable("ASPNET_ENV"));
UseSetting(WebHostDefaults.ServerUrlsKey, Environment.GetEnvironmentVariable("ASPNETCORE_SERVER.URLS"));
UseConfiguration(new ConfigurationBuilder().AddEnvironmentVariables(prefix: "ASPNETCORE_").Build());
```

JunTaoLuo: It wouldn't matter because it would just set those configurations again. It's idempotent so it's not wrong but I do think your suggestion looks a little cleaner. Though in this case we need to first create an InMemoryCollection, add the old keys, create a new configuration builder with environment variables and add call `UseConfiguration` which is a lot of fluff. *shrug*

JunTaoLuo: Interesting C# quirk. You cannot call an extension method of the interface from the class that is implementing the interface unless it's done explicitly. This means that `UseConfiguration` which is an extension on `IWebHostBuilder` cannot be called from `WebHostBuilder` unless you call it like `HostingAbstractionsWebHostBuilderExtensions.UseConfiguration(this, config)`. Hmm, since that looks ugly as well, I suggest we just keep the current implementation.

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -40,14 +40,21 @@ public WebHostBuilder()
...
+            _config = new ConfigurationBuilder()
+                .AddEnvironmentVariables(prefix: "ASPNETCORE_")
+                .Build();
 
-            // Add the default server.urls key
-            UseSetting(WebHostDefaults.ServerUrlsKey, Environment.GetEnvironmentVariable("ASPNETCORE_URLS"));
+            if (string.IsNullOrEmpty(GetSetting(WebHostDefaults.EnvironmentKey)))
+            {
+                // Try adding legacy environment keys, never remove these.
+                UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("Hosting:Environment") ?? Environment.GetEnvironmentVariable("ASPNET_ENV"));
```
Tratcher: nit: line wrap

Tratcher:
:shipit:

PR #795, opened 2016-06-09
Output "Application is shutting down" message

### LINKED ISSUE #794: Output to console that the application is shutting down gracefully
Until recently whenever I wanted to stop my app I felt that `ctrl+c` felt really slow without a reason, I didn't know that it was canceling the shutdown to wait for the main thread to exit gracefully.

So perhaps output a message after [this line of code](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs#L23) to say that "Application is shutting down." so that it doesn't feel like nothing is going on?

mrahhal:
closes #794


Tratcher:
Rebased and merged. Thanks

Issue #812, opened 2016-07-12
WebHost.Run() completes before ApplicationStopping


ZoolWay:
When using `applicationLifetime.ApplicationStopping.Register(HandleStopping);` the program exits `host.Run();` before that handler has completed also the documentation (IntelliSense) says *Shutdown will block until this event completes*.
This also means that ApplicationStopped will be executed before ApplicationStopping has completed.

Also it should be noted that ApplicationStopping is not executed on the main thread (while ApplicationStopped is) and that you should not use `async` for those handler as that will cause the methods to be completed.

This is ASP.NET 1.0.0 on Windows 10 using Kestrel (1.0.0).

Startup.cs
```
public void Configure(IApplicationLifetime applicationLifetime)
{
    applicationLifetime.ApplicationStopping.Register(HandleStopping);
    applicationLifetime.ApplicationStopped.Register(HandleStopped);
}

private void HandleStopped()
{
    System.Threading.Thread.Sleep(30000);
    // WebHost.Run() will not end until this method completes, running it in the main thread
}

private void HandleStopping()
{
    System.Threading.Thread.Sleep(30000);
    // WebHost.Run() will end even while this method is still working, it will even run HandleStopped before this completed. It run on a worker thread.
}
```

davidfowl:
Yep I can reproduce the problem.

GuardRex:
Not directly related, but speaking of doing things on "ProcessExit" in case someone is looking around at various options, see:  https://github.com/dotnet/corefx/issues/10012#issuecomment-232436326

mrahhal:
Is there anything we can do about this for now? Can't depend on `1.1.0-preview1` just yet. Had to work hard to track down this issue as I've been depending on `IApplicationLifetime` and started noticing background tasks abruptly exiting without notice.

Thought about doing some manual sync that makes `Stopped` wait for all `Stopping`, but since services are being disposed directly all tasks that depend on other services will probably face a lot of "object disposed exceptions" in all cases. Hmm this really looks hopeless without including a custom build of this package.

davidfowl:
It might be worth putting this fix in a patch release.

/cc @Eilon @muratg @DamianEdwards

davidfowl:
@mrahhal what problem is it causing you right now? What does your repro look like?

mrahhal:
@davidfowl that would be super cool. Because with the current state, `IApplicationLifetime.ApplicationStopping` is kinda useless (and harmful even).

[...]

PR #818, opened 2016-07-18
Update shutdown logic for hosted applications

JunTaoLuo:
#812 

This new logic will block until `ApplicationStopping` and `ApplicationStopped` are run until completion on the main thread before `host.Run()` returns. Adding a `ManualResetEvent` seems arbitrary though. Any ideas? cc @davidfowl @Tratcher 

I'll also look into adding some tests in the mean time.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
@@ -39,13 +39,19 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         /// </summary>
         public void StopApplication()
         {
-            try
+            // Lock on CTS to synchronize multiple calls to StopApplication. This guarantees that the first call 
+            // to StopApplication and its callbacks run to completion before subsequent calls to StopApplication, 
+            // which will no-op since the first call already requested cancellation, get a chance to execute.
+            lock (_stoppingSource)
             {
-                _stoppingSource.Cancel(throwOnFirstException: false);
-            }
-            catch (Exception)
-            {
-                // TODO: LOG
+                try
+                {
+                    _stoppingSource.Cancel(throwOnFirstException: false);
+                }
+                catch (Exception)
+                {
+                    // TODO: LOG
+                }
             }
         }
 
--- a/test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
+++ b/test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
@@ -167,6 +167,67 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         [Fact]
+        public void WebHostApplicationLifetimeEventsOrderedCorrectlyDuringShutdown()
+        {
+            var host = CreateBuilder()
+                .UseServer(this)
+                .UseStartup("Microsoft.AspNetCore.Hosting.Tests")
+                .Build();
+
+            var lifetime = host.Services.GetRequiredService<IApplicationLifetime>();
+            var applicationStartedEvent = new ManualResetEventSlim(false);
+            var applicationStoppingEvent = new ManualResetEventSlim(false);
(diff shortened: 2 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -184,6 +185,9 @@ private IServiceCollection BuildHostingServices()
                 configureLogging(_loggerFactory);
             }
 
+            // Create an event for stopping the hosted application
+            services.AddSingleton<ManualResetEvent>(new ManualResetEvent(false));
```
Tratcher: Why is this a service if it's only used in Run?

davidfowl: Yea, this shouldn't be a service.

Tratcher:
This only works around the issue for Run, it does not address the underlying problem with ApplicationStopping.

JunTaoLuo:
:up::date: Talked with @Tratcher  and a better idea is to just synchronize `ApplicationStopping` instead. Originally the second call to `ApplicationStopping` no-oped and immediately returned in the `Dispose` method of `WebHost` and allowed `ApplicationStopped` to execute. With the lock added, subsequent calls to `ApplicationStopping` will lock until the callbacks triggered by the first call completes. Note that we guarantee the completion of `ApplicationStopping` callbacks before `ApplicationStopped` but we are not guaranteeing that the `ApplicationStopping` callbacks are executed on the main thread.

src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
```diff
@@ -39,13 +39,16 @@ public class ApplicationLifetime : IApplicationLifetime
         /// </summary>
         public void StopApplication()
         {
-            try
+            lock(_stoppingSource)
```
Tratcher: This could use a comment. Normally you wouldn't need to lock around a CTS.

davidfowl:
Where's the test?

JunTaoLuo:
@davidfowl working on it

JunTaoLuo:
:up::date: with test added

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -167,6 +167,57 @@ public void WebHostShutsDownWhenTokenTriggers()
...
+            lifetime.ApplicationStarted.Register(() =>
+            {
+                Assert.Equal(1, shutdownOrder++);
+            });
+
+            lifetime.ApplicationStopping.Register(() =>
+            {
+                Assert.Equal(2, shutdownOrder++);
+                // Simulate work. ApplicationStopping is triggered first but make it slower than ApplicationStopped.
+                Thread.Sleep(3000);
```
Tratcher: ManualResetEvent

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -167,6 +167,57 @@ public void WebHostShutsDownWhenTokenTriggers()
...
+            var shutdownOrder = 0;
+
+            lifetime.ApplicationStarted.Register(() =>
+            {
+                Assert.Equal(1, shutdownOrder++);
+            });
+
+            lifetime.ApplicationStopping.Register(() =>
+            {
+                Assert.Equal(2, shutdownOrder++);
```
Tratcher: Asserts failures inside the callbacks may not be observed by the test

JunTaoLuo:
I've removed all but one call to `Thread.Sleep()` in the tests based on @Tratcher's comment. I'll see if I can find a clean way of removing the last instance while preserving the utility of the test.

davidfowl:
:shipit:

muratg:
https://github.com/aspnet/Hosting/pull/818/files

PR #827, opened 2016-08-04
Rybrande/working directory

ryanbrandenburg:
Set the working directory of the SelfHostDeployer to the ApplicationPath so that projects that do `UseContentRoot(Directory.GetCurrentDirectory())` (Like https://github.com/aspnet/Entropy/tree/dev/samples/Localization.StarterWeb) will be able to find their content.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs
@@ -94,7 +94,8 @@ namespace Microsoft.AspNetCore.Server.Testing
                 RedirectStandardError = true,
                 RedirectStandardOutput = true,
                 // Trying a work around for https://github.com/aspnet/Hosting/issues/140.
-                RedirectStandardInput = true
+                RedirectStandardInput = true,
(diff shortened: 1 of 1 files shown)


Tratcher:
What changed this week that caused Loc to stop working?

ryanbrandenburg:
CC @BrennanConroy, this should fix the Entropy failure caused by Localization.StarterWeb using it's ContentRoot. The problem ended up being that it couldn't find its appsettings.json because it was looking in the executing directory instead of in the application path. I ran all the Entropy and Hosting tests but someone with a little more domain knowledge might want to give this a once over to ensure I haven't missed a side effect.

ryanbrandenburg:
We did this (https://github.com/aspnet/Entropy/commit/5f0a1209986096436d94d7f7fbe07f21cd55582b) to fix the fact that that sample wouldn't launch normally, it only passed the test. It looks like @kichalla has backed that change out temporarily to fix the build, but I think we should re-instate it once this is in to keep the samples up to date with our recommended pattern.

ryanbrandenburg:
@Tratcher are you the person best able to review this?

ryanbrandenburg:
Nevermind, @kichalla beat me to the punch in #828.

PR #828, opened 2016-08-04
Set deployed application root folder as working directory

kichalla:
@Tratcher @pranavkm I verified these changes against Entorpy and ServerTests repos. I couldn't run MusicStore due to package restore errors.

@Tratcher Some background: This change fixes issues where ContentRoot is set as Directory.GetCurrentDirectory in an app


pranavkm:
:shipit: once we know MusicStore also works.

PR #830, opened 2016-08-08
ApplicationDeployer: Supply the build configuration to dotnet publish

kichalla:
@Tratcher I working on the following issue when I came across this being a potential cause for it.
https://github.com/aspnet/Coherence-Signed/issues/348

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
@@ -72,6 +72,11 @@ namespace Microsoft.AspNetCore.Server.Testing
         public string TargetFramework { get; set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Tratcher:
:shipit: though I don't see how that causes your issue.
Are you going to set this in any of the tests?

kichalla:
> Are you going to set this in any of the tests?

Yeah I am going to verify this change with changes to tests in MusicStore for now. I would like to verify it on CI specifically (as I am unable to repro the referenced issue on my machine). I was thinking that this fix is safe even it doesn't work as I thought because we however seem to be testing for `Debug` always (not sure if that's what we intended to do as CI nodes has a Configuration setting for `Release` and so the publish should have happened in `Release`?)

The other alternative is for the deployer's publish code to read the environment variable `Configuration` and use it if available, otherwise fallback to `Debug`. Do you think this is a better alternative?

PR #832, opened 2016-08-09
Updated ApplicationDeployer to conditionally delete published applica…

kichalla:
…tion folder

@JunTaoLuo

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
@@ -81,6 +81,8 @@ namespace Microsoft.AspNetCore.Server.Testing
         /// </summary>
         public bool PublishApplicationBeforeDeployment { get; set; }
 
+        public bool DeletePublishedApplicationOnDispose { get; set; } = true;
+
         public ApplicationType ApplicationType { get; set; }
 
         public string PublishedApplicationRootPath { get; set; }
--- a/src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
@@ -82,14 +82,23 @@ namespace Microsoft.AspNetCore.Server.Testing
 
         protected void CleanPublishedOutput()
         {
-            try
+            if (DeploymentParameters.DeletePublishedApplicationOnDispose)
             {
-                // We've originally published the application in a temp folder. We need to delete it.
-                Directory.Delete(DeploymentParameters.PublishedApplicationRootPath, true);
+                try
+                {
+                    // We've originally published the application in a temp folder. We need to delete it.
+                    Directory.Delete(DeploymentParameters.PublishedApplicationRootPath, true);
+                }
+                catch (Exception exception)
+                {
+                    Logger.LogWarning($"Failed to delete directory : {exception.Message}");
+                }
             }
-            catch (Exception exception)
+            else
             {
-                Logger.LogWarning($"Failed to delete directory : {exception.Message}");
+                Logger.LogWarning(
+                    "Skipping deleting the locally published folder as property " +
+                    $"'{nameof(DeploymentParameters.DeletePublishedApplicationOnDispose)}' is set to 'false'.");
             }
         }
 


Tratcher:
Why?
Also, why make it an environment variable rather than a DeploymentParameter?

kichalla:
> Why?
Also, why make it an environment variable rather than a DeploymentParameter?

Basically when a test fails/succeeds we currently delete the published folder which prevents from diagnosing the issue if I want to look at the published output and then run the app. I chose environment variable as other approaches (ex: deployment parameter) would require code to be changed to start debugging.

Yeah, I know that having an environment variable reduces discoverability but this would be the same with other approaches too?

Tratcher:
Are there any other places in the deployers that directly read environment variables? I'd still favor code-first.

kichalla:
No, there are no other places reading environment variables.

JunTaoLuo:
Instead of reading the environment variable in the deployers, maybe have the test projects read/parse it and then pass it along as a deployment parameter?

kichalla:
Yeah, I considered that but since all test projects need to do this wanted to avoid it. Anyways, I guess I will just use deployment parameters and each test project can read the environment variable and set it as needed.

src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -82,14 +82,31 @@ protected void DotnetPublish(string publishRoot = null)
 
         protected void CleanPublishedOutput()
         {
-            try
+            var deletePublishedFolderEnvironmentVariableName = "DELETE_PUBLISHED_FOLDER";
```
JunTaoLuo: Would also recommend to add a prefix `ASPNETCORE_` just to differentiate.

src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -82,14 +82,31 @@ protected void DotnetPublish(string publishRoot = null)
...
+            var deleteLocallyPublishedFolder = true;
+            var setting = Environment.GetEnvironmentVariable(deletePublishedFolderEnvironmentVariableName);
+            if (string.Equals(setting, "false", StringComparison.OrdinalIgnoreCase))
```
JunTaoLuo: I would take either `false` or `0` to disable the setting. See https://github.com/aspnet/Hosting/blob/41b4e6017e5d2c67e3a3f031558ded23abd28d58/src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs#L45-L49.

JunTaoLuo:
The rest of the change looks good, some minor comments on the parsing of environment variables that will be moved to the tests.

kichalla:
:up: :date:

src/Microsoft.AspNetCore.Server.Testing/Deployers/ApplicationDeployer.cs
```diff
@@ -82,14 +82,23 @@ protected void DotnetPublish(string publishRoot = null)
...
+                catch (Exception exception)
+                {
+                    Logger.LogWarning($"Failed to delete directory : {exception.Message}");
+                }
             }
-            catch (Exception exception)
+            else
             {
-                Logger.LogWarning($"Failed to delete directory : {exception.Message}");
+                Logger.LogWarning(
```
JunTaoLuo: should be `is set to 'false'` right?

JunTaoLuo:
Changes looks good but the exception message looks wrong. With that fixed: :shipit: .

src/Microsoft.AspNetCore.Server.Testing/Common/DeploymentParameters.cs
```diff
@@ -81,6 +81,8 @@ public class DeploymentParameters
         /// </summary>
         public bool PublishApplicationBeforeDeployment { get; set; }
 
+        public bool DeletePublishedApplicationOnDispose { get; set; } = true;
```
Tratcher: How about PreservePublishedApplicaitonForDebugging = false?

kichalla: Sounds good. I already pushed my changes but will submit one more commit.

Issue #829, opened 2016-08-06
Proposal: Support IServiceProviderFactory


davidfowl:
Based on https://github.com/aspnet/DependencyInjection/issues/442

The `WebHostBuilder` API would take advantage of this by using it as a way to override the default `IServiceProvider` implementation without returning it directly from the application's `Startup` class. The wire up could look like this:

```C#
public static void Main(string[] args)
{
    var host = new WebHostBuilder()
        .UseKestrel()
        .UseAutofac() // Replace the built-in IServiceProviderFactory with an autofac impl
        .UseContentRoot(Directory.GetCurrentDirectory())
        .UseStartup<Startup>()
        .Build();

    host.Run();
}
```

One downside to this is that you don't get direct access to the container APIs in the `Startup` class (which is part of the benefit besides being able to swap out the default impl).

In the above example, `UseAutofac` would wire up the `IServiceProviderFactory`. We could also extend the `ConventionBasedStartup` to allow authoring a `ConfigureContainer` method that takes the concrete ContainerBuilder directly:

```C#
public class Startup
{
    public void ConfigureServices(IServiceCollection services)
    {
         services.AddMvc();
    }

    public void ConfigureContainer(ContainerBuilder builder)
    {
        builder.RegisterInstance(new TaskRepository())
           .As<ITaskRepository>();
    }

    public void Configure(IApplicationBuilder app)
    {
        app.UseMvc();
    }
}
```

GuardRex:
> you don't get direct access to the container APIs in the Startup class

Will there be any consequences for those who despise the vile and dreaded `Startup` class? :smile:
```c#
public static void Main(string[] args)
{
    var host = new WebHostBuilder()
        .UseKestrel()
        .UseContentRoot(Directory.GetCurrentDirectory())
        .ConfigureServices(services =>
        {
            ...
        })
        .Configure((app) =>
        {
            ...
        })
        .Build();

    host.Run();
}
```

davidfowl:
You can't replace the `IServiceProvider` today if you configure everything with the `WebHostBuilder` today. See https://github.com/aspnet/Hosting/issues/550

GuardRex:
Fair enough ... just checking. Since I don't, it won't affect me. :smile:

Tratcher:
How does it work if UseKestrel is already adding and accessing services?

davidfowl:
[...]

PR #833, opened 2016-08-09
Service provider factory and Startup.ConfigureContainer

davidfowl:
https://github.com/aspnet/Hosting/issues/829

/cc @pakrym @halter73

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/ConfigureServicesBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/ConfigureServicesBuilder.cs
@@ -12,19 +12,6 @@ namespace Microsoft.AspNetCore.Hosting.Internal
     {
         public ConfigureServicesBuilder(MethodInfo configureServices)
         {
-            if (configureServices == null)
-            {
-                throw new ArgumentNullException(nameof(configureServices));
-            }
-
-            // Only support IServiceCollection parameters
-            var parameters = configureServices.GetParameters();
-            if (parameters.Length > 1 ||
-                parameters.Any(p => p.ParameterType != typeof(IServiceCollection)))
-            {
-                throw new InvalidOperationException("The ConfigureServices method must either be parameterless or take only one parameter of type IServiceCollection.");
-            }
-
             MethodInfo = configureServices;
         }
 
@@ -32,22 +19,29 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
         public Func<IServiceCollection, IServiceProvider> Build(object instance) => services => Invoke(instance, services);
 
-        private IServiceProvider Invoke(object instance, IServiceCollection exportServices)
+        private IServiceProvider Invoke(object instance, IServiceCollection services)
         {
-            if (exportServices == null)
+            if (MethodInfo == null)
             {
-                throw new ArgumentNullException(nameof(exportServices));
+                return null;
             }
 
-            var parameters = new object[MethodInfo.GetParameters().Length];
+            // Only support IServiceCollection parameters
+            var parameters = MethodInfo.GetParameters();
+            if (parameters.Length > 1 ||
+                parameters.Any(p => p.ParameterType != typeof(IServiceCollection)))
+            {
+                throw new InvalidOperationException("The ConfigureServices method must either be parameterless or take only one parameter of type IServiceCollection.");
+            }
 
-            // Ctor ensures we have at most one IServiceCollection parameter
+            var arguments = new object[MethodInfo.GetParameters().Length];
+
             if (parameters.Length > 0)
             {
-                parameters[0] = exportServices;
+                arguments[0] = services;
             }
 
-            return MethodInfo.Invoke(instance, parameters) as IServiceProvider ?? exportServices.BuildServiceProvider();
+            return MethodInfo.Invoke(instance, arguments) as IServiceProvider;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
@@ -11,18 +11,56 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class StartupLoader
     {
-        public static StartupMethods LoadMethods(IServiceProvider services, Type startupType, string environmentName)
+        public static StartupMethods LoadMethods(IServiceProvider hostingServiceProvider, Type startupType, string environmentName)
         {
             var configureMethod = FindConfigureDelegate(startupType, environmentName);
             var servicesMethod = FindConfigureServicesDelegate(startupType, environmentName);
+            var configureContainerMethod = FindConfigureContainerDelegate(startupType, environmentName);
 
             object instance = null;
             if (!configureMethod.MethodInfo.IsStatic || (servicesMethod != null && !servicesMethod.MethodInfo.IsStatic))
             {
-                instance = ActivatorUtilities.GetServiceOrCreateInstance(services, startupType);
+                instance = ActivatorUtilities.GetServiceOrCreateInstance(hostingServiceProvider, startupType);
             }
 
-            return new StartupMethods(configureMethod.Build(instance), servicesMethod?.Build(instance));
+            var configureServicesCallback = servicesMethod.Build(instance);
+            var configureContainerCallback = configureContainerMethod.Build(instance);
+
+            Func<IServiceCollection, IServiceProvider> configureServices = services =>
+            {
+                // Call ConfigureServices, if that returned an IServiceProvider, we're done
+                IServiceProvider applicationServiceProvider = configureServicesCallback.Invoke(services);
+
+                if (applicationServiceProvider != null)
+                {
+                    return applicationServiceProvider;
+                }
+
+                // If there's a ConfigureContainer method
+                if (configureContainerMethod.MethodInfo != null)
+                {
+                    // We have a ConfigureContainer method, get the IServiceProviderFactory<TContainerBuilder>
+                    var serviceProviderFactoryType = typeof(IServiceProviderFactory<>).MakeGenericType(configureContainerMethod.GetContainerType());
+                    var serviceProviderFactory = hostingServiceProvider.GetRequiredService(serviceProviderFactoryType);
+                    // var builder = serviceProviderFactory.CreateBuilder(services);
+                    var builder = serviceProviderFactoryType.GetMethod(nameof(DefaultServiceProviderFactory.CreateBuilder)).Invoke(serviceProviderFactory, new object[] { services });
+                    configureContainerCallback.Invoke(builder);
+                    // applicationServiceProvider = serviceProviderFactory.CreateServiceProvider(builder);
+                    applicationServiceProvider = (IServiceProvider)serviceProviderFactoryType.GetMethod(nameof(DefaultServiceProviderFactory.CreateServiceProvider)).Invoke(serviceProviderFactory, new object[] { builder });
+                }
+                else
+                {
+                    // Get the default factory
+                    var serviceProviderFactory = hostingServiceProvider.GetRequiredService<IServiceProviderFactory<IServiceCollection>>();
+
+                    // Don't bother calling CreateBuilder since it just returns the default service collection
+                    applicationServiceProvider = serviceProviderFactory.CreateServiceProvider(services);
+                }
+
+                return applicationServiceProvider ?? services.BuildServiceProvider();
+            };
+
+            return new StartupMethods(configureMethod.Build(instance), configureServices);
         }
 
         public static Type FindStartupType(string startupAssemblyName, string environmentName)
@@ -30,8 +68,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             if (string.IsNullOrEmpty(startupAssemblyName))
             {
                 throw new ArgumentException(
-                    string.Format("A startup method, startup type or startup assembly is required. If specifying an assembly, '{0}' cannot be null or empty.", 
-                    nameof(startupAssemblyName)), 
+                    string.Format("A startup method, startup type or startup assembly is required. If specifying an assembly, '{0}' cannot be null or empty.",
+                    nameof(startupAssemblyName)),
                     nameof(startupAssemblyName));
             }
 
@@ -83,11 +121,17 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             return new ConfigureBuilder(configureMethod);
         }
 
+        private static ConfigureContainerBuilder FindConfigureContainerDelegate(Type startupType, string environmentName)
+        {
+            var configureMethod = FindMethod(startupType, "Configure{0}Container", environmentName, typeof(void), required: false);
+            return new ConfigureContainerBuilder(configureMethod);
+        }
+
         private static ConfigureServicesBuilder FindConfigureServicesDelegate(Type startupType, string environmentName)
         {
             var servicesMethod = FindMethod(startupType, "Configure{0}Services", environmentName, typeof(IServiceProvider), required: false)
                 ?? FindMethod(startupType, "Configure{0}Services", environmentName, typeof(void), required: false);
-            return servicesMethod == null ? null : new ConfigureServicesBuilder(servicesMethod);
+            return new ConfigureServicesBuilder(servicesMethod);
         }
 
         private static MethodInfo FindMethod(Type startupType, string methodName, string environmentName, Type returnType = null, bool required = true)
--- a/src/Microsoft.AspNetCore.Hosting/Internal/StartupMethods.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/StartupMethods.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.DependencyInjection;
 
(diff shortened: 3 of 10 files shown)


src/Microsoft.AspNetCore.Hosting/ConfigureContainerBuilder.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
```
pakrym: (c)

test/Microsoft.AspNetCore.TestHost.Tests/Internal/MyContainerFactoryWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.TestHost.Tests.Internal
+{
+    public static class MyContainerFactoryWebHostBuilderExtensions
+    {
+        public static IWebHostBuilder UseMyContainer(this IWebHostBuilder host)
```
pakrym: Turn it into `UseContainer<T>()` method on `WebHostBuilder`

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
```diff
@@ -16,4 +16,25 @@ public virtual IServiceProvider ConfigureServices(IServiceCollection services)
...
+        private readonly IServiceProviderFactory _factory;
+
+        public StartupBase(IServiceProviderFactory factory)
+        {
+            _factory = factory;
+        }
+
+        public abstract void Configure(IApplicationBuilder app);
+
+        public IServiceProvider ConfigureServices(IServiceCollection services)
```
pakrym: If user derives from this class, how would his `ConfigureServices` look like?

davidfowl: This should be virtual. The user code would override and return `base.ConfigureServices(services)`

khellang: Why not make it explicit and have a `void ConfigureServices(IServiceCollection services)`, like [this](https://github.com/aspnet/Hosting/commit/7556d29381df67a69ee84b688a6a05b2e8825518#diff-4c7c346129759b4849377b3494d5b17fR40)?

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -11,27 +11,57 @@ namespace Microsoft.AspNetCore.Hosting.Internal
...
 
             object instance = null;
             if (!configureMethod.MethodInfo.IsStatic || (servicesMethod != null && !servicesMethod.MethodInfo.IsStatic))
             {
-                instance = ActivatorUtilities.GetServiceOrCreateInstance(services, startupType);
+                instance = ActivatorUtilities.GetServiceOrCreateInstance(hostingServiceProvider, startupType);
             }
 
-            return new StartupMethods(configureMethod.Build(instance), servicesMethod?.Build(instance));
+            var configureServicesCallback = servicesMethod.Build(instance);
```
khellang: `servicesMethod` can be `null`, no?

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -83,11 +113,17 @@ private static ConfigureBuilder FindConfigureDelegate(Type startupType, string e
             return new ConfigureBuilder(configureMethod);
         }
 
+        private static ConfigureContainerBuilder FindConfigureContainerDelegate(Type startupType, string environmentName)
+        {
+            var configureMethod = FindMethod(startupType, "Configure{0}Container", environmentName, typeof(void), required: false);
```
khellang: Since `required: false`, this method can return `null`, which means `ConfigureContainerBuilder` would throw.

khellang: Wait... Scratch that. I see you've made the `Invoke` method a no-op when `MethodInfo` is null.

src/Microsoft.AspNetCore.Hosting/Internal/StartupMethods.cs
```diff
@@ -9,17 +9,10 @@ namespace Microsoft.AspNetCore.Hosting.Internal
...
-        internal static Func<IServiceCollection, IServiceProvider> DefaultBuildServiceProvider = s => s.BuildServiceProvider();
-
-        public StartupMethods(Action<IApplicationBuilder> configure)
-            : this(configure, configureServices: null)
-        {
-        }
-
         public StartupMethods(Action<IApplicationBuilder> configure, Func<IServiceCollection, IServiceProvider> configureServices)
         {
             ConfigureDelegate = configure;
```
khellang: Is it worth adding preconditions to `configure` and `configureServices` here?

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
```diff
@@ -16,4 +16,25 @@ public virtual IServiceProvider ConfigureServices(IServiceCollection services)
...
+        {
+            var builder = _factory.CreateBuilder(services);
```
khellang: If you're supposed to call `Populate` on the builder before returning it from the factory, don't you have to call `ConfigureServices` so services are actually added before the builder's populated?

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
```diff
@@ -16,4 +16,25 @@ public virtual IServiceProvider ConfigureServices(IServiceCollection services)
...
+            ConfigureContainer((TContainer)builder);
```
khellang: What happens if the registered `IServiceProviderFactory`'s builder type doesn't match `TContainer`?

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
```diff
@@ -16,4 +16,25 @@ public virtual IServiceProvider ConfigureServices(IServiceCollection services)
             return services.BuildServiceProvider();
         }
     }
+
+    public abstract class StartupBase<TContainer> : IStartup
```
khellang: Isn't this `TContainerBuilder` or `TBuilder` and not the type of the container itself?

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
khellang: Nevermind. I see you've changed that to always return an instance.

src/Microsoft.AspNetCore.Hosting/ConfigureContainerBuilder.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                return;
+            }
+
+            // Only support IServiceCollection parameters
+            var parameters = MethodInfo.GetParameters();
+            if (parameters.Length != 1 ||
+                parameters.Any(p => !p.ParameterType.IsAssignableFrom(container.GetType())))
+            {
+                throw new InvalidOperationException($"The {MethodInfo.Name} method must either be parameterless or take only one parameter of type {container.GetType()}.");
```
khellang: Isn't this message wrong? The check is `parameters.Length != 1`, which means it cannot be parameterless. Having a parameterless version of this method makes no sense, IMO.

khellang: It can also be wrong in another way. Consider this scenario:

You have a `IContainerBuilder` interface and derive from `ServiceProviderFactory<IContainerBuilder>`. This class returns an instance of `ContainerBuilder`, which is an internal type.

What you really want the exception message to say is

> The ConfigureContainer method must take only one parameter of type IContainerBuilder.

But because you're using the runtime type, it'll end up saying

> The ConfigureContainer method must take only one parameter of type ContainerBuilder.

Which is slightly misleading because it's an internal type and you'll never be able to take that as a parameter.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -201,6 +201,8 @@ private IServiceCollection BuildHostingServices()
             // Conjure up a RequestServices
             services.AddTransient<IStartupFilter, AutoRequestServicesStartupFilter>();
 
+            services.AddTransient<IServiceProviderFactory, DefaultServiceProviderFactory>();
```
khellang: Since you've opted for a `ConfigureServices` call for `Use[Container]` extensions, won't this rely on ordering? Is that something that's part of the spec?

I think it would be better to move this below the `_configureServicesDelegates` code and do a `TryAddTransient`

src/Microsoft.AspNetCore.Hosting/Internal/StartupMethods.cs
davidfowl: Debug.Assert, sure. This is all "internal"

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
davidfowl: Invalid cast exception

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
davidfowl: It's the container builder

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: Yes this should do a TryAdd instead of an add. 

> Introduce a UseServiceProviderFactory method on the IWebHostBuilder interface, like I've done here

I considered this, the only question is do we want to let `IServiceProviderFactory` implementations rely on services? It's probably a bad idea anyways but you can imagine this relying on `IOptions<T>` configured in the `WebHostBuilder`.

I think I'll make it a field instead.

src/Microsoft.AspNetCore.Hosting/ConfigureContainerBuilder.cs
davidfowl: > Isn't this message wrong? The check is parameters.Length != 1, which means it cannot be parameterless. Having a parameterless version of this method makes no sense, IMO.

Copy pasta.

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
davidfowl: This should be virtual.

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
davidfowl: The base class would be in consistent with the non generic one. That might be fine in this case though.

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
davidfowl: I'm not sure about this base class though. The fact that you have to flow the factory to the base feels like a deal breaker.

khellang: You think that's the best exception we can cook up? :wink:

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
khellang: Have you considered what I mentioned in https://github.com/aspnet/Hosting/pull/833#discussion_r74199132?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: Can't add it as a field, would break the interface 😄 . Should have made the `IWebHostBuilder` an abstract base class.

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
davidfowl: We can do something here with open generics that might be nice

khellang: > The fact that you have to flow the factory to the base feels like a deal breaker.

Yeah, it sucks, but I'd still use the base class.

The only way I can think of, is to add it as a parameter to `IStartup.ConfigureServices`. You could easily "hide it" from the average user by having abstract classes that deals with it. It would break `IStartup` though :cry:

davidfowl:
🆙 📅

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -11,27 +11,66 @@ namespace Microsoft.AspNetCore.Hosting.Internal
...
-            return new StartupMethods(configureMethod.Build(instance), servicesMethod?.Build(instance));
+            var configureServicesCallback = servicesMethod.Build(instance);
+            var configureContainerCallback = configureContainerMethod.Build(instance);
+
+            Func<IServiceCollection, IServiceProvider> configureServices = services =>
+            {
+                // Call ConfigureServices, if that returned an IServiceProvider, we're done
+                IServiceProvider applicationServiceProvider = configureServicesCallback.Invoke(services);
+
+                // If there's a ConfigureContainer method
```
khellang: I think this comment is wrong...

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -11,27 +11,66 @@ namespace Microsoft.AspNetCore.Hosting.Internal
...
+                }
+
+                // If there's a ConfigureContainer method
+                if (configureContainerMethod.MethodInfo != null)
+                {
+                    // We have a ConfigureContainer method, get the IServiceProviderFactory<TContainerBuilder>
+                    var serviceProviderFactoryType = typeof(IServiceProviderFactory<>).MakeGenericType(configureContainerMethod.GetContainerType());
+                    var serviceProviderFactory = hostingServiceProvider.GetRequiredService(serviceProviderFactoryType);
+                    // var builder = serviceProviderFactory.CreateBuilder(services);
+                    var builder = serviceProviderFactoryType.GetMethod("CreateBuilder").Invoke(serviceProviderFactory, new object[] { services });
```
khellang: Use `nameof`?

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
```diff
@@ -11,9 +11,47 @@ public abstract class StartupBase : IStartup
...
+        {
+        }
+
+        IServiceProvider IStartup.ConfigureServices(IServiceCollection services)
+        {
+            ConfigureServices(services);
+            return CreateServiceProvider(services);
+        }
+
+        protected virtual IServiceProvider CreateServiceProvider(IServiceCollection services)
```
khellang: Why is this method needed?

khellang:
I like the generics :smile: At least we've eliminated the possibility to get an incompatible service provider factory injected into the startup class. If only there was a way to pass the factory to `IStartup.ConfigureServices` now, I'd be happy. Then you could simply obsolete the `BuildServiceProvider` extension method and make people use the default factory instead.

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
davidfowl: How?

khellang: `nameof(DefaultServiceProviderFactory.CreateBuilder)`?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -200,6 +201,7 @@ private IServiceCollection BuildHostingServices()
 
             // Conjure up a RequestServices
             services.AddTransient<IStartupFilter, AutoRequestServicesStartupFilter>();
+            services.AddTransient<IServiceProviderFactory<IServiceCollection>, DefaultServiceProviderFactory>();
```
Tratcher: Where is DefaultServiceProviderFactory defined?

davidfowl: In our container. Each container has their own impl.

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
davidfowl: That works if it's not static?

pakrym: Yes

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
Tratcher: Ah, over here: https://github.com/aspnet/DependencyInjection/pull/443

src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs
davidfowl: Support what used to work. I probably can't make this change anyways since it's breaking.

khellang:
:shipit: :grin:

src/Microsoft.AspNetCore.Hosting/Internal/ConfigureServicesBuilder.cs
```diff
@@ -12,42 +12,37 @@ public class ConfigureServicesBuilder
...
+                parameters.Any(p => p.ParameterType != typeof(IServiceCollection)))
             {
-                throw new ArgumentNullException(nameof(exportServices));
+                throw new InvalidOperationException("The ConfigureServices method must either be parameterless or take only one parameter of type IServiceCollection.");
             }
 
-            var parameters = new object[MethodInfo.GetParameters().Length];
+            var arguments = new object[MethodInfo.GetParameters().Length];
 
             // Ctor ensures we have at most one IServiceCollection parameter
```
halter73: Nit: The checks are no longer in the ctor.

PR #834, opened 2016-08-12
Rename to Microsoft.AspNetCore.Server.IntegrationTesting

### LINKED ISSUE #824: Consider renaming Microsoft.AspNetCore.Server.Testing
We produced versions of this package with 1.0.0-rc2-* versions and went back in time to 0.1.0 \ 0.2.0. It would be nicer to use a different name instead here.

pranavkm:
Fixes #824


pranavkm:
cc @Tratcher  \ @kichalla

Tratcher:
Where was this actually causing a problem?

This breaks ServerTests, IISIntegration, Localiazation, Entropy, etc..

pranavkm:
I can follow up with the remaining repos. The problem is that we released a version that was semantically higher and then dropped the version. We keep the 1.0.0 version around in the aspnetcore-master to build older branches. We risk accidentally pulling a super old build of the package any time that feed is referenced (typically in the master branch, patch branches).

Tratcher:
Bleh. :shipit:

PR #836, opened 2016-08-17
Remove existing implementation of StartupExceptionPage and use the on…

### LINKED ISSUE #831: Fix startup exception page to handle flattened exceptions
@Eilon looks like one more place we need to do the fixes :-(

Current behavior: https://jsfiddle.net/2s6g0xsx/ (notice that we do not display the code for other exceptions and also the `show exception details` link does not toggle as we have seen in developer exception page)

kichalla:
…e in Common

[Fixes #831] Fix startup exception page to handle flattened exceptions

@Eilon @pranavkm


NuGet.config
```diff
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
+    <add key="NufffGet" value="C:\temp\pagegeneratoroutput" />
```
kichalla: ignore this file. I will remove it before checkin

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -168,15 +170,27 @@ private RequestDelegate BuildApplication()
...
+                else
+                {
+                    model.ErrorDetails = new ExceptionDetails[] { };
+                }
+
+                var errorPage = new ErrorPage(model);
                 return context =>
                 {
                     context.Response.StatusCode = 500;
                     context.Response.Headers["Cache-Control"] = "private, max-age=0";
```
kichalla: DeveloperExceptionPage doesn't set this header. Seems like it should?

Tratcher: why not just no-cache?

Tratcher:
New screen shots, with and without details?

davidfowl:
What does this page look like? The diagnostics one?

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
kichalla: yeah, that's a question I too had...this was from earlier code..

kichalla:
Here you go
https://jsfiddle.net/xenv6d9s/1/ with ReflectionTypeLoadExceptions
https://jsfiddle.net/446cq538/1/ with exceptions in Configure method of Startup

davidfowl:
This startup page looks worse than the one we used to have 😄 . What happens when detailed errors aren't turned on?

kichalla:
:smile: 
https://jsfiddle.net/mxaw1u0s/ with detailed errors turned off

Tratcher:
You lost the detailed footer

kichalla:
Good catch. Will fix.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -168,15 +170,27 @@ private RequestDelegate BuildApplication()
...
+                {
+                    var exceptionDetailProvider = new ExceptionDetailsProvider(
+                        hostingEnv.ContentRootFileProvider,
+                        sourceCodeLineCount: 6);
+
+                    model.ErrorDetails = exceptionDetailProvider.GetDetails(ex);
+                }
+                else
+                {
+                    model.ErrorDetails = new ExceptionDetails[] { };
```
Eilon: `new ExceptionDetails[0]`

src/Microsoft.AspNetCore.Hosting/Properties/Resource.Designer.cs
```diff
@@ -0,0 +1,78 @@
+// <auto-generated />
+namespace Microsoft.AspNetCore.Hosting
+{
+    using System.Globalization;
+    using System.Reflection;
+    using System.Resources;
+
+    internal static class Resource
```
Eilon: Difference between Resource.Designer.cs and Resources.Designer.cs?

Eilon:
🕐

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -168,15 +170,27 @@ private RequestDelegate BuildApplication()
...
                 var hostingEnv = _applicationServices.GetRequiredService<IHostingEnvironment>();
                 var showDetailedErrors = hostingEnv.IsDevelopment() || _options.DetailedErrors;
-                var errorBytes = StartupExceptionPage.GenerateErrorHtml(showDetailedErrors, ex);
 
+                var model = new ErrorPageModel();
+                if (showDetailedErrors)
+                {
+                    var exceptionDetailProvider = new ExceptionDetailsProvider(
+                        hostingEnv.ContentRootFileProvider,
+                        sourceCodeLineCount: 6);
```
troydai: Is the 6 a magic number or convention? And it is not configurable here.

Eilon: Nothing magical. It's just the same number we use everywhere. No need to make it configurable.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
pranavkm: Alternatively, could you initialize this to an empty sequence in the `ErrorModel` type? No risk of accidental null refs.

src/Microsoft.AspNetCore.Hosting/Startup/ExceptionPage/Views/ErrorPage.cs
```diff
@@ -0,0 +1,616 @@
+namespace Microsoft.AspNetCore.Hosting.Views
```
pranavkm: Could you make the change to the generator to name this file ErrorPage.Generated.cs? If that suppresses diffs showing up in GitHub, that would be great.

src/Microsoft.AspNetCore.Hosting/Startup/ExceptionPage/Views/ErrorPage.js
```diff
@@ -0,0 +1,192 @@
+﻿(function (window, undefined) {
```
pranavkm: :cry: could we not copy in all of the terribleness from Diagnostics? This is a mini jquery impl and we don't need all of it.

Eilon: Unless it's a crazy amount of work to make it vary by repo... I'm not *too* worried if there's some extra stuff in here...

src/Microsoft.AspNetCore.Hosting/Properties/Resource.Designer.cs
kichalla: Yeah we do not need it. Added by mistake

kichalla:
:up: :date:

src/Microsoft.AspNetCore.Hosting/Startup/ExceptionPage/Views/ErrorPage.cshtml
```diff
@@ -156,5 +156,8 @@
             <%$ include: ErrorPage.js %>
             //-->
         </script>
-    </body>
+        <footer>
```
kichalla: @Eilon  Footer here is after the `<script>` tag because our razor tool generator modifies the file in a way which generates incorrect .cs file (the escaping of the double quotes get messed up...snapshot below)...probably should file a separate issue for it?

image

src/Microsoft.AspNetCore.Hosting/Startup/ExceptionPage/Views/ErrorPage.Generated.cs
```diff
@@ -609,7 +609,37 @@ public ErrorPage()
 #line default
```
pranavkm: Ok, figured out. The files need to be named either .designer.cs or .feature.cs (Per https://github.com/github/linguist/blob/c471990aa3ced30bb6edb25c3a3530ab90066a36/lib/linguist/generated.rb#L203 \ https://github.com/github/linguist/blob/c471990aa3ced30bb6edb25c3a3530ab90066a36/lib/linguist/generated.rb#L214) which would cause GitHub to suppress diffs for it by default.

.Designer.cs might be better name since .Feature doesn't necessarily make a lot of sense.

kichalla:
:up: :date:

pranavkm:
:shipit:

kichalla:
83e0d4798eb4fd9113d06c54bba9d69faefa3d96

PR #837, opened 2016-08-18
Add environment variables to dotnet publish

pranavkm:


pranavkm:
cc @kichalla \ @Tratcher 

This would allow us to redirect the NUGET_PACKAGES as part of publish.

Tratcher:
:shipit: Though it seems odd to apply the same variables to disparate processes. Ideally you'd have different collections for the different processes.

pranavkm:
Sure, I could add a `PublishEnvironmentVariables` or something like that if you'd prefer it.

Tratcher:
Yes please, that would be cleaner.

pranavkm:
:up: :date:

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
```diff
@@ -123,7 +126,7 @@ protected void ShutDownIfAnyHostProcess(Process hostProcess)
             }
         }
 
-        protected void AddEnvironmentVariablesToProcess(ProcessStartInfo startInfo)
+        protected void AddEnvironmentVariablesToProcess(ProcessStartInfo startInfo, List<KeyValuePair<string, string>> environmentVariables)
```
Tratcher: you forgot to use this paramter

pranavkm: Oops! Fixed

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
```diff
@@ -134,7 +137,7 @@ protected void AddEnvironmentVariablesToProcess(ProcessStartInfo startInfo)
 
             SetEnvironmentVariable(environment, "ASPNETCORE_ENVIRONMENT", DeploymentParameters.EnvironmentName);
```
Tratcher: Should this be moved?

pranavkm: Doesn't affect the publish process since this is pretty AspNet specific. Plus it avoids duplicating this in two places.

Tratcher:
:shipit:

PR #840, opened 2016-08-19
Make dotnet publish use LogInformation instead of LogTrace

pranavkm:
This would make it have the same LogLevel as dotnet run so we'd have better diagnostics without having to change tests.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
@@ -69,7 +69,7 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
 
             var hostProcess = new Process() { StartInfo = startInfo };
             hostProcess.ErrorDataReceived += (sender, dataArgs) => { Logger.LogWarning(dataArgs.Data ?? string.Empty); };
-            hostProcess.OutputDataReceived += (sender, dataArgs) => { Logger.LogTrace(dataArgs.Data ?? string.Empty); };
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @Tratcher \ @kichalla 

Most tests set the LogLevel to Information:

* https://github.com/aspnet/MusicStore/blob/dev/test/E2ETests/PublishAndRunTests.cs#L78
* https://github.com/aspnet/Entropy/blob/dev/test/FunctionalTests/TestServices.cs#L39

which means we (a) it looks like the test is hanging while it publishes (b) we don't get detailed publish logs by default.

Tratcher:
We intentionally lowered the level of some of these items to reduce log spam.
- How long does publish take?
- How much output does it generate?

pranavkm:
Here's the output for MusicStore:

```
Publishing MusicStore for .NETCoreApp,Version=v1.0
Project MusicStore (.NETCoreApp,Version=v1.0) was previously compiled. Skipping compilation.
Configuring the following project for use with IIS: 'D:\work\katana\MusicStore\samples\MusicStore\bin\Debug\netcoreapp1.0\publish'
Updating web.config at 'D:\work\katana\MusicStore\samples\MusicStore\bin\Debug\netcoreapp1.0\publish\web.config'
Configuring project completed successfully
publish: Published to D:\work\katana\MusicStore\samples\MusicStore\bin\Debug\netcoreapp1.0\publish
Published 1/1 projects successfully
```

It takes about 2.5 seconds on netcoreapp1.0 and about 1.5 on desktop. Doesn't seem super noisy.

Tratcher:
A 2.5 second pause makes you think it hung?

pranavkm:
In precompilation's case, this tends to be a lot longer (~5-6 seconds). Plus it's not about the pause in itself. The log prints spews from our tools, so it's pretty useful to diagnose in case they run in to issues.

Tratcher:
:shipit:

PR #841, opened 2016-08-19
Fix pinning of Microsoft.AspNetCore.Server.Testing

anurse:
Microsoft.AspNetCore.Server.Testing got mis-pinned in the master pinning. The released package was rolled back to 0.1.0 but ended up getting pinned at 1.0.0. This will mess with our ability to rebuild master, so we should remedy this.

**NOTE**: This is a PR to `master`, so it is changing our 1.0.0 release code. The change is not relevant to `dev` because the package has been renamed.

/cc @pranavkm @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Server.Testing/project.json
+++ b/src/Microsoft.AspNetCore.Server.Testing/project.json
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0-rc2-20896",
+  "version": "0.1.0-rtm-21431",
   "buildOptions": {
     "warningsAsErrors": true,
     "keyFile": "../../tools/Key.snk",


pranavkm:
:shipit: 

Could you also update the `1.0.0` tag in this repo?

anurse:
After merging, I'll do that. I do want @Eilon to see and approve this before I merge just to be sure it gets seen. There's no functional code change here, it's just a change to make sure we can build things properly in this branch moving forwards.

anurse:
Travis failure is because of the OpenSSL issue. We need to do a batch-update of our master travis.yml files to fix this, so I'm going to suggest holding off on that.

Eilon:
:shipit:

anurse:
`1.0.0` tag updated.

PR #850, opened 2016-09-08
Add retry to ApplicationDeployer delete

BrennanConroy:
@Tratcher @JunTaoLuo @victorhurdugaci

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/RetryHelper.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/RetryHelper.cs
@@ -72,5 +72,27 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
             logger.LogInformation("Failed to connect, retry limit exceeded.");
             throw new OperationCanceledException("Failed to connect, retry limit exceeded.");
         }
+
+        public static void RetryOperation(
+            Action retryBlock,
+            Action<Exception> exceptionBlock,
+            int retryCount = 3,
+            int retryDelayMilliseconds = 0)
+        {
+            for (var retry = 0; retry < retryCount; ++retry)
+            {
+                try
+                {
+                    retryBlock();
+                    break;
+                }
+                catch (Exception exception)
+                {
+                    exceptionBlock(exception);
+                }
+
+                Thread.Sleep(retryDelayMilliseconds);
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
```diff
@@ -90,15 +90,21 @@ protected void CleanPublishedOutput()
             }
             else
             {
-                try
```
JunTaoLuo: Extract this into a helper function?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/RetryHelper.cs
```diff
@@ -72,5 +72,24 @@ public class RetryHelper
...
+        public static void RetryOperation(
+            Action retryBlock,
+            Action<Exception> exceptionBlock,
+            int retryCount = 3)
+        {
+            for (var retry = 0; retry < retryCount; ++retry)
+            {
+                try
+                {
+                    retryBlock();
```
JunTaoLuo: Might want to add a slight delay between retries

JunTaoLuo:
:shipit:

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/RetryHelper.cs
```diff
@@ -72,5 +72,27 @@ public class RetryHelper
...
+            int retryDelayMilliseconds = 0)
+        {
+            for (var retry = 0; retry < retryCount; ++retry)
+            {
+                try
+                {
+                    retryBlock();
+                    break;
+                }
+                catch (Exception exception)
```
victorhurdugaci: Don't handle the exception here. Let the caller decide if they want or not to handle it and make `retryBlock` return a boolean that tells the caller if it should retry or not

JunTaoLuo: Actually why not just throw the exception when the retry limit has been reached?

JunTaoLuo: Which is what the RetryRequest does.

victorhurdugaci: I'm talking about something like this: https://github.com/aspnet/Logging/blob/e9e62d27d3719fe0abbb94ca93f5c2bc2d80fffd/test/Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Test/RetryHelper.cs

PR #853, opened 2016-09-12
[Fixes #852] TestHost: OnStarting and OnCompleted callbacks of response are not being awaited

### LINKED ISSUE #852: TestHost: OnStarting and OnCompleted callbacks of response are not being awaited
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.TestHost/ResponseFeature.cs#L42

Since they are not being awaited, the code within these call backs get executed at random times causing unexpected failures (ex: In the delegate accessing session on HttpContext would fail if the code gets executed way late by which the session middleware has removed the session from the context).

From @Tratcher : These call backs' signature used to be synchronous earlier and was later changed to async, but we missed changing the call sites to consider the new async behavior.

cc @rynowak

kichalla:
@Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
+++ b/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
@@ -81,7 +81,7 @@ namespace Microsoft.AspNetCore.TestHost
                     try
                     {
                         await _application.ProcessRequestAsync(state.Context);
-                        state.CompleteResponse();
+                        await state.CompleteResponseAsync();
                         state.ServerCleanup(exception: null);
                     }
                     catch (Exception ex)
@@ -165,7 +165,7 @@ namespace Microsoft.AspNetCore.TestHost
                     }
                 }
 
-                _responseStream = new ResponseStream(ReturnResponseMessage, AbortRequest);
+                _responseStream = new ResponseStream(ReturnResponseMessageAsync, AbortRequest);
                 httpContext.Response.Body = _responseStream;
                 httpContext.Response.StatusCode = 200;
                 httpContext.RequestAborted = _requestAbortedSource.Token;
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
```diff
@@ -187,27 +187,26 @@ internal void AbortRequest()
...
             }
 
-            internal void ReturnResponseMessage()
+            internal async Task ReturnResponseMessageAsync()
             {
                 if (!_responseTcs.Task.IsCompleted)
                 {
-                    var response = GenerateResponse();
-                    // Dispatch, as TrySetResult will synchronously execute the waiters callback and block our Write.
-                    Task.Factory.StartNew(() => _responseTcs.TrySetResult(response));
```
kichalla: I did not see any issues running tests with this change.

Tratcher: Please revert this, calling SetResult inline can cause threading issues and hangs

src/Microsoft.AspNetCore.TestHost/ResponseFeature.cs
```diff
@@ -36,33 +36,42 @@ public ResponseFeature()
...
+            var prior = _responseCompletedAsync;
+            _responseCompletedAsync = async () =>
             {
-                callback(state);
-                prior();
+                try
+                {
+                    await callback(state);
+                }
+                finally { }
```
Tratcher: this try/finally doesn't do anything, your exception gets rethrown after the finally. `await prior();` should be in the finally block

Tratcher: In a normal server we'd catch all the exceptions and log them, but here we stand a chance of surfacing them to the client. See if you can make one throw all the way back to the client.

src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
```diff
@@ -187,27 +187,33 @@ internal void AbortRequest()
...
+                if (!Context.HttpContext.Response.HasStarted)
                 {
-                    var response = GenerateResponse();
-                    // Dispatch, as TrySetResult will synchronously execute the waiters callback and block our Write.
-                    Task.Factory.StartNew(() => _responseTcs.TrySetResult(response));
+                    if (!_responseTcs.Task.IsCompleted)
+                    {
+                        var response = await GenerateResponseAsync();
+                        // Dispatch, as TrySetResult will synchronously execute the waiters callback and block our Write.
+                        var setResult = Task.Factory.StartNew(() => _responseTcs.TrySetResult(response));
```
campersau: Looks like `setResult` is unused?

Tratcher: It's set to avoid a compiler warning about not awaiting the task.

kichalla:
:up: :date:

src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
```diff
@@ -187,27 +187,33 @@ internal void AbortRequest()
...
-                if (!_responseTcs.Task.IsCompleted)
+                // Check if the response has already started because the TrySetResult below could happen a bit late
+                // (as it happens on a different thread) by which point the CompleteResponseAsync could run and calls this
+                // method again.
+                if (!Context.HttpContext.Response.HasStarted)
                 {
-                    var response = GenerateResponse();
-                    // Dispatch, as TrySetResult will synchronously execute the waiters callback and block our Write.
-                    Task.Factory.StartNew(() => _responseTcs.TrySetResult(response));
+                    if (!_responseTcs.Task.IsCompleted)
```
Tratcher: You shouldn't need to check both HasStarted and IsCompleted. HasStarted should be sufficient.

kichalla: Sounds good. :up: :date:

PR #856, opened 2016-09-27
Fixed naming abiguity

mikaelm12:
Fixing a coherence failure

`'HttpMethod' is an ambiguous reference between 'Microsoft.AspNetCore.Http.HttpMethod' and 'System.Net.Http.HttpMethod'`

@troydai @Tratcher


troydai:
:shipit: 

You can merge it now since it will fix the CI break.

PR #860, opened 2016-10-07
Updated IISDeployer

jhkimnew:
@pranavkm @Tratcher Please review this PR and approve if you agree so that I can merge this code change on my fork to dev branch. 

Brief descripton of the issue: 

IIS Deployer has an issue regarding the applicationPath pointing wrong directory instead of publish directory and so I fixed it following the way used by IISExpress Deployer and confirmed the issue is fixed.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISDeployer.cs
@@ -36,13 +36,15 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
 
             _application = new IISApplication(DeploymentParameters, Logger);
 
-            // Publish to IIS root\application folder.
-            DotnetPublish(publishRoot: _application.WebSiteRootFolder);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISDeployer.cs
```diff
@@ -36,13 +36,15 @@ public override DeploymentResult Deploy()
 
             _application = new IISApplication(DeploymentParameters, Logger);
 
-            // Publish to IIS root\application folder.
-            DotnetPublish(publishRoot: _application.WebSiteRootFolder);
+            if (DeploymentParameters.PublishApplicationBeforeDeployment)
```
Tratcher: Don't you always have to publish to run? I think we always publish for express.

jhkimnew: DeploymentParameters.PublishApplicationBeforeDeployment is set to true at the line 35. 
So, adding this if condition would not affect any change. IISExpress has the same code logic.

Issue #867, opened 2016-10-24
Object disposal issues in WebHostBuilder


pakrym:
Hosting service provider is never disposed.

When https://github.com/aspnet/DependencyInjection/issues/459 is fixed all services we register as instances need to be disposed in `WebHost`

muratg:
@pakrym Triaging this for 1.1 since you already have the PR.

pakrym:
https://github.com/aspnet/Hosting/pull/868

PR #868, opened 2016-10-25
Deterministically dispose instances created by WebHostBuilder

pakrym:
Fixes: https://github.com/aspnet/Hosting/issues/867
IServiceProvider would stop disposing instances: https://github.com/aspnet/DependencyInjection/pull/462

@davidfowl @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/ServiceCollectionExtensions.cs
@@ -1,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal static class ServiceCollectionExtensions
+    {
+        public static IServiceCollection Clone(this IServiceCollection serviceCollection)
+        {
+            IServiceCollection clone = new ServiceCollection();
+            foreach (var service in serviceCollection)
+            {
+                clone.Add(service);
+            }
+            return clone;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -31,6 +31,7 @@ namespace Microsoft.AspNetCore.Hosting
         private IConfiguration _config;
         private ILoggerFactory _loggerFactory;
         private WebHostOptions _options;
+        private bool _webHostBuilt;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WebHostBuilder"/> class.
@@ -48,7 +49,7 @@ namespace Microsoft.AspNetCore.Hosting
             if (string.IsNullOrEmpty(GetSetting(WebHostDefaults.EnvironmentKey)))
             {
                 // Try adding legacy environment keys, never remove these.
-                UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("Hosting:Environment") 
+                UseSetting(WebHostDefaults.EnvironmentKey, Environment.GetEnvironmentVariable("Hosting:Environment")
                     ?? Environment.GetEnvironmentVariable("ASPNET_ENV"));
             }
 
@@ -135,6 +136,12 @@ namespace Microsoft.AspNetCore.Hosting
         /// </summary>
         public IWebHost Build()
         {
+            if (_webHostBuilt)
+            {
+                throw new InvalidOperationException(Resources.WebHostBuilder_SingleInstance);
+            }
+            _webHostBuilt = true;
+
             // Warn about deprecated environment variables
             if (Environment.GetEnvironmentVariable("Hosting:Environment") != null)
             {
@@ -151,17 +158,24 @@ namespace Microsoft.AspNetCore.Hosting
                 Console.WriteLine("The environment variable 'ASPNETCORE_SERVER.URLS' is obsolete and has been replaced with 'ASPNETCORE_URLS'");
             }
 
-            var hostingServices = BuildHostingServices();
-            var hostingContainer = hostingServices.BuildServiceProvider();
+            var hostingServices = BuildCommonServices();
+            var applicationServices = hostingServices.Clone();
+            var hostingServiceProvider = hostingServices.BuildServiceProvider();
 
-            var host = new WebHost(hostingServices, hostingContainer, _options, _config);
+            AddApplicationServices(applicationServices, hostingServiceProvider);
+
+            var host = new WebHost(
+                applicationServices,
+                hostingServiceProvider,
+                _options,
+                _config);
 
             host.Initialize();
 
             return host;
         }
 
-        private IServiceCollection BuildHostingServices()
+        private IServiceCollection BuildCommonServices()
         {
             _options = new WebHostOptions(_config);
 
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -195,9 +195,8 @@ private IServiceCollection BuildHostingServices()
             services.AddTransient<IHttpContextFactory, HttpContextFactory>();
             services.AddOptions();
 
-            var diagnosticSource = new DiagnosticListener("Microsoft.AspNetCore");
-            services.AddSingleton<DiagnosticSource>(diagnosticSource);
-            services.AddSingleton<DiagnosticListener>(diagnosticSource);
+            services.AddSingleton<DiagnosticListener>(provider => new DiagnosticListener("Microsoft.AspNetCore"));
```
davidfowl: Does this need to be disposed?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -257,5 +256,10 @@ private string ResolveContentRootPath(string contentRootPath, string basePath)
             }
             return Path.Combine(Path.GetFullPath(basePath), contentRootPath);
         }
+
+        public void Dispose()
+        {
+            _loggerFactory?.Dispose();
```
davidfowl: Shouldn't this be disposed with the host? Why is the builder disposable?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -195,9 +195,8 @@ private IServiceCollection BuildHostingServices()
             services.AddTransient<IHttpContextFactory, HttpContextFactory>();
             services.AddOptions();
 
-            var diagnosticSource = new DiagnosticListener("Microsoft.AspNetCore");
-            services.AddSingleton<DiagnosticSource>(diagnosticSource);
-            services.AddSingleton<DiagnosticListener>(diagnosticSource);
+            services.AddSingleton<DiagnosticListener>(provider => new DiagnosticListener("Microsoft.AspNetCore"));
+            services.AddSingleton<DiagnosticSource>(provider => provider.GetService<DiagnosticListener>());
```
davidfowl: Not sure I like this.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
pakrym: It is IDisposible

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
pakrym: No, Builder creates it and will reuse for any number of WebHosts you build using it.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
pakrym: We don't have multi interface registration and there is no need to have this as an instance as far as I understand.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -22,7 +22,7 @@ namespace Microsoft.AspNetCore.Hosting
     /// <summary>
     /// A builder for <see cref="IWebHost"/>
     /// </summary>
-    public class WebHostBuilder : IWebHostBuilder
+    public class WebHostBuilder : IWebHostBuilder, IDisposable
```
Tratcher: Should IWebHostBuilder be IDisposable?

pakrym:
1. Changed WebHostBuilder to allow creating only single instance of WebHost.
2. LoggerFactory is now disposed with WebHost is it was automatically created.
3. If LoggerFactory was passed in it's not disposed.
4. Added tests for all these cases.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
pakrym: then we would need two different `BuildHostingServices` or at least parts that need to chain.

pakrym: Also we don't mutate `DiagnosticSource` so it might be fine to return different instances.

pakrym:
Updated again, now there is two methods to add host services and to forward them to application services. Also fixed server disposal issue.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -245,6 +262,17 @@ private IServiceCollection BuildHostingServices()
...
+        private void AddApplicationServices(IServiceCollection services, IServiceProvider hostingServiceProvider)
+        {
+            var loggerFactory = hostingServiceProvider.GetService<ILoggerFactory>();
+            services.Replace(ServiceDescriptor.Singleton(typeof(ILoggerFactory), loggerFactory));
+
+            var listener = hostingServiceProvider.GetService<DiagnosticListener>();
+            services.Replace(ServiceDescriptor.Singleton(typeof(DiagnosticListener), listener));
+            services.Replace(ServiceDescriptor.Singleton(typeof(DiagnosticSource), listener));
+        }
+
```
davidfowl: nit: extra space

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -245,6 +262,17 @@ private IServiceCollection BuildHostingServices()
             return services;
         }
 
+        private void AddApplicationServices(IServiceCollection services, IServiceProvider hostingServiceProvider)
+        {
+            var loggerFactory = hostingServiceProvider.GetService<ILoggerFactory>();
```
davidfowl: Add comments here explaining why thungs are so bizzare

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -175,30 +189,33 @@ private IServiceCollection BuildHostingServices()
             var services = new ServiceCollection();
             services.AddSingleton(_hostingEnvironment);
 
+            //The configured ILoggerFactory is added as a singleton here. AddLogging below will not add an additional one.
```
davidfowl: nit: space after //

pakrym:
Also undid change to `UseServer` and fixed tests for it.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -188,6 +189,7 @@ public void MultipleConfigureLoggingInvokedInOrder()
                 .UseStartup<StartupNoServices>();
 
             var host = (WebHost)hostBuilder.Build();
+            var factory = host.Services.GetService<ILoggerFactory>();
```
davidfowl: What is this for?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -494,6 +496,86 @@ public void Configure_SupportsStaticMethodDelegate()
...
+
+            var host = builder.UseServer(server)
+                .ConfigureServices(collection => collection.AddSingleton<ILoggerFactory>(factory))
+                .UseStartup<StartupWithILoggerFactory>()
+                .Build();
+
+            var factoryFromHost = host.Services.GetService<ILoggerFactory>();
+            Assert.Equal(factory, factoryFromHost);
+        }
+
```
davidfowl: nit: extra spaces

pakrym: Once `ILoggerFactory` was registered as factory so resolve was required to invoke callbacks. Not true anymore. Reverting.

src/Microsoft.AspNetCore.Hosting/Internal/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    public static class ServiceCollectionExtensions
```
muratg: Actually making the class internal?

PR #874, opened 2016-10-31
Patch Hosting to 1.0.1

JunTaoLuo:
Not actually merging this PR but just want to verify the changes we want for 1.0.3 release. The patch will include the fix for #812 and the related version updates. Note that I'm comparing the changes against master since I can't compare against the 1.0.0 tag but they are currently equivalent.

cc @pranavkm @muratg @Tratcher

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ branches:
     - dev
     - /^(.*\/)?ci-.*$/
 before_install:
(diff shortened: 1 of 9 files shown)


JunTaoLuo:
Will also wait for patch approval.

davidfowl:
Something broken on travis

JunTaoLuo:
I have updated path 1.0.3 to build with Hosting 1.0.1

Issue #871, opened 2016-10-31
Add methods to IApplicationLifetime for firing all events


davidfowl:
Today [IApplicationLifetime](https://github.com/aspnet/Hosting/blob/0f1eac5a98c14e63c8528a3e7ab11c61552ba6a6/src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetime.cs) has 3 events, ApplicationStarted, ApplicationStopping and ApplicationStopped. It also has a way to Stop the application (StopApplication). In order to make this service self contained and replaceable, we should have a way to fire the individual events as well as handle them. That way it would be possible for hosting to use a user provided implementation of `IApplicationLifetime` to trigger the hosting shutdown sequence. See https://github.com/aspnet/Hosting/issues/812#issuecomment-256917236

```C#
public interface IApplicationLifetime
{
    CancellationToken ApplicationStarted { get; }

    CancellationToken ApplicationStopping { get; }

    CancellationToken ApplicationStopped { get; }

    void FireApplicationStarted();

    void FireApplicationStopping();

    void FireApplicationStopped();

    void StopApplication();
}
```

Not sure about the names of things, could be Fire/Trigger or any other verb. This would be 2.0 since it's a breaking change

davidfowl:
Alternatively, we might want to change to a model where you have to implement something like an `IApplicationLifetimeEvents` to get notified of events instead of us exposing cancellation tokens:

```C#
public interface IApplicationLifetimeEvents
{
    void OnApplicationStarted();
    void OnApplicationStopping();
    void OnApplicationStopped();
}
```

We would need some place to put the StopApplication as it wouldn't be overridable in this model. We could leave it on `IApplicationLifetime`.

```C#
public class IApplicationLifetime
{
    void StopApplication();
}
```

Use code would implement `IApplicationLifetimeEvents` and handle the methods:

```C#
public class MyHandler : IApplicationLifetimeEvents
{
    public void OnApplicationStarted()
    {
    }

    public void OnApplicationStopping()
    {
    }

    public void OnApplicationStopped()
    {
    }
}
```

Hosting would resolve an `IEnumerable<IApplicationLifetimeEvents>` and invoke the callbacks during the shutdown sequence.

You could even implement the interface on your Startup class...

```C#
public class Startup : IApplicationLifetimeEvents
{
   public void ConfigureServices(IServiceCollection services) 
   { 
        // Add the startup class as an event handler
[...]

PR #875, opened 2016-11-01
[WIP]: Spike IApplicationLifetimeEvents

davidfowl:
- Introduce a new DI friendly API for handling lifetime events. IApplicationLifetime isn't replaceable so we introduce a new DI friendly API that can be implemented to handle
lifetime events of an ASP.NET application. It should also make it possible to write up extension
on IWebHostBuilder to wire up to external systems that need to register the state of the application (like systemd).
- Need to add tests

Here's what the demo application looks like:

```C#
public class Startup
{
    // This method gets called by the runtime. Use this method to add services to the container.
    // For more information on how to configure your application, visit http://go.microsoft.com/fwlink/?LinkID=398940
    {
        services.AddSingleton<IApplicationLifetimeEvents, Events>();
    }

    // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
    public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
    {
        loggerFactory.AddConsole();

        if (env.IsDevelopment())
        {
            app.UseDeveloperExceptionPage();
        }

        app.Run(async (context) =>
        {
            await context.Response.WriteAsync("Hello World!");
        });
    }
}

public class Events : IApplicationLifetimeEvents
{
    private readonly IApplicationLifetime _lifetime;

    public Events(IApplicationLifetime lifetime)
    {
        _lifetime = lifetime;
    }

    public void OnApplicationStarted()
    {
        Console.WriteLine("Application started");
    }

    public void OnApplicationStopped()
    {
        // Disposable things have been disposed by the time we get here.
        Console.WriteLine("Application stopped");
    }

    public void OnApplicationStopping()
    {
        Console.WriteLine("Application stopping");
    }
}
```

#871 

/cc @Tratcher @pakrym @muratg

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
@@ -1,5 +1,11 @@
-﻿namespace Microsoft.AspNetCore.Hosting
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Hosting
 {
+    /// <summary>
+    /// Allows consumers to perform cleanup during a graceful shutdown.
+    /// </summary>
     public interface IApplicationLifetimeEvents
     {
         /// <summary>
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
@@ -1,0 +1,25 @@
+﻿namespace Microsoft.AspNetCore.Hosting
+{
+    public interface IApplicationLifetimeEvents
+    {
+        /// <summary>
+        /// Triggered when the application host has fully started and is about to wait
+        /// for a graceful shutdown.
+        /// </summary>
+        void OnApplicationStarted();
+
+        /// <summary>
+        /// Triggered when the application host is performing a graceful shutdown.
+        /// Requests may still be in flight. Shutdown will block until this event completes.
+        /// </summary>
+        void OnApplicationStopping();
+
+        /// <summary>
+        /// Triggered when the application host is performing a graceful shutdown.
+        /// All requests should be complete at this point. Shutdown will block
+        /// until this event completes.
+        /// </summary>
+        void OnApplicationStopped();
+
+    }
+}
--- a/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
@@ -20,11 +20,6 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         private readonly IEnumerable<IApplicationLifetimeEvents> _handlers = Enumerable.Empty<IApplicationLifetimeEvents>();
         private readonly ILogger<ApplicationLifetime> _logger;
 
-        public ApplicationLifetime(ILogger<ApplicationLifetime> logger)
-        {
-            _logger = logger;
-        }
-
         public ApplicationLifetime(ILogger<ApplicationLifetime> logger, IEnumerable<IApplicationLifetimeEvents> handlers)
         {
             _logger = logger;
@@ -69,12 +64,12 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
                 try
                 {
+                    _stoppingSource.Cancel(throwOnFirstException: false);
+
                     foreach (var handler in _handlers)
                     {
                         handler.OnApplicationStopping();
                     }
-
-                    _stoppingSource.Cancel(throwOnFirstException: false);
                 }
                 catch (Exception ex)
                 {
@@ -92,12 +87,12 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             try
             {
+                _startedSource.Cancel(throwOnFirstException: false);
+
                 foreach (var handler in _handlers)
                 {
                     handler.OnApplicationStarted();
                 }
-
-                _startedSource.Cancel(throwOnFirstException: false);
             }
             catch (Exception ex)
             {
@@ -114,12 +109,12 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             try
             {
+                _stoppedSource.Cancel(throwOnFirstException: false);
+
                 foreach (var handler in _handlers)
                 {
                     handler.OnApplicationStopped();
                 }
-
-                _stoppedSource.Cancel(throwOnFirstException: false);
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
@@ -55,18 +55,9 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             // which will no-op since the first call already requested cancellation, get a chance to execute.
             lock (_stoppingSource)
             {
-                // Noop if this is already cancelled, user code can call this so 
-                // we should guard against multiple calls here
-                if (_stoppingSource.IsCancellationRequested)
-                {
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
```diff
@@ -14,6 +17,19 @@ public class ApplicationLifetime : IApplicationLifetime
...
         private readonly CancellationTokenSource _stoppedSource = new CancellationTokenSource();
+        private readonly IEnumerable<IApplicationLifetimeEvents> _handlers = Enumerable.Empty<IApplicationLifetimeEvents>();
+        private readonly ILogger<ApplicationLifetime> _logger;
+
+        public ApplicationLifetime(ILogger<ApplicationLifetime> logger)
+        {
+            _logger = logger;
+        }
+
+        public ApplicationLifetime(ILogger<ApplicationLifetime> logger, IEnumerable<IApplicationLifetimeEvents> handlers)
```
pakrym: Why have 2 constructors?

src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -69,6 +69,23 @@ public static void ApplicationError(this ILogger logger, Exception exception)
                 exception: exception);
         }
 
+        public static void ApplicationError(this ILogger logger, EventId eventId, string message, Exception exception)
+        {
+            var reflectionTypeLoadException = exception as ReflectionTypeLoadException;
+            if (reflectionTypeLoadException != null)
```
pakrym: Extract to a method.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -101,6 +99,7 @@ public virtual void Start()
         {
             Initialize();
 
+            _applicationLifetime = _applicationServices.GetRequiredService<IApplicationLifetime>() as ApplicationLifetime;
```
pakrym: Is it better to throw here if IApplicationLifetime is not ApplicationLifetime. Hosting would not be able to use it anyway.

src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Microsoft.AspNetCore.Hosting
```
JunTaoLuo: copyright

src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    /// <summary>
+    /// Allows consumers to perform cleanup during a graceful shutdown.
+    /// </summary>
+    public interface IApplicationLifetimeEvents
```
davidfowl: Should this be an abstract base class? If we ever add events we might need to evolve this. On the other hand, it should probably match the `IApplicationLifetime`.

src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
```diff
@@ -44,13 +55,27 @@ public void StopApplication()
...
+                if (_stoppingSource.IsCancellationRequested)
+                {
+                    return;
+                }
+
                 try
                 {
                     _stoppingSource.Cancel(throwOnFirstException: false);
+
+                    foreach (var handler in _handlers)
```
davidfowl: How should we handle errors thrown here? Do we want 2 layers of try catch? If errors happen invoking cancellation token callbacks, should events get called?

src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
pakrym: Maybe have multiple interfaces? One per each event?

src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
Tratcher: Note that `_stoppingSource.Cancel(throwOnFirstException: false)` makes sure every registered callback gets called and then throws an AggregateException with any errors. OnApplicationStopping should do the same.

src/Microsoft.AspNetCore.Hosting.Abstractions/IApplicationLifetimeEvents.cs
davidfowl: Might be overkill?

pakrym: Single method with event enum?

davidfowl: forcing people to write if statements 👎 . I rather have 3 interfaces than an enum...

pakrym: I think if we would add other event we would group them into different interface and everything would be fine.

davidfowl: Yep. The other problem with n interfaces today is the fact that our DI container doesn't let you do this:

services.AddSingleton<IApplicationLifetimeStarted, IApplicationLifetimeStopping, IApplicationLifetimeStopped, Thing>();

davidfowl:
@Tratcher @pakrym 🆙 📅

src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
```diff
@@ -115,9 +102,32 @@ public void NotifyStopped()
...
             List<Exception> exceptions = null;
+
+            try
+            {
+                // Run the cancellation token callbacks
+                cancel.Cancel(throwOnFirstException: false);
+            }
+            catch (Exception ex)
+            {
+                if (exceptions == null)
```
pakrym: nit: this code does not run often, we can handle 3 lists per life of application.

davidfowl:
@DamianEdwards @Eilon @muratg @glennc Can you guys review this feature next week. It's gone through enough rounds and I think it's ready to merge at this point.

/cc @Tratcher @pakrym

Eilon:
This all looks good to me. Any reason not to have some mechanism for fatal errors to be notified through this mechanism? Would some listeners of the Stopped event like to know that the server was stopped non-gracefully?

davidfowl:
> Any reason not to have some mechanism for fatal errors to be notified through this mechanism?

What do fatal errors look like? Is it a new event? Today we have these 3 events and we're just re-surfacing them in a different way.

> Would some listeners of the Stopped event like to know that the server was stopped non-gracefully?

Not sure, maybe but I'm not sure we know if the shutdown was graceful or not.

Eilon:
The doc says this:

```c#
        /// <summary> 
        /// Triggered when the application host is performing a graceful shutdown. 
        /// All requests should be complete at this point. Shutdown will block 
        /// until this event completes. 
        /// </summary> 
        void OnApplicationStopped(); 
```

Specifically, it says it's called only if it's graceful.

Who might typically subscribe to this event, and why? And would they care about non-graceful shutdowns?

But anyway, perhaps my question is outside the scope of this PR...

davidfowl:
> Who might typically subscribe to this event, and why? And would they care about non-graceful shutdowns?

The same people that use `IRegisteredObject` today, though at this stage, all of your services have been disposed, so I'm not sure what else you can do here.

muratg:
Does this need more tests?

davidfowl:
@muratg what tests?

muratg:
@davidfowl I see that you added a few tests in the latest commit. Are those enough, or is this still a WIP?

davidfowl:
It's no longer a wip

PR #876, opened 2016-11-01
Handle SIGTERMs for graceful shutdown

pakrym:


src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -16,18 +19,26 @@ public static class WebHostExtensions
         /// <param name="host">The <see cref="IWebHost"/> to run.</param>
         public static void Run(this IWebHost host)
         {
+            AutoResetEvent done = new AutoResetEvent(false);
```
muratg: var :D

src/Microsoft.AspNetCore.TestHost/project.json
```diff
@@ -30,7 +30,7 @@
         "System.Net.Http": ""
       }
     },
-    "netstandard1.3": {
+    "netstandard1.5": {
```
davidfowl: This is a pretty big deal. Can't this also target both?

pakrym: Yes, it would.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -16,18 +19,26 @@ public static class WebHostExtensions
...
+                };
+#endif
                 Console.CancelKeyPress += (sender, eventArgs) =>
                 {
-                    Console.WriteLine("Application is shutting down...");
-                    cts.Cancel();
+                    Shutdown(cts);
 
-                    // Don't terminate the process immediately, wait for the Main thread to exit gracefully.
-                    eventArgs.Cancel = true;
```
davidfowl: How big of a change is this from the default behavior? Any observable differences?

pakrym: If there is something in `Program.Main` after `host.Run` it would not get run. But at least it would be consistent between `SIGTERM` and `ctrl+c`

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -5,6 +5,9 @@
 using System.Threading;
 using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.Extensions.DependencyInjection;
+#if NETSTANDARD1_5
+using System.Runtime.Loader;
```
Tratcher: sort

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -16,18 +19,27 @@ public static class WebHostExtensions
...
-                    cts.Cancel();
+                    if (!cts.IsCancellationRequested)
+                    {
+                        Console.WriteLine("Application is shutting down...");
+                        cts.Cancel();
+                    }
 
-                    // Don't terminate the process immediately, wait for the Main thread to exit gracefully.
-                    eventArgs.Cancel = true;
+                    done.WaitOne();
```
Tratcher: Not running client code after Run is a major regression.

src/Microsoft.AspNetCore.TestHost/project.json
```diff
@@ -34,6 +34,11 @@
       "dependencies": {
         "System.Diagnostics.Contracts": "4.3.0-*"
       }
+    },
+    "netstandard1.5": {
```
Tratcher: why did TestHost need to change? Shouldn't it have been fine building against the 1.3 version of Hosting?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
pakrym: There is no way we can support this with SIGTERM. So we need to decide if we want to have two behaviors or one common.

Tratcher: Sounds like we need better CLR events for SIGTERM.

pakrym:
Fixed comments, brought old behavior for `ctrl+c` back

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -16,18 +19,32 @@ public static class WebHostExtensions
...
+                    {
+                        Console.WriteLine("Application is shutting down...");
+                        cts.Cancel();
+                    }
 
+                    done.WaitOne();
+                };
+
+#if NETSTANDARD1_5
+                AssemblyLoadContext.Default.Unloading += context => shutdown();
```
davidfowl: So 80% of the case will be using this in program.Main(), for the 20% case though you might want to use the load context associated with this assembly instance.

```C#
var alc = AssemblyLoadContext.GetLoadContext(typeof(WebHostExtensions).GetTypeInfo().Assembly);
alc.Unloading  += context => shutdown();
```

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -16,18 +19,32 @@ public static class WebHostExtensions
         /// <param name="host">The <see cref="IWebHost"/> to run.</param>
         public static void Run(this IWebHost host)
         {
+            var done = new ManualResetEvent(false);
```
davidfowl: Slim?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -16,18 +20,33 @@ public static class WebHostExtensions
...
+                    done.Wait();
+                };
+
+#if NETSTANDARD1_5
+                var assemblyLoadContext = AssemblyLoadContext.GetLoadContext(typeof(WebHostExtensions).GetTypeInfo().Assembly);
+                assemblyLoadContext.Unloading += context => shutdown();
+#endif
+                Console.CancelKeyPress += (sender, eventArgs) =>
+                {
+                    shutdown();
```
stepro: The behavior of shutdown() is now synchronous vs. asynchronous. So long as CancelKeyPress comes in on a random background thread, this shouldn't matter. Just pointing out there may be a behavioral difference from the existing code.

PR #879, opened 2016-11-09
Test support for bin deployed aspnetcore.dll

Tratcher:
https://github.com/aspnet/IISIntegration/issues/292 @JunTaoLuo @pranavkm 
// CC: @shirhatti @pan-wang

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
@@ -77,7 +77,8 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                         ancmPath = @"%ProgramFiles%\IIS Express\aspnetcore.dll";
                     }
                     // Bin deployed by Microsoft.AspNetCore.AspNetCoreModule.nupkg
-                    else if (DeploymentParameters.ApplicationType == ApplicationType.Portable)
+                    else if (DeploymentParameters.RuntimeFlavor == RuntimeFlavor.CoreClr
+                        && DeploymentParameters.ApplicationType == ApplicationType.Portable)
                     {
                         ancmPath = Path.Combine(contentRoot, @"runtimes\win7-x64\native\aspnetcore.dll");
                     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -57,8 +68,13 @@ private CancellationToken StartIISExpress(Uri uri, string contentRoot)
                 // Pass on the applicationhost.config to iis express. With this don't need to pass in the /path /port switches as they are in the applicationHost.config
                 // We take a copy of the original specified applicationHost.Config to prevent modifying the one in the repo.
 
+                // The nupkg build of ANCM does not support Win7. https://github.com/aspnet/AspNetCoreModule/issues/40.
+                // If Win7 were supported we could just hardcode "aspnetcore.dll" in the config files.
+                var ancmPath = IsWin8orLater ? "aspnetcore.dll" : @"%ProgramFiles%\IIS Express\aspnetcore.dll";
```
JunTaoLuo: So we are skipping Win2008 R2 as well?

Tratcher: Correct, it has the same issue and version.

JunTaoLuo: Also, the old path looks like it's `%SystemRoot%\System32\inetsrv\aspnetcore.dll` why are we changing that now?

Tratcher: Yes. When we first wrote these tests there was only an installer for IIS so we used that dll. Now there's also an installer for express.

JunTaoLuo:
Changes look good. :shipit: Though @shirhatti had a question about 32-bit.

Tratcher:
Updated to use an absolute path for the bin deployed aspnetcore.dll. I realized the relative reference was ambiguous and preferred the dll adjacent to IISExpress.exe.

src/Microsoft.AspNetCore.Server.IntegrationTesting/project.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "Microsoft.AspNetCore.Testing": "1.1.0-*",
+    "Microsoft.DotNet.PlatformAbstractions": "1.1.0-*",
```
pranavkm: Can we use https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.RuntimeEnvironment.Sources/RuntimeEnvironment.cs instead?

pranavkm:
:shipit:

Tratcher:
FYI: This is going to be rebased on rel/1.1.0 and then merged back to dev. It will also be cherry-picked down to LTS (rel/1.0.*)

Tratcher:
Replacing this PR with one targeting rel/1.1.0

PR #884, opened 2016-11-16
Added EventSource to Hosting

kichalla:
@rynowak @davidfowl @Tratcher @shirhatti 

image

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
@@ -43,6 +43,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 _diagnosticSource.Write("Microsoft.AspNetCore.Hosting.BeginRequest", new { httpContext = httpContext, timestamp = startTimestamp });
             }
 
+            HostingEventSource.Log.RequestStart(httpContext.Request.Method, httpContext.Request.Path);
+
             return new Context
             {
                 HttpContext = httpContext,
@@ -78,7 +80,11 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 {
                     _diagnosticSource.Write("Microsoft.AspNetCore.Hosting.UnhandledException", new { httpContext = httpContext, timestamp = currentTimestamp, exception = exception });
                 }
+
+                HostingEventSource.Log.UnhandledException();
             }
+
+            HostingEventSource.Log.RequestStop();
 
             context.Scope?.Dispose();
 
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics.Tracing;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    [EventSource(Name = "Microsoft-AspNetCore-Hosting")]
+    public sealed class HostingEventSource : EventSource
+    {
+        public static readonly HostingEventSource Log = new HostingEventSource();
+
+        private HostingEventSource() { }
+
+        // NOTE
+        // - The 'Start' and 'Stop' suffixes on the following event names have special meaning in EventSource. They
+        //   enable creating 'activities'.
+        //   For more information, take a look at the following blog post:
+        //   https://blogs.msdn.microsoft.com/vancem/2015/09/14/exploring-eventsource-activity-correlation-and-causation-features/
+        // - A stop event's event id must be next one after its start event.
+
+        [Event(1, Level = EventLevel.Informational)]
+        public void HostStart()
+        {
+            WriteEvent(1);
+        }
+
+        [Event(2, Level = EventLevel.Informational)]
+        public void HostStop()
+        {
+            WriteEvent(2);
+        }
+
+        [Event(3, Level = EventLevel.Informational)]
+        public void RequestStart(string method, string path)
+        {
+            WriteEvent(3, method, path);
+        }
+
+        [Event(4, Level = EventLevel.Informational)]
+        public void RequestStop()
+        {
+            WriteEvent(4);
+        }
+
+        [Event(5, Level = EventLevel.Error)]
+        public void UnhandledException()
+        {
+            WriteEvent(5);
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -97,15 +97,15 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
         public virtual void Start()
         {
+            HostingEventSource.Log.HostStart();
+            _logger = _applicationServices.GetRequiredService<ILogger<WebHost>>();
+            _logger.Starting();
+
             Initialize();
 
             _applicationLifetime = _applicationServices.GetRequiredService<IApplicationLifetime>() as ApplicationLifetime;
-            _logger = _applicationServices.GetRequiredService<ILogger<WebHost>>();
             var diagnosticSource = _applicationServices.GetRequiredService<DiagnosticSource>();
             var httpContextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
-
-            _logger.Starting();
-
             Server.Start(new HostingApplication(_application, _logger, diagnosticSource, httpContextFactory));
 
             _applicationLifetime?.NotifyStarted();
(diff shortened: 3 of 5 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -99,6 +99,8 @@ public void Initialize()
 
         public virtual void Start()
         {
+            HostingEventSource.Log.HostStart();
```
Tratcher: move down to line 111 next to _logger? or to the very end of the method.

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
```diff
@@ -0,0 +1,94 @@
...
+                contentType,
+                contentLength,
+                scheme,
+                host,
+                pathBase,
+                path,
+                queryString);
+        }
+
+        [Event(4, Level = EventLevel.Informational)]
```
rynowak: Should we have a separate event for unhandled exceptions? (We'd fire both that **and** RequestStop) in that case... Just thinking aloud

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
```diff
@@ -0,0 +1,94 @@
...
+                requestTraceIdentifier,
+                protocol,
+                method,
+                contentType,
+                contentLength,
+                scheme,
+                host,
+                pathBase,
+                path,
+                queryString);
```
rynowak: I'm a little torn on the tradeoffs of concating the whole URL vs recording each part as you are doing here

test/Microsoft.AspNetCore.Hosting.Tests/Internal/HostingEventSourceTests.cs
```diff
@@ -0,0 +1,282 @@
...
+
+            // Assert
+            Assert.NotNull(eventSourceType);
+            Assert.Equal("Microsoft-AspNetCore-Hosting", EventSource.GetName(eventSourceType));
+            Assert.Equal(Guid.Parse("9e620d2a-55d4-5ade-deb7-c26046d245a8"), EventSource.GetGuid(eventSourceType));
+            Assert.NotEmpty(EventSource.GenerateManifest(eventSourceType, "assemblyPathToIncludeInManifest"));
+        }
+
+        [Fact]
+        public void HostStart()
```
rynowak: Is there any value in having these tests? For instance they can't really be parallelized.

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
```diff
@@ -0,0 +1,94 @@
...
+                    context.Request.QueryString.ToString());
+            }
+        }
+
+        [NonEvent]
+        public void RequestStop(HttpContext context, Exception exception = null)
+        {
+            if (IsEnabled())
+            {
+                RequestStop(
```
cwe1ss: What do you think about adding all fields from RequestStart to RequestStop as well?

Let's say I want to log each request to an external system (file system, Azure Log analytics, ...) and I want to be able to do reports on request durations per URL. 

Right now, the logger has to subscribe to both events and send both to the external system. This results in more traffic and data overhead and the external system must be able to merge the two events to do the reporting.
Alternatively, the logger could do the merging of the RequestStart/RequestStop events before it sends them but this also results in some complexity overhead (the RequestStart-event must be cached, etc if I'm correct).

Having all fields on RequestStop would allow these loggers to subscribe to just one event which contains all necessary information.
Systems that want to subscribe to both events are free to ignore these fields from RequestStop.

kichalla: That's a good point! I like the idea but will wait for others also to weigh in on this.

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
```diff
@@ -0,0 +1,94 @@
...
+        public void RequestStart(HttpContext context)
+        {
+            if (IsEnabled())
+            {
+                RequestStart(
+                    context.TraceIdentifier,
+                    context.Request.Protocol,
+                    context.Request.Method,
+                    context.Request.ContentType ?? string.Empty,
+                    context.Request.ContentLength.HasValue ? context.Request.ContentLength.Value.ToString() : string.Empty,
```
Tratcher: use context.Request.Headers[HeaderNames.ContentLength] instead.

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
```diff
@@ -0,0 +1,94 @@
...
+                    context.Request.Scheme,
+                    context.Request.Host.ToString(),
```
Tratcher: Host.Value

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
```diff
@@ -0,0 +1,94 @@
...
+                    context.Request.PathBase,
+                    context.Request.Path,
+                    context.Request.QueryString.ToString());
```
Tratcher: .Value

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
cwe1ss: Another thing I'd like to see in RequestStop is the RemoteIp. I think this would then cover most of the essential fields required to do IIS-like request logging.

It's important that the IP is in RequestStop and not just in RequestStart because if the application runs behind a proxy and uses the `ForwardedHeadersMiddleware`, RequestStart will always use the proxy IP.

However, if this mechanism should be used for request logging, the question will arise how people can log custom fields which won't be possible with this. But people could always write/use different middleware so I think this should be fine?!

davidfowl:
@kichalla  why is the build failing?

davidfowl:
@kichalla What's the perf hit with this on?

src/Microsoft.AspNetCore.Hosting/Internal/HostingEventSource.cs
kichalla: I based it on this with the idea that these fields could be query-able (but seems like this is not a goal for this exercise) https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs#L182

kichalla:
> why is the build failing?

The tests in this PR really cannot be really parallelized and are failing due to that. I have removed all but one test in the latest update.

kichalla:
> What's the perf hit with this on?

I haven't yet measured this but will check.

kichalla:
With the latest commit, the events look like below:

image

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
kichalla: Discussed offline. Keeping this line here and instead moved `_logger.Starting()` up

kichalla:
🆙 📅

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -78,6 +82,9 @@ public void DisposeContext(Context context, Exception exception)
                 {
                     _diagnosticSource.Write("Microsoft.AspNetCore.Hosting.UnhandledException", new { httpContext = httpContext, timestamp = currentTimestamp, exception = exception });
                 }
+
+                HostingEventSource.Log.UnhandledException();
+                HostingEventSource.Log.RequestStop();
```
davidfowl: And nuke this as a result.

davidfowl:
Are the tests supposed to be failing still?

test/Microsoft.AspNetCore.Hosting.Tests/Internal/HostingEventSourceTests.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Diagnostics.Tracing;
+using System.Reflection;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    public class HostingEventSourceTests
+    {
+        [Fact]
+        public void MatchesNameAndGuid()
```
davidfowl: Is this the only test worth writing? Any way we can wire up an event listener to make sure the events with the right data gets fired?

test/Microsoft.AspNetCore.Hosting.Tests/Internal/HostingEventSourceTests.cs
davidfowl: How do we know that any of this works when people make changes, or do we not care because it's for us?

kichalla: You mean inadvertently breaking contracts(example: changing parameter names on events etc.)? Yeah this is an issue but as you noted should not impact others as its for now used by us.

kichalla:
ping

davidfowl:
Looks good. Why are the tests failing again?

kichalla:
> Why are the tests failing again?

It appears this was because the CI nodes detected merge conflicts with your recent commit to this repo. I fixed the merge conflicts and pushed again. We can wait and see how the CI goes.

PR #886, opened 2016-11-18
[Prototype] Event counters

kichalla:
@davidfowl @rynowak 

`EventCounters` is an event that is logged for the interval time (in seconds) that we specify. Here the interval has been set as 1 sec. There is a timer which keeps writing out an EventCounters event for every event counter that is created and so you see the mean, standard deviation, count to be NaN initially (when the client hasn't made requests yet). Regarding using this data, just as any other event source event, these  `EventCounters` events can be processed via the TraceEvent nuget package and a UI could be built on top of it (if you are trying to compare the experience of perfmon).

image

image

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
@@ -42,18 +42,20 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             var httpContext = _httpContextFactory.Create(contextFeatures);
 
+            var eventLogEnabled = HostingEventSource.Log.IsEnabled();
+
             // These enabled checks are virtual dispatch and used twice and so cache to locals
             var diagnoticsEnabled = _diagnosticSource.IsEnabled(DiagnosticsBeginRequestKey);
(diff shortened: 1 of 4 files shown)


rynowak:
Ugh, still don't like that these are JSON-ified in the payload instead of their own columns... Did we have a particular need in mind for adding these? Or just general diagnostics?

davidfowl:
@kichalla this needs an update after @benaadams PR.

kichalla:
@davidfowl sure, updated

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -146,7 +148,7 @@ public void DisposeContext(Context context, Exception exception)
             if (eventLogEnabled)
             {
                 // Non-inline
-                hostingEventLog.RequestStop();
+                hostingEventLog.RequestStop(startTimestamp, currentTimestamp);
```
benaadams: Both values could be zero

benaadams: Ah, you check in the call

kichalla:
Closing this as we are not taking it in

PR #888, opened 2016-11-23
Fix IIS Express deployer for Win7

Tratcher:
@pranavkm @JunTaoLuo 2 issues:
1) File.Exists doesn't support environment variables like %ProgramFiles%. We only do this on Win7 because the nightly build doesn't work there.
2) The OS version check uses RuntimeEnvironment, which p-invokes into API sets that aren't available on Win7 by default. See https://github.com/dotnet/corefx/issues/8091#issuecomment-225641743. Workaround: add a reference to the ApiSets nupkg and add a runtimes section to the test project (separate PR).


pranavkm:
:shipit:

Issue #895, opened 2016-12-03
Consider renaming the methods on IApplicationLifetimeEvents (and maybe the interface)


davidfowl:
Instead of On* which looks like an event, maybe they should be proper method names (verbs).

```C#
public interface IApplicationLifetimeEvents
{
   // Maps to ApplicationStarted, this is where you populate caches etc (should it be async even though startup isn't).
   void InitializeApplication();

    // Maps to ApplicationStopping. Gives the application a chance to run logic before requests are drained.
   void StopApplication(); 
}
```

I can't think of a reason to handle ApplicationStopped. At that point the service provider is already disposed so you can't *really* call into anything safely.

NickCraver:
+1 on stripping "On"

However, the `OnApplicationStopped` event was useful. We planned on using it for telemetry after moving There are many uses in hosting environments for this: telemetry, removal from the load balancer, shared cache notifications, anything with queues and master professing, etc. We even do things like stop logging errors while a shutdown occurs because they're so much racing happening as to be unpredictable.

Please don't remove this call...unless there's another equivalent in the pipe we can hook into here?

davidfowl:
I'm on putting it back if we can find scenarios where it's usable. The fact that the service provider has already disposed everything is my motivation for removing it. If you call into things that have been disposed, they'll go boom.

Tratcher:
Design review:
IHostedService
- ctor
- Start
- Stop
- Dispose (if IDisposable)

benaadams:
Hmm... `Start` and `Stop` sound like commands rather than things happening.

Present participle better: `Starting` and `Stopping`?

benaadams:
Hmm... `Start` and `Stop` sound like commands rather than things happening.

Present participle better: `Starting` and `Stopping` if early in pipeline?
Past tense if later `Started` and `Stopped`?

davidfowl:
Thats the point. It's it's more like [IRegisteredObject](https://msdn.microsoft.com/en-us/library/system.web.hosting.iregisteredobject%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396) in previous ASP.NET. It's a service that gets notified on Start up of the application (after listening) and after before dispose. It's explicitly *not* about application events, thats the fundamental change.

benaadams:
Aha, gotcha. So its the declaration of the `Start` and `Stop` methods in the interface. Makes sense then 😄

davidfowl:
Yep!

Issue #894, opened 2016-12-03
Consider removing the dependency between IApplicationLifetime and IApplicationLifetimeEvents


davidfowl:
The following class will cause a circular dependency error in the DI system:

```C#
public class LifetimeEvents : IApplicationLifetimeEvents
{
    public LifetimeEvents(IApplicationLifetime lifetime)
    {

    }

    public void OnApplicationStarted()
    {
        
    }

    public void OnApplicationStopping()
    {
        
    }

    public void OnApplicationStopped()
    {

    }
}
```

This is because internally, the default implementation of `IApplicationLifetime` takes an `IEnumerable<IApplicationLifetimeEvents>`. The reason you might want to inject `IApplicationLifetime` is because you might want to call StopApplication at some point. Either way, the error is impossible to figure out without understanding how it's implemented.

PR #898, opened 2016-12-06
Review feedback for IApplicationLifetimeEvents

davidfowl:
- Renamed the type to IHostedService and added Start and Stop.
- Split up the IHostedService execution and IApplicationLifetime to avoid
circular references
- Trigger IHostedService.Start after starting the server
- Trigger IHostedService.Stop before disposing the service provider

#895 #894 

/cc @Tratcher @pakrym


benaadams:
Hmm... `Start` and `Stop` sound like commands rather than things happening.

Present participle better: `Starting` and `Stopping`?

src/Microsoft.AspNetCore.Hosting/Internal/HostedServiceExecutor.cs
```diff
@@ -0,0 +1,70 @@
...
+                    }
+
+                    exceptions.Add(ex);
+                }
+            }
+
+            // Throw an aggregate exception if there were any exceptions
+            if (exceptions != null)
+            {
+                throw new AggregateException(exceptions);
```
davidfowl: Sure. I don't think it matters though.

src/Microsoft.AspNetCore.Hosting/Internal/HostedServiceExecutor.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+            catch (Exception ex)
+            {
+                _logger.ApplicationError(LoggerEventIds.HostedServiceStopException, "An error occurred stopping the application", ex);
+            }
+        }
+
+        private void Execute(Action<IHostedService> callback)
+        {
+            List<Exception> exceptions = null;
```
mikaelm12: Would lazy initialization work better here?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -340,31 +337,96 @@ public void WebHostStopCallbacksDontMultipleTimes()
...
+                   .UseServer(this)
+                   .ConfigureServices(services =>
+                   {
+                       services.AddSingleton<IHostedService, TestHostedService>();
+                   })
+                   .Build();
+            var lifetime = host.Services.GetRequiredService<IApplicationLifetime>();
             lifetime.StopApplication();
+
+
```
mikaelm12: nit: Remove extra new line

src/Microsoft.AspNetCore.Hosting/Internal/HostedServiceExecutor.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
```
mikaelm12: Copyright and license header here

src/Microsoft.AspNetCore.Hosting/Internal/HostedServiceExecutor.cs
davidfowl: Happens once per app, so it doesn't matter and doesn't need to be thread safe.

src/Microsoft.AspNetCore.Hosting/Internal/HostedServiceExecutor.cs
davidfowl: Good catch

PR #902, opened 2016-12-14
Pass in cancellation token

mikaelm12:
@BrennanConroy @Tratcher


Tratcher:
Why, what will this actually do?

davidfowl:
It doesn't hurt but it doesn't do much either. All it does is not run the task if the ct is already cancelled.

PR #906, opened 2016-12-30
Add RunAsync Method, to help hosting in a GUI Env without blocking an…

waynebaby:
…y thread.

I am working on a wpf project and trying to host a net46 version of my site in process. a thread blocking is to much for a GUI app. This change could help.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
@@ -9,7 +9,7 @@ using System.Runtime.Loader;
 using System.Threading;
 using Microsoft.AspNetCore.Hosting.Server.Features;
 using Microsoft.Extensions.DependencyInjection;
-
+using System.Threading.Tasks;
 namespace Microsoft.AspNetCore.Hosting
 {
     public static class WebHostExtensions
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -57,15 +57,26 @@ public static void Run(this IWebHost host)
         /// <param name="token">The token to trigger shutdown.</param>
         public static void Run(this IWebHost host, CancellationToken token)
         {
-            host.Run(token, shutdownMessage: null);
+            host.RunAsync(token, shutdownMessage: null).Wait();
```
andriysavin: Should probably use .GetAwaiter().GetResult() for synchronous-like exception propagation.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -86,13 +97,17 @@ private static void Run(this IWebHost host, CancellationToken token, string shut
                     Console.WriteLine(shutdownMessage);
                 }
 
-                token.Register(state =>
+                token.Register(() =>
+                {
+                    applicationLifetime.StopApplication();
```
andriysavin: In the original code the `applicationLifetime` was passed as a parameter. I think that was done to avoid extra code generation for capturing local variables into the lambda.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -57,15 +57,26 @@ public static void Run(this IWebHost host)
...
 
-        private static void Run(this IWebHost host, CancellationToken token, string shutdownMessage)
+
+        private static async Task RunAsync(this IWebHost host, CancellationToken token, string shutdownMessage)
         {
             using (host)
             {
                 host.Start();
-
+                var tcs = new TaskCompletionSource<object>();
```
andriysavin: I'd move the declaration closer to its actual usage.

andriysavin:
You have actually made waiting for the host to shutdown asynchronously. But are you sure the code which (synchronously) **starts** the host won't introduce noticeable delay to the GUI app?

PR #913, opened 2017-01-12
Add setting to extend dotnet-publish arguments in ApplicationDeployer

natemcmaster:
Required to port MusicStore to vs2017.


src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -77,6 +77,12 @@ public class DeploymentParameters
         public string Configuration { get; set; } = "Debug";
 
         /// <summary>
+        /// Space separated command line arguments to be passed to dotnet-publish
+        /// </summary>
+        /// <returns></returns>
```
BrennanConroy: Don't need returns

BrennanConroy:
:shipit:

PR #925, opened 2017-02-01
Upgrade to VS 2017

natemcmaster:
Remove project.json and upgrade to MSBuild.

Other notable changes:
Removed unused baseline.json files. There were baseline files for "Microsoft.AspNetCore.Server.Testing" package, even though there was no project by that name. cc @javiercn @Eilon


Tratcher:
Those baselines should be moved from Microsoft.AspNetCore.Server.Testing to Microsoft.AspNetCore.Server.IntegrationTesting

natemcmaster:
That project isn't an automated test _project_. It's a test _app_. `.notest` prevents korebuild from running `dotnet test` on the project.

Eilon:
Regarding baselines, if the project got renamed, there's no point renaming the baseline, because the whole thing is a breaking change. Is this a package we publish? Should we even be checking for breaking changes in it?

natemcmaster:
I'm not sure the baseline files were ever used. They were added in 911da31476615a5ce4bfe2c9a59cffeddb8c2422, but even then, there was no project named Microsoft.AspNetCore.Server.Testing. See https://github.com/aspnet/Hosting/tree/911da31476615a5ce4bfe2c9a59cffeddb8c2422/src/Microsoft.AspNetCore.Server.Testing

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Microsoft.AspNetCore.Hosting.Server.Abstractions.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <Description>ASP.NET Core hosting server abstractions for web applications.</Description>
+    <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
davidfowl: Should this go into common.props?

natemcmaster: I think not. It's nice to keep this a little more visible as a reminder that we are missing XML docs in some API

davidfowl: If we used the `internal` keyword more I would agree. The fact that we have this in every single project is why I think it should go into the common props files.

natemcmaster: I still prefer it in the project, right next to the GenerateDocumentationFile flag, and for the sake of consistency about what common.props is really for... but I don't care a whole lot either way.

samples/SampleStartups/SampleStartups.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net451;netcoreapp1.1</TargetFrameworks>
```
BrennanConroy: Need a `RuntimeIdentifier` here?

natemcmaster: No. This is a class library. RID is only required for exe projects, and as a workaround for https://github.com/dotnet/sdk/issues/396

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Microsoft.AspNetCore.Hosting.Server.Abstractions.csproj
natemcmaster: ...and I'd guess most devs won't care either.

samples/SampleStartups/SampleStartups.csproj
BrennanConroy: Looking closer its neither, not supposed to be run, just showing examples of how startups work

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Microsoft.AspNetCore.Hosting.Server.Abstractions.csproj
Eilon: I'm with @natemcmaster here, we should leave it per-project. We should ideally do what we did in EF and put explicit docs for `.Internal` types saying they're not supported. But, until then...

PR #931, opened 2017-02-08
Check hostingLog enabled

benaadams:
Is showing up in traces; `IsEnabled()` can be inlined `Write(..` cannot, even though it then exits early with an enabled test.

Resolves https://github.com/aspnet/Hosting/issues/932

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
@@ -45,7 +45,9 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
             var hostingLog = HostingEventSource.Log;
             if (hostingLog.IsEnabled())
+            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -43,7 +43,9 @@ public Context CreateContext(IFeatureCollection contextFeatures)
                 _diagnosticSource.Write("Microsoft.AspNetCore.Hosting.BeginRequest", new { httpContext = httpContext, timestamp = startTimestamp });
             }
 
-            HostingEventSource.Log.RequestStart(httpContext.Request.Method, httpContext.Request.Path);
+            var hostingLog = HostingEventSource.Log;
+            if (hostingLog.IsEnabled())
```
Tratcher: style: always use curly brackets.

benaadams: Updated

PR #937, opened 2017-02-12
Fast-path for disabled logging 

benaadams:
`logger.IsEnabled(LogLevel.Information)` is a double check via generic + interface and is already checked for the timestamp; so can be used to short-circuit the `_logger.RequestStarting` call which doesn't inline; from which the `IsEnabled` test can then be removed.

If the `startTimestamp == 0` then logging was disabled and `_logger.RequestFinished` can be skipped. The test was in the function; but the function was non-inlinable.

Added `logger.IsEnabled(LogLevel.Information)` test to `_logger.RequestFinished` as `startTimestamp` could be non-zero if diagnostics was enabled but logging wasn't.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http;
@@ -13,6 +14,12 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class HostingApplication : IHttpApplication<HostingApplication.Context>
     {
+        private const string DiagnosticsBeginRequestKey = "Microsoft.AspNetCore.Hosting.BeginRequest";
+        private const string DiagnosticsEndRequestKey = "Microsoft.AspNetCore.Hosting.EndRequest";
+        private const string DiagnosticsUnhandledExceptionKey = "Microsoft.AspNetCore.Hosting.UnhandledException";
+
+        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
         private readonly RequestDelegate _application;
         private readonly ILogger _logger;
         private readonly DiagnosticSource _diagnosticSource;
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -34,19 +34,26 @@ public Context CreateContext(IFeatureCollection contextFeatures)
         {
             var httpContext = _httpContextFactory.Create(contextFeatures);
             var diagnoticsEnabled = _diagnosticSource.IsEnabled("Microsoft.AspNetCore.Hosting.BeginRequest");
-            var startTimestamp = (diagnoticsEnabled || _logger.IsEnabled(LogLevel.Information)) ? Stopwatch.GetTimestamp() : 0;
+            var loggingEnabled = _logger.IsEnabled(LogLevel.Information);
+
+            var startTimestamp = (diagnoticsEnabled || loggingEnabled) ? Stopwatch.GetTimestamp() : 0;
 
             var scope = _logger.RequestScope(httpContext);
```
benaadams: @davidfowl I assume the scope can't be put in the `if`?

davidfowl: No.

src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -23,21 +23,19 @@ public static IDisposable RequestScope(this ILogger logger, HttpContext httpCont
 
         public static void RequestStarting(this ILogger logger, HttpContext httpContext)
```
davidfowl: One thing that bugs me about this change is the fact that `RequestStarting` and `RequestFinished` pretty much rely on things that happen in the `HostingApplication`. Feels like those 2 methods should be moved there for proximity.

benaadams:
Refactoring the logging variants would look something like the last commit?

benaadams:
I've I move logging after diagnostics in `DisposeContext` its a bit cleaner?

Tratcher:
https://github.com/aspnet/Hosting/pull/944

benaadams:
> #944

Wouldn't be incompatable...

davidfowl:
I don't think #944 is viable.

davidfowl:
@benaadams Before I merge this thing. It needs comments. This code is very very very hard to follow 😄

benaadams:
@davidfowl comments better?

davidfowl:
Yep!

PR #940, opened 2017-02-14
Added default MiddlewareFactory service as scoped

davidfowl:
- This is a follow up to changes made in HttpAbstractions to
support a new IMiddleware interface


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -214,6 +214,7 @@ private IServiceCollection BuildCommonServices()
 
             services.AddTransient<IApplicationBuilderFactory, ApplicationBuilderFactory>();
             services.AddTransient<IHttpContextFactory, HttpContextFactory>();
+            services.AddScoped<IMiddlewareFactory, MiddlewareFactory>();
```
Tratcher: scoped because the IServiceProvider it accepts is scoped?

davidfowl: Right. To avoid passing the `IServiceProvider` as part of the interface contract the default factory itself is scoped.

PR #941, opened 2017-02-14
Faster RequestServices

benaadams:
Its an oddly chunky middleware

Changes

* `httpContext.Features` to local to reduce virtual lookups from 3 to 1 in common path
* Merge two finally blocks (`using` and `finally`)
* Inlinable `RequestServicesFeature` .ctor 

Pre

Post

I tried using a `ThreadStatic` cache for `RequestServicesFeature` and while it reduced most of the allocations; it was slower than allocating 😭

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
     public class RequestServicesContainerMiddleware
     {
         private readonly RequestDelegate _next;
-        private IServiceScopeFactory _scopeFactory;
+        private readonly IServiceScopeFactory _scopeFactory;
 
         public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFactory scopeFactory)
         {
@@ -25,38 +25,48 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 throw new ArgumentNullException(nameof(scopeFactory));
             }
 
-            _scopeFactory = scopeFactory;
             _next = next;
+            _scopeFactory = scopeFactory;
         }
 
         public async Task Invoke(HttpContext httpContext)
         {
             if (httpContext == null)
             {
-                throw new ArgumentNullException(nameof(httpContext));
+                // Defer Exception creation and throw to non-returning throw helper
+                ThrowArgumentNullExceptionHttpContext();
             }
 
-            var existingFeature = httpContext.Features.Get<IServiceProvidersFeature>();
+            // local cache for virtual disptach result
+            var features = httpContext.Features;
+            var existingFeature = features.Get<IServiceProvidersFeature>();
 
-            // All done if request services is set
+            // All done if RequestServices is set
             if (existingFeature?.RequestServices != null)
             {
                 await _next.Invoke(httpContext);
                 return;
             }
 
-            using (var feature = new RequestServicesFeature(_scopeFactory))
+            // _scopeFactory is pre-validated so use inlinable .ctor and property
+            var replacementFeature = new RequestServicesFeature() { ServiceScopeFactory = _scopeFactory };
+
+            try
             {
-                try
-                {
-                    httpContext.Features.Set<IServiceProvidersFeature>(feature);
-                    await _next.Invoke(httpContext);
-                }
-                finally
-                {
-                    httpContext.Features.Set(existingFeature);
-                }
+                features.Set<IServiceProvidersFeature>(replacementFeature);
+                await _next.Invoke(httpContext);
             }
+            finally
+            {
+                replacementFeature.Dispose();
+                // Restore previous feature state
+                features.Set(existingFeature);
+            }
+        }
+
+        private static void ThrowArgumentNullExceptionHttpContext()
+        {
+            throw new ArgumentNullException("httpContext");
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
@@ -45,7 +45,6 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 return;
             }
 
-            // _scopeFactory is pre-validated so use inlinable .ctor and property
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesFeature.cs
```diff
@@ -24,6 +24,20 @@ public RequestServicesFeature(IServiceScopeFactory scopeFactory)
             _scopeFactory = scopeFactory;
         }
 
+        internal RequestServicesFeature()
```
davidfowl: This is an "internal" type. We're ok changing it.

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -25,38 +25,48 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFac
...
             }
 
-            _scopeFactory = scopeFactory;
             _next = next;
+            _scopeFactory = scopeFactory;
         }
 
         public async Task Invoke(HttpContext httpContext)
         {
             if (httpContext == null)
```
davidfowl: This can't really happen either. I wonder if we should just Debug.Assert here as well. We don't check for a null context in our middleware in general...

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -25,38 +25,48 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFac
...
+            var features = httpContext.Features;
+            var existingFeature = features.Get<IServiceProvidersFeature>();
 
-            // All done if request services is set
+            // All done if RequestServices is set
             if (existingFeature?.RequestServices != null)
             {
                 await _next.Invoke(httpContext);
                 return;
             }
```
rynowak: My immediate thought is to care this up into a non-async path and an async path, but my guess is that we'd always take the async path.

benaadams: Yeah always takes the async path; it does change them into tail call jmps but still ends up doing a little more

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesFeature.cs
benaadams: Changed

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
benaadams: Changed

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -25,37 +26,37 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFac
...
 
-            using (var feature = new RequestServicesFeature(_scopeFactory))
+            var replacementFeature = new RequestServicesFeature(_scopeFactory);
```
davidfowl: Any reason for removing the using?

benaadams: It becomes
```csharp
try
{
    try
    {
        // ...
    }
    finally
    {
        feature.Dispose();
    }
}
finally
{
    features.Set(existingFeature);
}
```
So you pay the cost for the double try/finally; but Dispose should never throw so its just merging them

benaadams: Though with the state machine; I think its 4 try/finallys rather than just 2 😱

davidfowl: Because of the async state machine?

benaadams: State machine adds at least another two `try` blocks; one for execution context restore and one for sticking exception in task (may also be a third - the asm/il is a little hard to follow).

But this is just dealing with the `using` as it creates a `try` block with the `Dispose` in the `finally` and then there is the `try`/`finally` for setting the feature back that's inside the using. So its just turning them into one `try`/`finally` block rather than two.

davidfowl: I was being dense. We *already* had an explicit try finally. Sure it makes sense to merge them

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -25,37 +26,37 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFac
...
-                try
-                {
-                    httpContext.Features.Set<IServiceProvidersFeature>(feature);
-                    await _next.Invoke(httpContext);
-                }
-                finally
-                {
-                    httpContext.Features.Set(existingFeature);
-                }
+                replacementFeature.Dispose();
```
davidfowl: The order here changed from the previous implementation. I'm not sure it's relevant though.

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
benaadams: 😄

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
benaadams: Reversed the order as know for certain that `Dispose` doesn't throw; but not sure what features is so it could throw on the `Set`? So it has the same disposal guarantees that the previous arrangement had, even though the execution is different.

davidfowl: I guess `IFeatureCollection` *is* an abstraction and in theory it could throw. If you were relying on the order of dispose vs set it would probably be broken in other subtle ways... 😄

PR #943, opened 2017-02-14
Enable configuring the default service provider

davidfowl:
- Added UseDefaultServiceProvider method
- Made DelegateStartup use the IServiceProviderFactory. One downside
here is that we can't use 3rd party DI containers with the Configure
delegate since it's hardcoded to the the specific Startup type but that's
not a regression.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
@@ -29,7 +29,10 @@ namespace Microsoft.AspNetCore.Hosting
             return hostBuilder.UseSetting(WebHostDefaults.ApplicationKey, startupAssemblyName)
                               .ConfigureServices(services =>
                               {
-                                  services.AddSingleton<IStartup>(new DelegateStartup(configureApp));
+                                  services.AddSingleton<IStartup>(sp =>
+                                  {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -72,5 +75,21 @@ public static IWebHostBuilder UseStartup(this IWebHostBuilder hostBuilder, Type
...
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="configure">A callback used to configure the <see cref="ServiceProviderOptions"/> for the default <see cref="IServiceProvider"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hostBuilder, Action<ServiceProviderOptions> configure)
+        {
+            return hostBuilder.ConfigureServices(services =>
+            {
+                var options = new ServiceProviderOptions();
+                configure(options);
+                services.AddSingleton<IServiceProviderFactory<IServiceCollection>>(new DefaultServiceProviderFactory(options));
```
pakrym: Use `services.Replace` so you don't end up with two `IServiceProviderFactory<IServiceCollection>` in collection.

pakrym:
LGTM

PR #945, opened 2017-02-15
Allow for an alternate application name to be used as part of deploym…

pranavkm:
…ent.

This lets us test applications that are compiled with a different assembly name


pranavkm:
Required to E2E test https://github.com/aspnet/MvcPrecompilation/pull/82

davidfowl:
👍

Issue #896, opened 2016-12-05
TestServer should call IHttpContextFactory.Create with valid featureCollection


cortex93:
Hi,

I'm trying to set the TraceIdentifier from request headers early in the request processing before HostingApplication begins the scope and logs it.
To do so, I’m reading HttpContext.Request.Headers in a custom HttpContextFactory.Create method. While it works with Kestrel, it doesn't with TestServer.

From what I understand, TestServer do not populate HttpContext from the IFeatureCollection but it calls IHttpApplication.CreateContext with a fake RequestFeature and modify the returned HttpContext before calling IHttpApplication.ProcessRequestAsync.

If that's not a bug, what is the correct usage and how could I achieve my requirement

cortex93:
In RequestState, setting the RequestFeature properties directly with the one from given request.
I do the same with response,and finally create the Context to set the abortedSource token on it.

This make it works but I really don't know the implication.

Tratcher:
Yes, we could do that. The only reason it hadn't been done that way was that it was more convenient to set fields via the HttpContext. See https://github.com/aspnet/Hosting/blob/b6da89f54cff11474f17486cdc55c2f21f2bbd6b/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs#L128-L138

cortex93:
In fact, using the feature is also that convenient. Just replacing
```
Context = application.CreateContext(contextFeatures);
var httpContext = Context.HttpContext;
var serverRequest = httpContext.Request
```
with 
`var serverRequest = contextFeatures.Get<IHttpRequestFeature>();`

does the trick. QueryString is of different type so .Value will be required for QueryString.FromUriComponent

At the end, 
```
httpContext.Response.Body = _responseStream;
httpContext.Response.StatusCode = 200;
httpContext.RequestAborted = _requestAbortedSource.Token;
```

may be replace with
```
var responseFeature = contextFeatures.Get<IHttpResponseFeature>()
responseFeature.Body = _responseStream;
responseFeature.StatusCode = 200;
Context = application.CreateContext(contextFeatures);
Context.HttpContext.RequestAborted = _requestAbortedSource.Token;

```

davidfowl:
@Tratcher let's make this change.

/cc @muratg

PR #952, opened 2017-02-28
Have TestHost populate features before calling IHttpContextFactory.Create

Tratcher:
#896 IHttpContextFactory.Create may want to act on the values.

### DIFF
--- a/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
+++ b/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
@@ -126,49 +126,51 @@ namespace Microsoft.AspNetCore.TestHost
                 }
 
                 var contextFeatures = new FeatureCollection();
-                contextFeatures.Set<IHttpRequestFeature>(new RequestFeature());
+                var requestFeature = new RequestFeature();
+                contextFeatures.Set<IHttpRequestFeature>(requestFeature);
                 _responseFeature = new ResponseFeature();
                 contextFeatures.Set<IHttpResponseFeature>(_responseFeature);
-                Context = application.CreateContext(contextFeatures);
-                var httpContext = Context.HttpContext;
+                var requestLifetimeFeature = new HttpRequestLifetimeFeature();
+                contextFeatures.Set<IHttpRequestLifetimeFeature>(requestLifetimeFeature);
 
-                var serverRequest = httpContext.Request;
-                serverRequest.Protocol = "HTTP/" + request.Version.ToString(2);
-                serverRequest.Scheme = request.RequestUri.Scheme;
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
```diff
@@ -126,49 +126,51 @@ internal RequestState(HttpRequestMessage request, PathString pathBase, IHttpAppl
...
-                Context = application.CreateContext(contextFeatures);
-                var httpContext = Context.HttpContext;
+                var requestLifetimeFeature = new HttpRequestLifetimeFeature();
+                contextFeatures.Set<IHttpRequestLifetimeFeature>(requestLifetimeFeature);
 
-                var serverRequest = httpContext.Request;
-                serverRequest.Protocol = "HTTP/" + request.Version.ToString(2);
-                serverRequest.Scheme = request.RequestUri.Scheme;
-                serverRequest.Method = request.Method.ToString();
+                requestFeature.Protocol = "HTTP/" + request.Version.ToString(2);
```
cesarbs: I know it was already there, but what does the 2 stand for here?

Tratcher: two segments (1.0) rather than the full 4 segments.

cesarbs: Ok. I think switching to a named parameter makes it clearer.

test/Microsoft.AspNetCore.TestHost.Tests/ClientHandlerTests.cs
```diff
@@ -50,6 +50,30 @@ public Task ExpectedKeysAreAvailable()
...
+                Assert.Equal("/A/Path", features.Get<IHttpRequestFeature>().PathBase);
+                Assert.Equal("/and/file.txt", features.Get<IHttpRequestFeature>().Path);
+                Assert.Equal("?and=query", features.Get<IHttpRequestFeature>().QueryString);
+                Assert.NotNull(features.Get<IHttpRequestFeature>().Body);
+                Assert.NotNull(features.Get<IHttpRequestFeature>().Headers);
+                Assert.NotNull(features.Get<IHttpResponseFeature>().Headers);
+                Assert.NotNull(features.Get<IHttpResponseFeature>().Body);
+                Assert.Equal(200, features.Get<IHttpResponseFeature>().StatusCode);
+                Assert.Null(features.Get<IHttpResponseFeature>().ReasonPhrase);
+                Assert.Equal("example.com", features.Get<IHttpRequestFeature>().Headers["host"]);
```
JunTaoLuo: Also check `IHttpRequestLifetimeFeature` is not null?

Issue #917, opened 2017-01-19
Stop adding "http://localhost:5000" to empty IServerAddressesFeatures


halter73:
Servers such as Kestrel and HttpSysServer should define their own default listening endpoints (which will likely still listen on port 5000) so that the servers don't have to assume that a single address of "http://localhost:5000" indicates that no listening endpoint has been configured.

[Here is the relevant line of code](https://github.com/aspnet/Hosting/blob/80ae7f056c08b740820ee42a7df9eae34541e49e/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L245)

halter73:
https://github.com/aspnet/KestrelHttpServer/pull/1312 is Kestrel's reaction to this proposed change. The Kestrel PR can be merged first, since adding "http://localhost:5000" to IServerAddressesFeature only results in a warning.

shirhatti:
@Tratcher Do we have an issue to react to this in HttpSysServer as well?

Tratcher:
https://github.com/aspnet/HttpSysServer/issues/297

PR #953, opened 2017-03-01
Remove addition of default address to IServerAddressesFeature when no…

JunTaoLuo:
…ne is specified

#917


davidfowl:
Before merging this I'd like to discuss it in person

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -67,15 +67,6 @@ public void UseStartupThrowsWithNull()
         }
 
         [Fact]
-        public void CanDefaultAddressesIfNotConfigured()
```
cesarbs: Replace with test that verifies there are no default addresses.

PR #957, opened 2017-03-07
Avoid having to return anything in StartupBase.ConfigureServices

khellang:
This is more a proposal than anything else, as it's a breaking change. Maybe it could go in as part of 2.0?

For the vast majority of the time, when deriving from `StartupBase` you don't need or want to provide a custom `IServiceProvider`. You usually just call (and return) `base.ConfigureServices(services)` or `services.BuildServiceProvider()`. This is just noise and should be handled by the framework IMO.

This proposal makes the `IStartup.ConfigureServices` implementation explicit and adds a method, `CreateServiceProvider` to `StartupBase` that the user can override if they want to customize the provider.

It also aligns `StartupBase` and `StartupBase<TBuilder>` by having the latter derive from the former and override the new `CreateServiceProvider` method.

// @davidfowl

### DIFF
--- a/samples/SampleStartups/StartupBlockingOnStart.cs
+++ b/samples/SampleStartups/StartupBlockingOnStart.cs
@@ -11,11 +11,10 @@ namespace SampleStartups
 {
     public class StartupBlockingOnStart : StartupBase
     {
-        public override IServiceProvider ConfigureServices(IServiceCollection services)
+        public override void ConfigureServices(IServiceCollection services)
         {
             // This method gets called by the runtime. Use this method to add services to the container.
             // For more information on how to configure your application, visit http://go.microsoft.com/fwlink/?LinkID=398940
-            return base.ConfigureServices(services);
         }
 
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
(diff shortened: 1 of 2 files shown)


davidfowl:
I like it.

sebastienros:
The only special thing we need to do it register the service collection in the container like this:
https://github.com/OrchardCMS/Orchard2/blob/3d77b2184e3bc4774aba0b42e5a5a2219db94cd4/src/OrchardCore/Microsoft.AspNetCore.Modules/Extensions/ModularServiceCollectionExtensions.cs#L20

So as long as we still have access to it we should be good. We actually don't even use the default `StartupBase` base class in the modules, we still use our own which has the same method names.

davidfowl:
> So as long as we still have access to it we should be good. We actually don't even use the default StartupBase base class in the modules, we still use our own which has the same method names.

That's not confusing at all 😄

sebastienros:
It's not confusing they explicitly implement our `IStartup` or inherit from `StartupBase`. In a standard web app it's more confusing as it's convention based.
https://github.com/OrchardCMS/Orchard2/blob/master/src/OrchardCore/Microsoft.AspNetCore.Modules.Abstractions/Startup.cs

But I want to change it to the way the standard Startup files are used (by convention, not our own implementation).

davidfowl:
@sebastienros It's confusing because there's another `StartupBase` in Hosting that doesn't match.

khellang:
@davidfowl Any word on this making it into 2.0?

khellang:
Also, if you decide to pull this, it would be cool if you reconsidered including this as part of the templates as well. I think It's a much better default to steer people away from magic convention based methods. Especially now that a lot of the drift of the 1.x templates has been removed (like `ILoggerFactory` and `IConfiguration` setup).

Issue #951, opened 2017-02-28
Add Platform lightup hook


glennc:
https://github.com/dotnet/core-setup/issues/1597 allows us to do platform lightup on Azure.

We need to implement a hook that allows a server to inject packages and implement a package for Azure App Service that adds the Azure App Service logger.

This issue is for the hook. The implementation would go in the AzureIntegration repo.

muratg:
@davidfowl when can I review the code? :)

davidfowl:
I haven't been working on this.

davidfowl:
Follow up items
- If the application throws and the hosting startup assembly throws the hosting startup errors aren't show in the error page.
- Review error messages and logged error experiences

davidfowl:
@DamianEdwards and I agreed the exception behavior is correct. I don't think there's anything left to do here but remove the `/REVIEW:` from the codez

Tratcher:
What was the rational?

DamianEdwards:
Startup errors capturing is on by default with IIS integration and thus Azure. If it's on, and a loaded module fails, it's logged and the app runs as normal.

davidfowl:
Am closing this bug (again). We can open new bugs as we find issues with the experience

PR #961, opened 2017-03-11
Add support for executing IHostingStartup in specified assemblies

davidfowl:
- Assemblies that are specified in the "hostingStartupAssemblies" configuration (; delimited)
  setting can specify assemblies that use an assembly level attribute (HostingStartupAttribute)
  to specify a type that implements IHostingStartup. This allows hosting environments to
  extend the IWebHostBuilder with platform specific behavior before the application runs.
- If Startup fails and CaptureStartupErrors is true, we'll show the error in the usual
  Startup page.
- Added tests

#951

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingStartup.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingStartup.cs
@@ -6,7 +6,7 @@ using System;
 namespace Microsoft.AspNetCore.Hosting
 {
     /// <summary>
-    /// Represents platform specific configuration that will be applied to a <see cref="IWebHostBuilder"/> when building an <see cref="IWebHost"/>
+    /// Represents platform specific configuration that will be applied to a <see cref="IWebHostBuilder"/> when building an <see cref="IWebHost"/>.
     /// </summary>
     public interface IHostingStartup
     {
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingStartup.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingStartup.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    /// <summary>
+    /// Represents platform specific configuration that will be applied to a <see cref="IWebHostBuilder"/> when building an <see cref="IWebHost"/>
+    /// </summary>
+    public interface IHostingStartup
+    {
+        /// <summary>
+        /// Configure the <see cref="IWebHostBuilder"/>.
+        /// </summary>
+        /// <remarks>
+        /// Configure is intended to be called before user code, allowing a user to overwrite any changes made.
+        /// </remarks>
+        /// <param name="builder"></param>
+        void Configure(IWebHostBuilder builder);
+    }
+}
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Builder;
@@ -33,6 +34,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         private readonly IServiceProvider _hostingServiceProvider;
         private readonly WebHostOptions _options;
         private readonly IConfiguration _config;
+        private readonly ExceptionDispatchInfo _startupError;
 
         private IServiceProvider _applicationServices;
         private RequestDelegate _application;
@@ -47,7 +49,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             IServiceCollection appServices,
             IServiceProvider hostingServiceProvider,
             WebHostOptions options,
-            IConfiguration config)
+            IConfiguration config,
+            ExceptionDispatchInfo startupError)
         {
             if (appServices == null)
             {
@@ -65,6 +68,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             }
 
             _config = config;
+            _startupError = startupError;
             _options = options;
             _applicationServiceCollection = appServices;
             _hostingServiceProvider = hostingServiceProvider;
@@ -143,6 +147,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             try
             {
+                _startupError?.Throw();
+
                 EnsureApplicationServices();
                 EnsureServer();
 
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
-using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Builder;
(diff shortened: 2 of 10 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
```diff
@@ -26,10 +27,13 @@ public WebHostOptions(IConfiguration configuration)
             Environment = configuration[WebHostDefaults.EnvironmentKey];
             WebRoot = configuration[WebHostDefaults.WebRootKey];
             ContentRootPath = configuration[WebHostDefaults.ContentRootKey];
+            PlatformAssemblies = configuration[WebHostDefaults.HostingStartupAssembliesKey]?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
```
Tratcher: This naming is inconcistent and vague

davidfowl: HostingStartupAssemblies?

davidfowl:
> How well will this feature work when there are dependency version conflicts? E.g. how will it manage binding redirects?

As poorly/well as it does today.

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
Tratcher: Sure

src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingStartup.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    /// <summary>
+    /// Represents platform specific configuration that will be applied to a <see cref="IWebHostBuilder"/> when building an <see cref="IWebHost"/>
```
Eilon: End sentence with a period.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -677,12 +830,29 @@ public void Dispose()
...
             {
-                return NullLogger.Instance;
+                return new NoopLogger();
             }
 
             public void AddProvider(ILoggerProvider provider)
             {
             }
+
+            private class NoopLogger : ILogger
```
Eilon: Can't use the null logger stuff we have for this?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -643,6 +750,52 @@ public void Start<TContext>(IHttpApplication<TContext> application)
...
+            {
+                var loggerProvider = new TestLoggerProvider();
+                builder.UseSetting("testhostingstartup", "1")
+                       .ConfigureServices(services => services.AddSingleton<ServiceA>())
+                       .ConfigureServices(services => services.AddSingleton<ITestSink>(loggerProvider.Sink))
+                       .ConfigureLogging(lf => lf.AddProvider(loggerProvider));
+            }
+        }
+
+        public class TestLoggerProvider : ILoggerProvider
```
Eilon: Anything in https://github.com/aspnet/Logging/tree/dev/src/Microsoft.Extensions.Logging.Testing you can use?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -15,11 +17,13 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Extensions.ObjectPool;
 using Microsoft.Extensions.PlatformAbstractions;
 using Xunit;
 
+[assembly: HostingStartup(typeof(WebHostBuilderTests.TestHostingStartup))]
```
Eilon: Add a test scenario where there is more than 1 attribute specified?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
davidfowl: It clashes!

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
davidfowl: Hard to test both 1 and 2, it's a single assembly and I'd rather not add another assembly just to test this (unless we figured that was important somehow).

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
davidfowl: Once @ajaybhargavb removes the duplication.

Tratcher:
I was talking to @glennc about this and he doesn't want failures in these injected modules to crash the app, only to log them as warnings. Your current version crashes the app if their are Load errors or Configure errors.

Tratcher:
Is the opt-out approach to set the `hostingStartupAssemblies` config value to empty?

DamianEdwards:
@Tratcher they should log errors, not warnings, if the module throws.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
ajaybhargavb: Not anymore. You can now use them all you want 😄

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -200,6 +203,25 @@ private IServiceCollection BuildCommonServices()
...
+                    var assembly = Assembly.Load(new AssemblyName(assemblyName));
+
+                    foreach (var attribute in assembly.GetCustomAttributes<HostingStartupAttribute>())
+                    {
+                        var hostingStartup = (IHostingStartup)Activator.CreateInstance(attribute.HostingStartupType);
+                        hostingStartup.Configure(this);
+                    }
+                }
+            }
+            catch (Exception ex) when (_options.CaptureStartupErrors)
```
Tratcher: This needs to be capture and log per assembly, not throw on first failure.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -118,6 +121,26 @@ public virtual void Start()
...
+                {
+                    _logger.HostingStartupAssemblyError(exception);
+                }
+            }
+            else
+            {
+                // If there were no errors then just log the fact that we did load hosting startup assemblies.
+                if (_logger.IsEnabled(LogLevel.Debug))
+                {
+                    foreach (var assembly in _options.HostingStartupAssemblies)
```
Tratcher: If one fails and one succeeds then the success does not get logged. How about _hostingStartupErrors becomes _hostingStartupResults, `IEnumerable<KeyValuePair<string, exception>>` for the assembly name and optional error.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -200,6 +203,39 @@ private IServiceCollection BuildCommonServices()
...
+                    exceptions.Add(new InvalidOperationException($"Startup assembly {assemblyName} failed to execute. See the inner exception for more details.", ex));
+                }
+            }
+
+            if (exceptions.Count > 0)
+            {
+                hostingStartupErrors = new AggregateException(exceptions);
+
+                // Throw directly if we're not capturing startup errors
+                if (!_options.CaptureStartupErrors)
```
Tratcher: Still shouldn't throw, ever. Only log.

davidfowl: I don't agree. We can discuss this with @glennc and friends.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
davidfowl: How about just logging this all the time. That's simpler.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -118,6 +121,26 @@ public virtual void Start()
             _hostedServiceExecutor.Start();
 
             _logger.Started();
+
+            // REVIEW: Is this the right place to log these errors?
```
Eilon: Is there a bug logged for this? I'm not a fan of REVIEW/TODO comments in checked-in code (unless they point to an open issue).

PR #964, opened 2017-03-15
Add Activity tracking to HostingApplication

pakrym:
Ignore version/feed changes.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Diagnostics;
-using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http;
@@ -14,70 +13,31 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class HostingApplication : IHttpApplication<HostingApplication.Context>
     {
-        private const string DiagnosticsBeginRequestKey = "Microsoft.AspNetCore.Hosting.BeginRequest";
-        private const string DiagnosticsEndRequestKey = "Microsoft.AspNetCore.Hosting.EndRequest";
-        private const string DiagnosticsUnhandledExceptionKey = "Microsoft.AspNetCore.Hosting.UnhandledException";
-
-        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
-
         private readonly RequestDelegate _application;
-        private readonly ILogger _logger;
-        private readonly DiagnosticSource _diagnosticSource;
         private readonly IHttpContextFactory _httpContextFactory;
+        private HostingApplicationDiagnostics _diagnostics;
 
         public HostingApplication(
             RequestDelegate application,
             ILogger logger,
-            DiagnosticSource diagnosticSource,
+            DiagnosticListener diagnosticSource,
             IHttpContextFactory httpContextFactory)
         {
             _application = application;
-            _logger = logger;
-            _diagnosticSource = diagnosticSource;
+            _diagnostics = new HostingApplicationDiagnostics(logger, diagnosticSource);
             _httpContextFactory = httpContextFactory;
         }
 
         // Set up the request
         public Context CreateContext(IFeatureCollection contextFeatures)
         {
+            var context = new Context();
             var httpContext = _httpContextFactory.Create(contextFeatures);
 
-            // These enabled checks are virtual dispatch and used twice and so cache to locals
-            var diagnoticsEnabled = _diagnosticSource.IsEnabled(DiagnosticsBeginRequestKey);
-            var loggingEnabled = _logger.IsEnabled(LogLevel.Information);
+            _diagnostics.BeginRequest(httpContext, ref context);
 
-            if (HostingEventSource.Log.IsEnabled())
-            {
-                // To keep the hot path short we defer logging in this function to non-inlines
-                RecordRequestStartEventLog(httpContext);
-            }
-
-            // Only make call GetTimestamp if its value will be used, i.e. of the listenters is enabled
-            var startTimestamp = (diagnoticsEnabled || loggingEnabled) ? Stopwatch.GetTimestamp() : 0;
-
-            // Scope may be relevant for a different level of logging, so we always create it
-            // see: https://github.com/aspnet/Hosting/pull/944
-            var scope = _logger.RequestScope(httpContext);
-
-            if (loggingEnabled)
-            {
-                // Non-inline
-                LogRequestStarting(httpContext);
-            }
-
-            if (diagnoticsEnabled)
-            {
-                // Non-inline
-                RecordBeginRequestDiagnostics(httpContext, startTimestamp);
-            }
-
-            // Create and return the request Context
-            return new Context
-            {
-                HttpContext = httpContext,
-                Scope = scope,   // Scope can be null if logging is not on.   
-                StartTimestamp = startTimestamp,
-            };
+            context.HttpContext = httpContext;
+            return context;
         }
 
         // Execute the request
@@ -89,138 +49,10 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         // Clean up the request
         public void DisposeContext(Context context, Exception exception)
         {
-            // Local cache items resolved multiple items, in order of use so they are primed in cpu pipeline when used
-            var hostingEventLog = HostingEventSource.Log;
-            var startTimestamp = context.StartTimestamp;
             var httpContext = context.HttpContext;
-            var eventLogEnabled = hostingEventLog.IsEnabled();
-
-            // If startTimestamp is 0, don't call GetTimestamp, likely don't need the value
-            var currentTimestamp = (startTimestamp != 0) ? Stopwatch.GetTimestamp() : 0;
-
-            // To keep the hot path short we defer logging to non-inlines
-            if (exception == null)
-            {
-                // No exception was thrown, request was sucessful
-                if (_diagnosticSource.IsEnabled(DiagnosticsEndRequestKey))
-                {
-                    // Diagnostics is enabled for EndRequest, but it may not be for BeginRequest
-                    // so call GetTimestamp if currentTimestamp is zero (from above)
-                    RecordEndRequestDiagnostics(
-                        httpContext, 
-                        (currentTimestamp != 0) ? currentTimestamp : Stopwatch.GetTimestamp());
-                }
-            }
-            else
-            {
-                // Exception was thrown from request
-                if (_diagnosticSource.IsEnabled(DiagnosticsUnhandledExceptionKey))
-                {
-                    // Diagnostics is enabled for UnhandledException, but it may not be for BeginRequest 
-                    // so call GetTimestamp if currentTimestamp is zero (from above)
-                    RecordUnhandledExceptionDiagnostics(
-                        httpContext, 
-                        (currentTimestamp != 0) ? currentTimestamp : Stopwatch.GetTimestamp(), 
-                        exception);
-                }
-
-                if (eventLogEnabled)
-                {
-                    // Non-inline
-                    hostingEventLog.UnhandledException();
-                }
-            }
-
-            // If startTimestamp was 0, then Information logging wasn't enabled at for this request (and calcuated time will be wildly wrong)
-            // Is used as proxy to reduce calls to virtual: _logger.IsEnabled(LogLevel.Information)
-            if (startTimestamp != 0)
-            {
-                // Non-inline
-                LogRequestFinished(httpContext, startTimestamp, currentTimestamp);
-            }
-
-            // Logging Scope and context are finshed with
-            context.Scope?.Dispose();
+            _diagnostics.RequestEnd(httpContext, exception, context);
             _httpContextFactory.Dispose(httpContext);
-
-            if (eventLogEnabled)
-            {
-                // Non-inline
-                hostingEventLog.RequestStop();
-            }
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private void LogRequestStarting(HttpContext httpContext)
-        {
-            // IsEnabled is checked in the caller, so if we are here just log
-            _logger.Log(
-                logLevel: LogLevel.Information,
-                eventId: LoggerEventIds.RequestStarting,
-                state: new HostingRequestStartingLog(httpContext),
-                exception: null,
-                formatter: HostingRequestStartingLog.Callback);
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private void LogRequestFinished(HttpContext httpContext, long startTimestamp, long currentTimestamp)
-        {
-            // IsEnabled isn't checked in the caller, startTimestamp > 0 is used as a fast proxy check
-            // but that may be because diagnostics are enabled, which also uses startTimestamp, so check here
-            if (_logger.IsEnabled(LogLevel.Information))
-            {
-                var elapsed = new TimeSpan((long)(TimestampToTicks * (currentTimestamp - startTimestamp)));
-
-                _logger.Log(
-                    logLevel: LogLevel.Information,
-                    eventId: LoggerEventIds.RequestFinished,
-                    state: new HostingRequestFinishedLog(httpContext, elapsed),
-                    exception: null,
-                    formatter: HostingRequestFinishedLog.Callback);
-            }
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private void RecordBeginRequestDiagnostics(HttpContext httpContext, long startTimestamp)
-        {
-            _diagnosticSource.Write(
-                DiagnosticsBeginRequestKey,
-                new
-                {
-                    httpContext = httpContext,
-                    timestamp = startTimestamp
-                });
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private void RecordEndRequestDiagnostics(HttpContext httpContext, long currentTimestamp)
-        {
-            _diagnosticSource.Write(
-                DiagnosticsEndRequestKey,
-                new
-                {
-                    httpContext = httpContext,
-                    timestamp = currentTimestamp
-                });
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private void RecordUnhandledExceptionDiagnostics(HttpContext httpContext, long currentTimestamp, Exception exception)
-        {
-            _diagnosticSource.Write(
-                DiagnosticsUnhandledExceptionKey,
-                new
-                {
-                    httpContext = httpContext,
-                    timestamp = currentTimestamp,
-                    exception = exception
-                });
-        }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private static void RecordRequestStartEventLog(HttpContext httpContext)
-        {
-            HostingEventSource.Log.RequestStart(httpContext.Request.Method, httpContext.Request.Path);
+            _diagnostics.ContextDisposed(context);
         }
 
         public struct Context
@@ -228,6 +60,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             public HttpContext HttpContext { get; set; }
             public IDisposable Scope { get; set; }
             public long StartTimestamp { get; set; }
+            public bool EventLogEnabled { get; set; }
+            public Activity Activity { get; set; }
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
@@ -151,7 +151,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void ContextDisposed(HostingApplication.Context context)
         {
-            if (HostingEventSource.Log.IsEnabled())
+            if (context.EventLogEnabled)
             {
                 // Non-inline
                 HostingEventSource.Log.RequestStop();
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
@@ -20,7 +20,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
         private const string DeprecatedDiagnosticsBeginRequestKey = "Microsoft.AspNetCore.Hosting.BeginRequest";
         private const string DeprecatedDiagnosticsEndRequestKey = "Microsoft.AspNetCore.Hosting.EndRequest";
-        private const string DeprecatedDiagnosticsUnhandledExceptionKey = "Microsoft.AspNetCore.Hosting.UnhandledException";
+        private const string DiagnosticsUnhandledExceptionKey = "Microsoft.AspNetCore.Hosting.UnhandledException";
 
         private const string RequestIdHeaderName = "Request-Id";
         private const string CorrelationContextHeaderName = "Correlation-Context";
@@ -37,7 +37,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void BeginRequest(HttpContext httpContext, ref HostingApplication.Context context)
         {
-            long startTimestamp;
+            long startTimestamp = 0;
 
             if (HostingEventSource.Log.IsEnabled())
             {
@@ -60,20 +60,24 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 }
             }
 
-            // Scope may be relevant for a different level of logging, so we always create it
-            // see: https://github.com/aspnet/Hosting/pull/944
-            // Scope can be null if logging is not on.
-            context.Scope = HostingLoggerExtensions.RequestScope(_logger, httpContext);
-
-            if (_logger.IsEnabled(LogLevel.Information))
+            // To avoid allocation, return a null scope if the logger is not on at least to some degree.
+            if (_logger.IsEnabled(LogLevel.Critical))
             {
-                if (startTimestamp == 0)
-                {
-                    startTimestamp = Stopwatch.GetTimestamp();
-                }
+                // Scope may be relevant for a different level of logging, so we always create it
+                // see: https://github.com/aspnet/Hosting/pull/944
+                // Scope can be null if logging is not on.
+                context.Scope = _logger.RequestScope(httpContext);
 
-                // Non-inline
-                LogRequestStarting(httpContext);
+                if (_logger.IsEnabled(LogLevel.Information))
+                {
+                    if (startTimestamp == 0)
+                    {
+                        startTimestamp = Stopwatch.GetTimestamp();
+                    }
+
+                    // Non-inline
+                    LogRequestStarting(httpContext);
+                }
             }
 
             context.StartTimestamp = startTimestamp;
@@ -84,11 +88,23 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             // Local cache items resolved multiple items, in order of use so they are primed in cpu pipeline when used
             var startTimestamp = context.StartTimestamp;
-            long currentTimestamp;
+            long currentTimestamp = 0;
+
+            // If startTimestamp was 0, then Information logging wasn't enabled at for this request (and calcuated time will be wildly wrong)
+            // Is used as proxy to reduce calls to virtual: _logger.IsEnabled(LogLevel.Information)
+            if (startTimestamp != 0)
+            {
+                currentTimestamp = Stopwatch.GetTimestamp();
+                // Non-inline
+                LogRequestFinished(httpContext, startTimestamp, currentTimestamp);
+            }
 
             if (_diagnosticListener.IsEnabled())
             {
-                currentTimestamp = Stopwatch.GetTimestamp();
+                if (currentTimestamp == 0)
+                {
+                    currentTimestamp = Stopwatch.GetTimestamp();
+                }
 
                 if (exception == null)
                 {
@@ -103,7 +119,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 else
                 {
                     // Exception was thrown from request
-                    if (_diagnosticListener.IsEnabled(DeprecatedDiagnosticsUnhandledExceptionKey))
+                    if (_diagnosticListener.IsEnabled(DiagnosticsUnhandledExceptionKey))
                     {
                         // Diagnostics is enabled for UnhandledException, but it may not be for BeginRequest
                         // so call GetTimestamp if currentTimestamp is zero (from above)
@@ -124,18 +140,6 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             {
                 // Non-inline
                 HostingEventSource.Log.UnhandledException();
-            }
-
-            // If startTimestamp was 0, then Information logging wasn't enabled at for this request (and calcuated time will be wildly wrong)
-            // Is used as proxy to reduce calls to virtual: _logger.IsEnabled(LogLevel.Information)
-            if (startTimestamp != 0)
-            {
-                if (currentTimestamp == 0)
-                {
-                    currentTimestamp = Stopwatch.GetTimestamp();
-                }
-                // Non-inline
-                LogRequestFinished(httpContext, startTimestamp, currentTimestamp);
             }
 
             // Logging Scope is finshed with
@@ -210,7 +214,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         private void RecordUnhandledExceptionDiagnostics(HttpContext httpContext, long currentTimestamp, Exception exception)
         {
             _diagnosticListener.Write(
-                DeprecatedDiagnosticsUnhandledExceptionKey,
+                DiagnosticsUnhandledExceptionKey,
                 new
                 {
                     httpContext = httpContext,
@@ -248,7 +252,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 }
             }
 
-            if (_diagnosticListener.IsEnabled(ActivityStartKey, activity, httpContext))
+            if (_diagnosticListener.IsEnabled(ActivityStartKey))
             {
                 _diagnosticListener.StartActivity(activity, new { HttpContext = httpContext });
             }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
@@ -1,0 +1,276 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal class HostingApplicationDiagnostics
+    {
+        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
+        private const string ActivityName = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+        private const string ActivityStartKey = "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start";
+
+        private const string DiagnosticsBeginRequestKey = "Microsoft.AspNetCore.Hosting.BeginRequest";
+        private const string DiagnosticsEndRequestKey = "Microsoft.AspNetCore.Hosting.EndRequest";
+        private const string DiagnosticsUnhandledExceptionKey = "Microsoft.AspNetCore.Hosting.UnhandledException";
+
+        private const string RequestIdHeaderName = "Request-Id";
+        private const string CorrelationContextHeaderName = "Correlation-Context";
+
+        private readonly DiagnosticListener _diagnosticListener;
+        private readonly ILogger _logger;
+
+        public HostingApplicationDiagnostics(ILogger logger, DiagnosticListener diagnosticListener)
+        {
+            _logger = logger;
+            _diagnosticListener = diagnosticListener;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void BeginRequest(HttpContext httpContext, ref HostingApplication.Context context)
+        {
+            // These enabled checks are virtual dispatch and used twice and so cache to locals
+            var diagnosticsEnabled = _diagnosticListener.IsEnabled();
+            var legacyDiagnosticsEnabled = diagnosticsEnabled && _diagnosticListener.IsEnabled(DiagnosticsBeginRequestKey);
+            var loggingEnabled = _logger.IsEnabled(LogLevel.Information);
+            var eventLogEnabled = HostingEventSource.Log.IsEnabled();
+
+            if (eventLogEnabled)
+            {
+                context.EventLogEnabled = true;
+                // To keep the hot path short we defer logging in this function to non-inlines
+                RecordRequestStartEventLog(httpContext);
+            }
+
+            // Only make call GetTimestamp if its value will be used, i.e. of the listenters is enabled
+            var startTimestamp = (legacyDiagnosticsEnabled || loggingEnabled) ? Stopwatch.GetTimestamp() : 0;
+
+            // Scope may be relevant for a different level of logging, so we always create it
+            // see: https://github.com/aspnet/Hosting/pull/944
+            context.Scope = HostingLoggerExtensions.RequestScope(_logger, httpContext);
+
+            if (diagnosticsEnabled && _diagnosticListener.IsEnabled(ActivityName, httpContext))
+            {
+                context.Activity = StartActivity(httpContext);
+            }
+            if (loggingEnabled)
+            {
+                // Non-inline
+                LogRequestStarting(httpContext);
+            }
+            if (legacyDiagnosticsEnabled)
+            {
+                // Non-inline
+                RecordBeginRequestDiagnostics(httpContext, startTimestamp);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void RequestEnd(HttpContext httpContext, Exception exception, HostingApplication.Context context)
+        {
+            // Local cache items resolved multiple items, in order of use so they are primed in cpu pipeline when used
+            var startTimestamp = context.StartTimestamp;
+            var diagnosticsEnabled = _diagnosticListener.IsEnabled();
+
+            // If startTimestamp is 0, don't call GetTimestamp, likely don't need the value
+            var currentTimestamp = (startTimestamp != 0) ? Stopwatch.GetTimestamp() : 0;
+
+            // To keep the hot path short we defer logging to non-inlines
+            if (exception == null)
+            {
+                // No exception was thrown, request was sucessful
+                if (diagnosticsEnabled && _diagnosticListener.IsEnabled(DiagnosticsEndRequestKey))
+                {
+                    // Diagnostics is enabled for EndRequest, but it may not be for BeginRequest
+                    // so call GetTimestamp if currentTimestamp is zero (from above)
+                    RecordEndRequestDiagnostics(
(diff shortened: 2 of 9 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -64,21 +74,65 @@ public Context CreateContext(IFeatureCollection contextFeatures)
                 // Non-inline
                 LogRequestStarting(httpContext);
             }
-
             if (diagnoticsEnabled)
```
lmolkova: Besides sampling, we want to support other scenarios. 
Note that Activity carries duration, that makes Begin event almost redundant and increases log verbosity.
So there are 2 scenarios:
1. Users may listen to both: Begin and End Request
2. Users may want to receive End only.

Currently `diagnoticsEnabled` depends on Begin and changing it might break existing customers that listen to Begin only. I'm not sure if it makes any sense, since End event carries response.

I suggest 2 options:
1. Change diagnoticsEnabled to `_diagnosticSource.IsEnabled(DiagnosticsEndRequestKey)` and most probably no one will be hurt.

2. Deprecate Begin/End events and add a set of Activity events: "Microsoft.AspNetCore.Activity" (for IsEnabled), and Microsoft.AspNetCore.Activity.Start/Stop. This will not break anyone, but would require some deprecation plan and won't bring Activity to the old consumers

If we leave it as is, End-only scenario will be impossible. It also will be inconsistent with System.Net.Http instrumentation approach and ASP.NET Classic approach that is being implemented

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -64,21 +74,65 @@ public Context CreateContext(IFeatureCollection contextFeatures)
...
+                            }
+                            catch (ArgumentException ex)
+                            {
+                                _logger.LogWarning(0, ex, "Invalid baggage item '{ItemName}' with value '{ItemValue}'", baggageItem.Name, baggageItem.Value);
+                            }
+                        }
+                    }
+                }
+            }
+
```
lmolkova: There may need for advanced users/tracing systems to reduce performance impact from instrumentation by sampling some requests (e.g every 10% of requests; or requests with certain path(method, etc..); or requests which Activity.Current.Id falls into some bucket).
We planned to do so with  `DiagnosticSource.IsEnabled(string, object, object)` method.

Please call 
`DiagnosticSource.IsEnabled(event_name, activity, context)` before starting an activity.
event_name depends on what we'll decide in the other discussion I started for `diagnosticEnabled` condition

if it's false, then don't start activity and return null.

It will not affect exiting consumers since they provide callback for event name only. Then they will just receive IsEnabled(event_name) twice

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -98,6 +152,13 @@ public void DisposeContext(Context context, Exception exception)
             // If startTimestamp is 0, don't call GetTimestamp, likely don't need the value
             var currentTimestamp = (startTimestamp != 0) ? Stopwatch.GetTimestamp() : 0;
 
+#if NETSTANDARD2_0
+            // TODO: Ability to set duration directly or set end time from timestamp
+            context.Activity?
```
lmolkova: End event must be sent before Activity is stopped: consumers will receive End event and use Activity.Current props to log it.

For the same reason, Exception event must also be sent before activity is stopped

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -98,6 +152,13 @@ public void DisposeContext(Context context, Exception exception)
...
+                .SetEndTime(DateTime.UtcNow)
+                .Stop();
+#endif
+
             // To keep the hot path short we defer logging to non-inlines
             if (exception == null)
```
lmolkova: I'm not sure what happens now if request is cancelled, will exception be TaskCancelledException?
End (or Stop, depending on what we decide in other discussion) event must be sent for all requests including faulted or cancelled.

Imagine in your service you've got only Begin event and never got End. Does it mean there was an exception or user cancelled the request? Or log record was lost somehow because of bug or infrastructure problem. Or request is still being processed? Always sending End/Stop event helps consumer to do logging without guessing what happened.

End/Stop event payload must also carry TaskStatus that helps consumer to understand what happened with the request. It does not break any existing consumers.

I see also that `IsEnabled(RecordEndRequestDiagnostics)` is checked before sending the event. As I mentioned before, it's hard to imagine that someone listens to Begin, but not listens to End, so I think this check may be removed.

test/Microsoft.AspNetCore.Hosting.Tests/HostingApplicationTests.cs
```diff
@@ -20,16 +21,152 @@ public class HostingApplicationTests
...
+            var diagnosticSource = new DiagnosticListener("DummySource");
+            var hostingApplication = CreateApplication(out var features, diagnosticSource: diagnosticSource);
+
+            bool endCalled = false;
+            diagnosticSource.Subscribe(new CallbackDiagnosticListener(pair =>
+            {
+                if (pair.Key == "Microsoft.AspNetCore.Hosting.EndRequest")
+                {
+                    endCalled = true;
+                    Assert.Null(Activity.Current);
```
lmolkova: Activity.Current must not be null

test/Microsoft.AspNetCore.Hosting.Tests/HostingApplicationTests.cs
```diff
@@ -20,16 +21,152 @@ public class HostingApplicationTests
...
```
lmolkova: not null

lmolkova: check that End event is also sent and there is a TaskStatus in the payload of End

test/Microsoft.AspNetCore.Hosting.Tests/HostingApplicationTests.cs
```diff
@@ -20,16 +21,152 @@ public class HostingApplicationTests
...
+                }
+            }));
+
+            var context = hostingApplication.CreateContext(features);
+            hostingApplication.DisposeContext(context, null);
+            Assert.True(endCalled);
+        }
+
```
lmolkova: check that End is sent for cancelled requests is possible

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
lmolkova: @vancem please have a look

src/Microsoft.AspNetCore.Hosting/Internal/ActivityTrackingOptions.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    public class ActivityTrackingOptions
```
cwe1ss: Why not remove the options and stick to the default header names for now? There's no real interoperability story right now anyway since you require the values in a fixed format. It might be better to add options at a later point when it's clearer how the interoperability story will look like.

test/Microsoft.AspNetCore.Hosting.Tests/HostingApplicationTests.cs
davidfowl: There's is no cancelled request per se.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
pakrym: Why isn't listening only to a Begin event a valid scenario?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
pakrym: That if consumer want to log `Activity` with duration?

src/Microsoft.AspNetCore.Hosting/Internal/ActivityTrackingOptions.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
```
Tratcher: Needs its own off switch. Right now the only on/off control you get is linked to an unrelated diagnostic source.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -9,11 +9,17 @@
...
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class HostingApplication : IHttpApplication<HostingApplication.Context>
     {
+#if NETSTANDARD2_0
```
lmolkova: Yes, the latest DiagnosticSource nuget has an Activity from 1.3

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
lmolkova: It could be a valid scenario, but probably useless.
Why would you want to log only start of the request without end of it (i.e. result and duration)?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
lmolkova: Good catch! The order should be

SetEndTime
Write(RecordEndRequestDiagnostics)
Stop()

test/Microsoft.AspNetCore.Hosting.Tests/HostingApplicationTests.cs
lmolkova: @davidfowl What about `HttpContext.Abort()` and `HttpContext.RequestAborted`?

davidfowl: Those have nothing to do with the application function. Those merely trigger the cancellation token and will later cause calls to `Request.Body.ReadAsync` to fail. Well written servers will call `CreateContext`, `ProcessRequest` and `DisposeContext` in all cases regardless of the exception type.

src/Microsoft.AspNetCore.Hosting/Internal/ActivityTrackingOptions.cs
vancem: I would strongly suggest following Christian's suggestion.    It is HARD to get extensibility correct, and any PUBLIC surface has a compatibility requirement.  Thus it has the potential to be 'instant legacy' (that is no one uses it, but you have to support it forever).    We have learned the hard wary, that generally speaking if you don't have a reason to make something public, DON'T.    Thus unless you have people who ACTUALLY need this extensibility (and this is ALL they need to make their scenario work), you should not expose it at this time.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
pakrym: I just don't want to condition Activity creation on a single event subscription either begin or end.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
pakrym: Creating activity and reading headers is already performance impact does not running `Start()` actually save us from anything?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
lmolkova: In other places we have another set of events and checks:
* IsEnabled("*.Activity") used as a instrumentation switch 
* "*.Activity.Start" - test for start and event name
* "*.Activity.Stop" - stop event (no test)

However there is an intention to use Begin/End events to keep things the same for existing consumers. If this is the case, one of them has to be used

pakrym: We can check both of them in `CreateContext` and just store End value for later

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
lmolkova: if `DiagnosticSource.IsEnabled(event_name, activity, context)` is false, no events are sent. 
We assume on every Stop event consumer will write a log record and it would mean also going through all Activity properties: Id, StartTime, Duration, copying Tags, Baggage and sending log record to the storage. Heavily loaded services may want to avoid it for every request.
It also could be the case, that logging storage has some quota on log rate and customers have limited budget on it.

So there are many performance and business reasons to have sampling.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -9,11 +9,17 @@
...
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class HostingApplication : IHttpApplication<HostingApplication.Context>
     {
+#if !NET451
```
vancem: @lmolkova  I think we have now made DiagnosticSource work on V4.5 right?     (that is what the Activity.Current.net45.cs is about right?)   If so we should not need #If defs around this.    We really should figure this out and get rid of them.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -64,21 +75,68 @@ public Context CreateContext(IFeatureCollection contextFeatures)
...
+        {
+            var activity = new Activity(ActivityName);
+            StringValues requestId;
+            if (httpContext.Request.Headers.TryGetValue(_activityTrackingOptions.RequestIdHeaderName, out requestId))
+            {
+                try
+                {
+                    activity.SetParentId(requestId);
+                }
+                catch (ArgumentException ex)
```
vancem: If you are going to try to swallow exceptions you should catch them all, which means the  type the catch with 'Exception' rather than ArgumentException (This issue is in multiple places).    I would also prefer a try-catch around most of the routine (simpler).   I realize this gives slightly poorer error messages, but frankly the likelihood of these errors even happening, let alone the more detailed error message being useful is very low.

lmolkova: If we got invalid parentId or one invalid baggage item, it should not stop us from logging the request

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
lmolkova: I initially added 4.5 target, but it was removed in #15956 because of some build changes and also it looked like we won't have end-to-end story on .NET 4.5.
But now it seems we'll have it in AppInsight and we need Activity on NET4.5
Let me return it back.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
vancem: The try-catch should cause the Write() call but it can skip anything associated with setting up the activity.  It can even set Activity.Current.     I am just trying to simplify the code (less error handling more real logic).  

However thinking about it, I think I would prefer if Activity never throws (DiagnosticSource never throws today (assuming its target does not)) and that is good.   If I look at when Activity throws, it is mostly things that could be contracts (programmer errors, not runtime conditions).   We should probably have the ability to register a 'error callback' on Activity, and then otherwise never fail (simply do the best you can).     This allows us to get rid of try-catch 'goo' like this completely.   I will look into it.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -9,11 +9,17 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public class HostingApplication : IHttpApplication<HostingApplication.Context>
     {
+#if !NET451
+        private const string ActivityName = "Microsoft.AspNetCore.Hosting.Activity";
```
vancem: Don't call it 'Activity'.  It should be the name what the activity is.  Thus logically is should be 

    Microsoft.AspNetCore.Hosting.IncommingHttpRequest

But I would shorten it to 

    Microsoft.AspNetCore.Hosting.HttpReqIn

src/Microsoft.AspNetCore.Hosting/Internal/ActivityTrackingOptions.cs
vancem: The minimum concrete change is to simply make ActivityTrackingOptions be internal rather than public (but you could also rip out goo associated with passing it around.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -64,21 +75,68 @@ public Context CreateContext(IFeatureCollection contextFeatures)
...
+            }
+            if (_diagnosticSource.IsEnabled(DiagnosticsBeginRequestKey, activity, httpContext))
             {
-                HttpContext = httpContext,
-                Scope = scope,
-                StartTimestamp = startTimestamp,
-            };
+                return null;
+            }
+            activity.Start();
```
cwe1ss: Why doesn't this use the `StartActivity` extension method to also raise the corresponding event with access to properties like the HttpContext? Aren't the old `BeginRequest`/`EndRequest` events legacy now - similar to what you've done with HttpRequest?

Same for the `StopActivity` extension method.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
benaadams: > and as far as I know AppInsights will be enabled by default

Can someone provide assistance on [their use of async](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/src/Core/Managed/Shared/Channel/InMemoryTransmitter.cs#L100-L181) then?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -64,21 +75,68 @@ public Context CreateContext(IFeatureCollection contextFeatures)
...
+                }
+                catch (ArgumentException ex)
+                {
+                    _logger.LogWarning(LoggerEventIds.BadRequestIdHeader, ex, "Request ID received in '{RequestIdHeaderName}' header is invalid '{RequestId}'", _activityTrackingOptions.RequestIdHeaderName, requestId);
+                }
+
+                // We expect baggage to be empty by default
+                // Only very advanced users will be using it in near future, we encouradge them to keep baggage small (few items)
+                string[] baggage = httpContext.Request.Headers.GetCommaSeparatedValues(_activityTrackingOptions.BaggageHeaderName);
+                if (baggage != StringValues.Empty)
```
davidfowl: Is this comparison valid?

test/Microsoft.AspNetCore.Hosting.FunctionalTests/Microsoft.AspNetCore.Hosting.FunctionalTests.csproj
```diff
@@ -26,4 +26,8 @@
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
davidfowl: uh

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
vancem: Sigh, I see that we named the outgoing Http request System.Net.Http.Activity (I did not catch that in review), a much better name for that is System.Net.Http.HttpReqOut.   It may not be too late to fix that.  I will research this.

vancem: I have submitted a PR https://github.com/dotnet/corefx/pull/17296 to rename System.Net.Http.Activity to Ssytem.Net.Http.HttpReqOut.   I recommend we rename this to Microsoft.AspNetCore.Hosting.HttpReqIn.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -64,21 +75,68 @@ public Context CreateContext(IFeatureCollection contextFeatures)
...
-            return new Context
+            // Fill and return the request Context
+            context.HttpContext = httpContext;
+            context.Scope = scope;
+            context.StartTimestamp = startTimestamp;
+            return context;
+        }
+
+#if !NET451
+        private Activity StartActivity(HttpContext httpContext)
```
vancem: I would like this code to either follow the lead of the [outgoing HTTP code](https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/DiagnosticsHandler.cs#L33-L147) or we have good reasons for the differences (we can change the outgoing code if we feel we made a mistake).

On of the big things that the outgoing HTTP code did was NOT Try to reuse the existing events.  It kept them, but made it clear in the source code at least that they are deprecated.   This allows the events to use better names (making activities have 'Start' and 'Stop' suffixes) that will be consistent with all future activities.     The code is reasonably nicely segregated to make it easy to remove the old events after a time.   

It also starts conventions about filtering.    It assumes there are three levels of interesting filtering

1.  Off completely
2. Propagate (if something else turned on activity tracking, continue to do so) don't log events however 
3. On (or sampling) - low overhead   Turn on activity tracking even if it was off (that is initiate a new sample) only log Stop events (you want to log as little as possible, and Stop is the obvious choice since you know the most at that point)
4. On  with start events - everything in (3) plus start events.   Gives the receiver the most control/information.   

It achieves this by 

- if DiagnosticSource.IsEnabled() (no parameters) return true then we are in propagate.
- if DiagnosticSource.IsEnabled(ActivityName, request) returns true then we are ON (you get stops, you create an activity even if there was no parent) 
- if DiagnosticSource.IsEnabled(ActivityStartName) is true then we also log Start as well as Stop events. 

I would like us to 
1. Come to a consensus on what the right semantics are (I am suggesting the semantics above, both for compat as well as for how to turn on various levels.).
2. Insure that both of these implementations conform to what we decided.  

Comments?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -64,21 +75,68 @@ public Context CreateContext(IFeatureCollection contextFeatures)
...
+                            }
+                            catch (ArgumentException ex)
+                            {
+                                _logger.LogWarning(LoggerEventIds.BadRequestIdHeader, ex, "Baggage item received in '{BaggageHeaderName}' header is invalid '{ItemName}' with value '{ItemValue}'", _activityTrackingOptions.BaggageHeaderName, baggageItem.Name, baggageItem.Value);
+                            }
+                        }
+                    }
+                }
+            }
+            if (_diagnosticSource.IsEnabled(DiagnosticsBeginRequestKey, activity, httpContext))
```
lmolkova: typo: if **!**IsEnabled()

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplication.cs
```diff
@@ -89,145 +49,22 @@ public Task ProcessRequestAsync(Context context)
...
         }
 
         public struct Context
         {
             public HttpContext HttpContext { get; set; }
             public IDisposable Scope { get; set; }
             public long StartTimestamp { get; set; }
+            public bool EventLogEnabled { get; set; }
+
+#if !NET451
```
benaadams: Meaningless? Aren't building for `NET451` only for `netstandard1.3` and `netstandard1.5`

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal class HostingApplicationDiagnostics
+    {
+        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
+#if !NET451
```
benaadams: Meaningless? Aren't building for `NET451` only for `netstandard1.3` and `netstandard1.5`

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+            // Scope may be relevant for a different level of logging, so we always create it
+            // see: https://github.com/aspnet/Hosting/pull/944
+            context.Scope = HostingLoggerExtensions.RequestScope(_logger, httpContext);
+
+            if (loggingEnabled)
+            {
+                // Non-inline
+                LogRequestStarting(httpContext);
+            }
+#if !NET451
```
benaadams: Meaningless? Aren't building for `NET451` only for `netstandard1.3` and `netstandard1.5`

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+                }
+
+                if (context.EventLogEnabled)
+                {
+                    // Non-inline
+                    HostingEventSource.Log.UnhandledException();
+                }
+            }
+
+#if !NET451
```
benaadams: Meaningless? Aren't building for `NET451` only for `netstandard1.3` and `netstandard1.5`

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+                });
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static void RecordRequestStartEventLog(HttpContext httpContext)
+        {
+            HostingEventSource.Log.RequestStart(httpContext.Request.Method, httpContext.Request.Path);
+        }
+
+#if !NET451
```
benaadams: etc

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,272 @@
...
+                            catch (ArgumentException ex)
+                            {
+                                _logger.LogWarning(LoggerEventIds.InvalidRequestIdHeader, ex, "Baggage item received in '{BaggageHeaderName}' header is invalid '{ItemName}' with value '{ItemValue}'", CorrelationContextHeaderName, baggageItem.Name, baggageItem.Value);
+                            }
+                        }
+                    }
+                }
+            }
+
+            if (_diagnosticSource.IsEnabled(ActivityStartKey, activity, httpContext))
```
lmolkova: IsEnabled(ActivityStartKey)

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,272 @@
...
+            // Scope may be relevant for a different level of logging, so we always create it
+            // see: https://github.com/aspnet/Hosting/pull/944
+            context.Scope = HostingLoggerExtensions.RequestScope(_logger, httpContext);
+
+            if (loggingEnabled)
+            {
+                // Non-inline
+                LogRequestStarting(httpContext);
+            }
+            if (_diagnosticSource.IsEnabled(ActivityName, httpContext))
```
lmolkova: IsEnabled(ActivityName, activity, httpContext)

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            _logger = logger;
+            _diagnosticSource = diagnosticSource;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void BeginRequest(HttpContext httpContext, ref HostingApplication.Context context)
+        {
+            // These enabled checks are virtual dispatch and used twice and so cache to locals
+            var diagnoticsEnabled = _diagnosticSource.IsEnabled(DiagnosticsBeginRequestKey);
```
lmolkova: typo: diagno**S**ticsEnabled
It also seems that someone could subscribe to both: BeginRequest/EnsRequest and Activity events. 
It will be confusing to users and will not let to deprecate BeginRequest/EndRequest, so I suggest to check for IsEnabled(ActivityName) and only if it is false, try to send legacy events

benaadams:
Can you use the non-virtual `DiagnosticListener.IsEnabled()` test and push the `IsEnabled(string)` tests into the function if enabled?

Or is that not going to work?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,272 @@
...
+            {
+                _diagnosticSource.StopActivity(activity, new { HttpContext = httpContext });
+            }
+
+            // If startTimestamp was 0, then Information logging wasn't enabled at for this request (and calcuated time will be wildly wrong)
+            // Is used as proxy to reduce calls to virtual: _logger.IsEnabled(LogLevel.Information)
+            if (startTimestamp != 0)
+            {
+                // Non-inline
+                LogRequestFinished(httpContext, startTimestamp, currentTimestamp);
```
lmolkova: could you log it before activity is stopped?
It could be quite useful for custom ILogger implementation to use Activity.Current to augment logs

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,272 @@
...
+            var startTimestamp = (diagnoticsEnabled || loggingEnabled) ? Stopwatch.GetTimestamp() : 0;
+
+            // Scope may be relevant for a different level of logging, so we always create it
+            // see: https://github.com/aspnet/Hosting/pull/944
+            context.Scope = HostingLoggerExtensions.RequestScope(_logger, httpContext);
+
+            if (loggingEnabled)
+            {
+                // Non-inline
+                LogRequestStarting(httpContext);
```
lmolkova: could you log it after activity is started?
It could be quite useful for custom ILogger implementation to use Activity.Current to augment logs

lmolkova:
> Can you use the non-virtual DiagnosticListener.IsEnabled() test and push the IsEnabled(string) tests into the function if enabled?
>
> Or is that not going to work?
It works. Nice catch!
We have previously discussed that: `IsEnabled()` is  `DiagnosticListener` method, and `HostingApplication` receives `DiagnosticSource` object in ctor.
@pakrym is it possible to create HostingApplication with DiagnosticListener? It will allow eliminate no-instrumentation costs.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
pakrym: Quoting @vancem:

> •	if DiagnosticSource.IsEnabled(ActivityName, request) returns true then we are ON (you get stops, you create an activity even if there was no parent) Stop events.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
pakrym: Why don't we want to pass more information to `IsEnabled` if we have it anyway?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
pakrym: There might also be multiple consumers and one subscribing to new events would break other that would expect old to fire.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+        private const string ActivityStartKey = "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start";
+
+        private const string DiagnosticsBeginRequestKey = "Microsoft.AspNetCore.Hosting.BeginRequest";
+        private const string DiagnosticsEndRequestKey = "Microsoft.AspNetCore.Hosting.EndRequest";
+        private const string DiagnosticsUnhandledExceptionKey = "Microsoft.AspNetCore.Hosting.UnhandledException";
+
+        private const string RequestIdHeaderName = "Request-Id";
+        private const string CorrelationContextHeaderName = "Correlation-Context";
+
+
```
davidfowl: nit: remove extra space

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+            return activity;
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void StopActivity(HttpContext httpContext, Activity activity)
+        {
+            _diagnosticListener.StopActivity(activity, new { HttpContext = httpContext });
+        }
+
```
davidfowl: nit: remove extra space

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+                }
+
+                // We expect baggage to be empty by default
+                // Only very advanced users will be using it in near future, we encouradge them to keep baggage small (few items)
+                string[] baggage = httpContext.Request.Headers.GetCommaSeparatedValues(CorrelationContextHeaderName);
+                if (baggage != StringValues.Empty)
+                {
+                    foreach (var item in baggage)
+                    {
+                        NameValueHeaderValue baggageItem;
```
davidfowl: needs more C# 7, out var

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,280 @@
...
+        {
+            HostingEventSource.Log.RequestStart(httpContext.Request.Method, httpContext.Request.Path);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private Activity StartActivity(HttpContext httpContext)
+        {
+            var activity = new Activity(ActivityName);
+            StringValues requestId;
+            if (httpContext.Request.Headers.TryGetValue(RequestIdHeaderName, out requestId))
```
davidfowl: needs more C# 7

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,276 @@
...
+        {
+            _logger = logger;
+            _diagnosticListener = diagnosticListener;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void BeginRequest(HttpContext httpContext, ref HostingApplication.Context context)
+        {
+            // These enabled checks are virtual dispatch and used twice and so cache to locals
+            var diagnosticsEnabled = _diagnosticListener.IsEnabled();
```
vancem: This is a hot path (happens on every request whether logging is on or not), so we should try to minimize the amount of tests we make.   It is also nice if we use few local vars so that the prolog of the method is small.   Thus we would like at the top most level only three tests (one for diagnosticListener, one for ilogger logging and one for EventSource logging).    Everything else is under those three branches.   (thus when everything is off we do the minimum work).  Inside the diagnosticEnabled branch, we only test for Activity, and only test for the legacy if that FAILS.  That saves us a test in the ActivityCase (Which we expect to be common).

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,276 @@
...
+                RecordBeginRequestDiagnostics(httpContext, startTimestamp);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void RequestEnd(HttpContext httpContext, Exception exception, HostingApplication.Context context)
+        {
+            // Local cache items resolved multiple items, in order of use so they are primed in cpu pipeline when used
+            var startTimestamp = context.StartTimestamp;
+            var diagnosticsEnabled = _diagnosticListener.IsEnabled();
```
vancem: Same idea as for RequestStart.   We should have a minimum of work (tests) when things are off.  Thus 3 tests.   Will look a lot like the RequestStart case.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
pakrym: Wouldn't we break legacy events consumers as soon as someone subscribes to activity?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,276 @@
...
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal class HostingApplicationDiagnostics
+    {
+        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
+        private const string ActivityName = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
+        private const string ActivityStartKey = "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start";
+
+        private const string DiagnosticsBeginRequestKey = "Microsoft.AspNetCore.Hosting.BeginRequest";
```
vancem: Can we put 'Deprecated into the names for DiagnosticsBeginRequestKey  and DiagnosticsEndRequestKey.  This makes it clear (if you read the source) that these should not be used.   Strategic comments that tell you to use Activity instead are also welcome.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
vancem: Go ahead and test for both the legacy and non-legacy unconditionally.  The important case is when both are off, and in that case both need to be tested anyway.

pakrym: Also we are avoiding calling `Stopwatch.GetTimeStamp` twice to report same time to logger and diagnostic source.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,276 @@
...
+            HostingEventSource.Log.RequestStart(httpContext.Request.Method, httpContext.Request.Path);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private Activity StartActivity(HttpContext httpContext)
+        {
+            var activity = new Activity(ActivityName);
+            if (httpContext.Request.Headers.TryGetValue(RequestIdHeaderName, out var requestId))
+            {
+                try
```
vancem: I have created a pull request https://github.com/dotnet/corefx/pull/17626 that makes it so that Activity APIs catch any exceptions they throw (thus you see it in debuggers and monitors, but not code).  Thus you can remove the try-catch blocks here and down below since Activity is handling that for you.

test/Microsoft.AspNetCore.Hosting.Tests/HostingApplicationTests.cs
```diff
@@ -20,16 +22,243 @@ public class HostingApplicationTests
...
+                    Assert.NotNull(Activity.Current);
+                    Assert.True(Activity.Current.Duration > TimeSpan.Zero);
+                    Assert.Equal("Microsoft.AspNetCore.Hosting.HttpRequestIn", Activity.Current.OperationName);
+                    AssertProperty<HttpContext>(pair.Value, "HttpContext");
+                }
+            }));
+
+            var context = hostingApplication.CreateContext(features);
+            hostingApplication.DisposeContext(context, null);
+            Assert.True(endCalled);
```
lmolkova: could you please also check that exception event is sent before Stop?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,269 @@
...
+                {
+                    startTimestamp = Stopwatch.GetTimestamp();
+                    RecordBeginRequestDiagnostics(httpContext, startTimestamp);
+                }
+            }
+
+            // Scope may be relevant for a different level of logging, so we always create it
+            // see: https://github.com/aspnet/Hosting/pull/944
+            // Scope can be null if logging is not on.
+            context.Scope = HostingLoggerExtensions.RequestScope(_logger, httpContext);
```
vancem: I believe have reached a decision on https://github.com/aspnet/Hosting/pull/944  that Scopes should not be unconditional.   Thus you should put this inside a if ((_logger.IsEnabled(LogLevel.Critical) (and all other code for ILogger should go there as well (since critical -> informational).

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -0,0 +1,269 @@
...
+            // If startTimestamp was 0, then Information logging wasn't enabled at for this request (and calcuated time will be wildly wrong)
+            // Is used as proxy to reduce calls to virtual: _logger.IsEnabled(LogLevel.Information)
+            if (startTimestamp != 0)
+            {
+                if (currentTimestamp == 0)
+                {
+                    currentTimestamp = Stopwatch.GetTimestamp();
+                }
+                // Non-inline
+                LogRequestFinished(httpContext, startTimestamp, currentTimestamp);
```
lmolkova: could you call `LogRequestFinished` before stopping Activity. This way some custom ILogger implementation may just use Activity.Current.Id and other properties along with the log record

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
pakrym: We have that logic inside the `HostingLoggerExtensions.RequestScope` method, having it inside and outside seems strange.

vancem: The way it is now, we have two checks for the _logger case when it is turned off (one inside RequestScope and one here that guards the call to LogRequestStarting.   We could avoid that by doing what I suggest.   I agree that having the check both places is bad.  We should remove the check in RequestScope (my preference is frankly to remove RequestScope completely and put the call to BeginScope directly in this method).

test/Microsoft.AspNetCore.Hosting.Tests/HostingApplicationTests.cs
```diff
@@ -172,7 +170,29 @@ public void ActivityIsStoppedDuringUnhandledExceptionCall()
...
+        {
+            var diagnosticSource = new DiagnosticListener("DummySource");
+            var hostingApplication = CreateApplication(out var features, diagnosticSource: diagnosticSource);
+
+            bool endCalled = false;
+            diagnosticSource.Subscribe(new CallbackDiagnosticListener(pair =>
+            {
+                if (pair.Key == "Microsoft.AspNetCore.Hosting.UnhandledException")
+                {
+                    endCalled = true;
```
lmolkova: In case of exception we expect  Exception event followed by Stop event. 
It seems that test does not check Stop event, is it correct?

pakrym: If activity is not null then stop have not been called yet.

lmolkova: Right, but it does not mean Stop will be sent

pakrym: Test right above this checks that Stop is sent for exceptions

Issue #967, opened 2017-03-18
TestServer IServer.Features is null and makes it overly difficult to test middleware relying on ServerFeatures being populated.


mattmazzola:
**Use Case:**
I have written a stateless middleware which uses the address the server is configured to listen on and I want to write test for this middleware.

Example of extension method trying to extract listening addresses from server features:
```
public static IApplicationBuilder UseServiceFabricRouting(this IApplicationBuilder builder)
{
    var serverAddressesFeature = builder.ServerFeatures.Get<IServerAddressesFeature>();
    return builder.UseStatelessHttpMiddleware(new ServiceFabricRoutingMiddleware(serverAddressesFeature.Addresses.ToList()));
}
```

**Expected:**
I should be able to write tests using the normal aspnetcore convention of using `TestServer` such as below:

```
// Arrange
var webHostBuilder = new WebHostBuilder()
    .UseUrls("http://localhost:8000/appName/serviceName")
    .Configure(b => b
        .UseServiceFabricRouting()
        .Run(c =>
        {
            c.Response.StatusCode = (int)HttpStatusCode.OK;
            return Task.FromResult(0);
        }));

var testServer = new TestServer(webHostBuilder);

// Act
var response = await testServer
    .CreateRequest(path: "someOtherService")
    .SendAsync("GET");

// Assert
response.StatusCode.Should().Be(HttpStatusCode.Gone);
```

**Actual:**
I had to copy the implementation of `TestServer` into my own `ExtendedTestServer` class which is burdensome.  The whole point is for the tests to feel familiar and not be one off solution so I want my copy class to feel just like TestServer. This means I also have to duplicate the associated RequestBuilder class since it takes concrete dependency of TestServer.  Also the TestServer has method which invokes internal constructor of WebSocket so I can't even implement that method.

I think all of the above could be avoided with some very simple tweaks.

More background on server setup and why coping implementation was only solution (extending class or wrapping class does not work)

**Normal server setup:**
1: `UseKestrel()` will automatically setup the FeatureCollection
```
Features = new FeatureCollection();
_serverAddresses = new ServerAddressesFeature();
Features.Set<IServerAddressesFeature>(_serverAddresses);
```
https://github.com/aspnet/KestrelHttpServer/blob/632780dd16729298933f96796e6a969f3470d810/src/Microsoft.AspNetCore.Server.Kestrel/KestrelServer.cs#L54

[...]

PR #969, opened 2017-03-20
Add new constructor to TestSever which allows providing preconfigured FeatureCollection to use before Build / Start is invoked.

mattmazzola:
This allows setting up Features for tests to match behavior of Kestrel and WebListener.

I added a test which verifies it will address the issue I opened.

I hope this is non-breaking change since it's new constructor which people would have to opt into using. However, it does initialize the `IServer.Features` to new empty `FeatureCollection` so if someone was relying on that being null perhaps it is change.  From my understanding the WebHostBuilder was only populating ServerFeatures if the feature was set in the collection.

See related issue:
#967

### DIFF
--- a/src/Microsoft.AspNetCore.TestHost/TestServer.cs
+++ b/src/Microsoft.AspNetCore.TestHost/TestServer.cs
@@ -21,12 +21,21 @@ namespace Microsoft.AspNetCore.TestHost
         private IHttpApplication<Context> _application;
 
         public TestServer(IWebHostBuilder builder)
-            : this(builder, null)
+            : this(builder, new FeatureCollection())
         {
         }
 
         public TestServer(IWebHostBuilder builder, IFeatureCollection featureCollection)
         {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+            if (featureCollection == null)
+            {
+                throw new ArgumentNullException(nameof(featureCollection));
+            }        
+        
             Features = featureCollection;
 
             var host = builder.UseServer(this).Build();
--- a/src/Microsoft.AspNetCore.TestHost/TestServer.cs
+++ b/src/Microsoft.AspNetCore.TestHost/TestServer.cs
@@ -27,6 +27,14 @@ namespace Microsoft.AspNetCore.TestHost
             _hostInstance = host;
         }
 
+        public TestServer(IWebHostBuilder builder, Action<TestServer> configureServer)
+        {
+            configureServer(this);
+            var host = builder.UseServer(this).Build();
+            host.Start();
+            _hostInstance = host;
+        }
+
         public Uri BaseAddress { get; set; } = new Uri("http://localhost/");
 
         public IWebHost Host
@@ -37,7 +45,7 @@ namespace Microsoft.AspNetCore.TestHost
             }
         }
 
-        IFeatureCollection IServer.Features { get; }
+        IFeatureCollection IServer.Features { get; } = new FeatureCollection();
 
         public HttpMessageHandler CreateHandler()
         {
--- a/src/Microsoft.AspNetCore.TestHost/TestServer.cs
+++ b/src/Microsoft.AspNetCore.TestHost/TestServer.cs
@@ -21,10 +21,8 @@ namespace Microsoft.AspNetCore.TestHost
         private IHttpApplication<Context> _application;
 
         public TestServer(IWebHostBuilder builder)
+            : this(builder, null)
         {
-            var host = builder.UseServer(this).Build();
-            host.Start();
-            _hostInstance = host;
         }
 
         public TestServer(IWebHostBuilder builder, IFeatureCollection featureCollection)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.TestHost/TestServer.cs
```diff
@@ -27,6 +27,14 @@ public TestServer(IWebHostBuilder builder)
             _hostInstance = host;
         }
 
+        public TestServer(IWebHostBuilder builder, Action<TestServer> configureServer)
+        {
+            configureServer(this);
```
Tratcher: It's hazardous to expose a type instance externally from its own constructor. The type initialization hasn't completed yet so the state is indeterminate. e.g. Host is still null and code that operates on a TestServer expects Host to be non-null.

It may be cleaner to have a separate TestServerOptions class that can use used to flow the state you want.

mattmazzola: Hmm, I see. And I can't move `configureServer` down below initialization of host because by then `Build` has already been called.  I guess the configureServer solution won't work then.  I wanted to make it a function because I wanted to keep it flexible to accommodate more needs than just my own issue.

I'm imagining `TestSeverOptions` as a bunch of flags to which map to one off cases and I don't think that is scalable.

Was imagining something like:
```
public class TestServerOptions {
  public bool AddServerAddressesFeature { get; set; }
}
...
public TestServer(IWebHostBuilder builder, TestServerOptions testServerOptions) {
  if (testServerOptions.AddServerAddressesFeatures) {
    var server = this as IServer;
    server.Features.Set<IServerAddressesFeature>(new ServerAddressesFeature);
  }
}
```
Maybe you were imaging something different, do you have example?  If it is something like the above, this would likely lead TestServerOptions having many flags added to it for unrelated purposes.

I pushed new change which takes in `IFeatureCollection` and copies them into the internal collection.  I think it's clear about use case and doesn't expose `this` like old iteration.


```
public class TestServerOptions {
  public bool AddServerAddresses { get; set; }
}
...
public

File: src/Microsoft.AspNetCore.TestHost/TestServer.cs [hunk 2]
```diff
@@ -27,6 +27,19 @@ public TestServer(IWebHostBuilder builder)
             _hostInstance = host;
         }
 
+        public TestServer(IWebHostBuilder builder, IFeatureCollection featureCollection)
```
Tratcher: If you're going to pass in the collection then just directly assign it to IServer.Features. No need to copy it.

mattmazzola: I originally tried this:
`server.Features = featureCollection;`
but it says:
> Property or indexer `IServer.Features` cannot be assigned to - it is read only.

That's also why added the property initializer of:
`IFeatureCollection IServer.Features { get; } = new FeatureCollection();`
because I wasn't sure how else to give it a collection to work with.

Tratcher: It should be assignable in the constructor, especially if you don't initialize it otherwise. Try removing the `IServer.` prefix from the Features property to see if that's confusing it.

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
```diff
@@ -119,6 +120,30 @@ public void Configure(IApplicationBuilder app)
...
+        public void TestServerConstructorWithFeatureCollectionAllowsInitializingServerFeatures()
+        {
+            // Arrange
+            var url = "http://localhost:8000/appName/serviceName";
+            var builder = new WebHostBuilder()
+                .UseUrls(url)
+                .Configure(applicationBuilder =>
+                {
+                    var serverAddressesFeature = applicationBuilder.ServerFeatures.Get<IServerAddressesFeature>();
+                    Assert.Contains(serverAddressesFeature.Addresses, s => string.Compare(s, url, true) == 0);
```
JunTaoLuo: nit: would prefer `string.Equals(s, url, StringComparison.Ordinal)` instead

mattmazzola: Ah, yea that's nicer. Will change.

src/Microsoft.AspNetCore.TestHost/TestServer.cs
mattmazzola: Ok, I updated it with your suggested changes. That certainly cleans up the code a lot.
I originally thought Features was an explicit interface implementation because `Features` was common word and this was to avoid conflicts with future features from other types.

src/Microsoft.AspNetCore.TestHost/TestServer.cs
```diff
@@ -27,6 +27,15 @@ public TestServer(IWebHostBuilder builder)
             _hostInstance = host;
         }
```
Tratcher: This old constructor can call the new one now.

Issue #974, opened 2017-03-21
Shutdown Test fail on CI


jbagga:
[ShutdownTest](https://github.com/aspnet/Hosting/blob/dev/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs#L19) is failing on the Ubuntu Universe CI test run.

http://aspnetci/viewLog.html?buildId=190441&tab=buildResultsDiv&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR

@pakrym

JunTaoLuo:
I found the problem.

PR #975, opened 2017-03-21
Disable test

jbagga:
See #974 

cc @pakrym


test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
```diff
@@ -14,6 +14,8 @@ namespace Microsoft.AspNetCore.Hosting.FunctionalTests
     public class ShutdownTests
     {
         [ConditionalFact]
+        //Disabled until https://github.com/aspnet/Hosting/issues/974 resolved.
```
JunTaoLuo: Generally, it's better to disable using `[ConditionalFact(Skip = "Skip reason, url_to_issue")]` Easier than to remember to remove a OSSkipCondition when re-enabling the test, especially when there are more than one.

Issue #955, opened 2017-03-02
First class Configuration and Logging in WebHostBuilder


glennc:
We want to enable the following program main:

```csharp
public static void Main(string[] args)
        {
            var host = new WebHostBuilder()
                .UseContentRoot(Directory.GetCurrentDirectory())
                .UseConfiguration((configBuilder, env) => 
                {
                    configBuilder.AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
                                 .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true)
                                 .AddEnvironmentVariables()
                                 .AddCommandLine(args);
                })
                .UseLogging((factory, config) =>
                {
                    factory.AddFilter(config.GetSection("Logging"));
                    factory.AddConsole();
                    factory.AddDebug();
                })
                .UseIISIntegration()
                .UseStartup<Startup>()
                .UseKestrel()
                .Build();

            host.Run();
        }
```

This issue is to track the WebHostBuilder part of this.

Tratcher:
* With an eye to removing these parts from Startup and exposing config via DI.

cwe1ss:
Sounds great! How will users be able to get the configuration object?

cwe1ss:
> With an eye to removing these parts from Startup and exposing config via DI.

Would I be able to resolve `IConfiguration` or would it be a separate type? My concern with IConfiguration are scenarios where applications use multiple independent configuration objects. (e.g. one for appSettings and one for secrets from Azure Key Vault). It might not be clear what gets injected when people use IConfiguration. Not sure if this is a real concern though?!

glennc:
At the moment I think we will just put IConfigurationRoot in DI. We are hearing from many people that it's what they want.

Not completely obvious in this, but relevent to you @cwe1ss is that this will play into the other issue about deprecating AddProvider on LoggerFactory and allowing Serilog, NLog, etc to replace the LoggerFactory instead of having a provider.

glennc:
Having IConfiguration in DI doesn't stop you from using multiple, but you would need to define them yourselves and pick which one is the one that goes into DI. Ideally use Options as well instead of Configuration directly.

cwe1ss:
Yep, just wanted to point it out... it might not be an issue. 
Will you add IConfigurationRoot and IConfiguration? IConfiguration seems nicer to me for DI because it's the more generic interface. People who use 3rd party DI containers could probably do more stuff with it (e.g. Injecting just a section).

[...]

PR #978, opened 2017-03-22
Configuration on WebHostBuilder

glennc:
The intent here was to see what it will look like if we have the ability to configure configuration and logging in `main` as well as setup for when `ILoggerFactory` no longer has `AddProvider`. I didn't obsolete things here, but the intent would be to obsolete `ConfigureLogging(ILoggerFactory)` and possibly `UseLoggerFactory(ILoggerFactory)`.

See: https://github.com/aspnet/Hosting/issues/955 and https://github.com/aspnet/Logging/issues/420

The configuration part of it seems fine, but logging not so much.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
@@ -39,6 +40,13 @@ namespace Microsoft.AspNetCore.Hosting
         IWebHostBuilder ConfigureLogging(Action<ILoggerFactory> configureLogging);
 
         /// <summary>
+        /// Adds a delegate for configuring the provided <see cref="ILoggerFactory"/>. This may be called multiple times.
+        /// </summary>
+        /// <param name="configureLogging">The delegate that configures the <see cref="ILoggerFactory"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder ConfigureLogging<T>(Action<T> configureLogging) where T : ILoggerFactory;
+
+        /// <summary>
         /// Add or replace a setting in the configuration.
         /// </summary>
         /// <param name="key">The key of the setting to add or replace.</param>
@@ -52,5 +60,21 @@ namespace Microsoft.AspNetCore.Hosting
         /// <param name="key">The key of the setting to look up.</param>
         /// <returns>The value the setting currently contains.</returns>
         string GetSetting(string key);
+
+        /// <summary>
+        /// Adds a delegate to construct the <see cref="ILoggerFactory"/> that will be registered
+        /// as a singleton and used by the application.
+        /// </summary>
+        /// <param name="createLoggerFactory">The delegate that constructs an <see cref="IConfigurationBuilder" /></param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder UseLoggerFactory(Func<WebHostBuilderContext, ILoggerFactory> createLoggerFactory);
+
+
+        /// <summary>
+        /// Adds a delegate for configuring the <see cref="IConfigurationBuilder"/> that will construct an <see cref="IConfiguration"/>.
+        /// </summary>
+        /// <param name="configureDelegate">The delegate for configuring the <see cref="IConfigurationBuilder" /> that will be used to construct an <see cref="IConfiguration" />.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder ConfigureConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate);
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -29,9 +29,10 @@ namespace Microsoft.AspNetCore.Hosting
         private readonly List<Action<ILoggerFactory>> _configureLoggingDelegates;
 
         private IConfiguration _config;
-        private ILoggerFactory _loggerFactory;
         private WebHostOptions _options;
         private bool _webHostBuilt;
+        private Func<WebHostBuilderContext, ILoggerFactory> _createLoggerFactoryDelegate;
+        private List<Action<WebHostBuilderContext, IConfigurationBuilder>> _configureConfigurationBuilderDelegates;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WebHostBuilder"/> class.
@@ -41,6 +42,7 @@ namespace Microsoft.AspNetCore.Hosting
             _hostingEnvironment = new HostingEnvironment();
             _configureServicesDelegates = new List<Action<IServiceCollection>>();
             _configureLoggingDelegates = new List<Action<ILoggerFactory>>();
+            _configureConfigurationBuilderDelegates = new List<Action<WebHostBuilderContext, IConfigurationBuilder>>();
 
             _config = new ConfigurationBuilder()
                 .AddEnvironmentVariables(prefix: "ASPNETCORE_")
@@ -94,7 +96,7 @@ namespace Microsoft.AspNetCore.Hosting
                 throw new ArgumentNullException(nameof(loggerFactory));
             }
 
-            _loggerFactory = loggerFactory;
+            _createLoggerFactoryDelegate = _ => loggerFactory;
             return this;
         }
 
@@ -128,6 +130,60 @@ namespace Microsoft.AspNetCore.Hosting
             }
 
             _configureLoggingDelegates.Add(configureLogging);
+            return this;
+        }
+
+        /// <summary>
+        /// Adds a delegate to construct the <see cref="ILoggerFactory"/> that will be registered
+        /// as a singleton and used by the application.
+        /// </summary>
+        /// <param name="createLoggerFactory">The delegate that constructs an <see cref="IConfigurationBuilder" /></param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public IWebHostBuilder UseLoggerFactory(Func<WebHostBuilderContext, ILoggerFactory> createLoggerFactory)
+        {
+            if (createLoggerFactory == null)
+            {
+                throw new ArgumentNullException(nameof(createLoggerFactory));
+            }
+
+            _createLoggerFactoryDelegate = createLoggerFactory;
+            return this;
+        }
+
+        /// <summary>
+        /// Adds a delegate for configuring the provided <see cref="ILoggerFactory"/>. This may be called multiple times.
+        /// </summary>
+        /// <param name="configureLogging">The delegate that configures the <see cref="ILoggerFactory"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public IWebHostBuilder ConfigureLogging<T>(Action<T> configureLogging) where T : ILoggerFactory
+        {
+            if (configureLogging == null)
+            {
+                throw new ArgumentNullException(nameof(configureLogging));
+            }
+            _configureLoggingDelegates.Add(factory =>
+            {
+                if (factory is T typedFactory)
+                {
+                    configureLogging(typedFactory);
+                }
+            });
+            return this;
+        }
+
+        /// <summary>
+        /// Adds a delegate for configuring the <see cref="IConfigurationBuilder"/> that will construct an <see cref="IConfiguration"/>.
+        /// </summary>
+        /// <param name="configureDelegate">The delegate for configuring the <see cref="IConfigurationBuilder" /> that will be used to construct an <see cref="IConfiguration" />.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public IWebHostBuilder ConfigureConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate)
+        {
+            if (configureDelegate == null)
+            {
+                throw new ArgumentNullException(nameof(configureDelegate));
+            }
+
+            _configureConfigurationBuilderDelegates.Add(configureDelegate);
             return this;
         }
 
@@ -185,23 +241,35 @@ namespace Microsoft.AspNetCore.Hosting
             var appEnvironment = PlatformServices.Default.Application;
             var contentRootPath = ResolveContentRootPath(_options.ContentRootPath, appEnvironment.ApplicationBasePath);
             var applicationName = _options.ApplicationName ?? appEnvironment.ApplicationName;
+            var hostingContext = new WebHostBuilderContext
+            {
+                Configuration = _config
+            };
 
             // Initialize the hosting environment
             _hostingEnvironment.Initialize(applicationName, contentRootPath, _options);
+            hostingContext.HostingEnvironment = _hostingEnvironment;
 
             var services = new ServiceCollection();
             services.AddSingleton(_hostingEnvironment);
 
+            var builder = new ConfigurationBuilder()
+                .SetBasePath(_hostingEnvironment.ContentRootPath)
+                .AddInMemoryCollection(_config.AsEnumerable());
+
+            foreach (var configureConfiguration in _configureConfigurationBuilderDelegates)
+            {
+                configureConfiguration(hostingContext, builder);
+            }
+
+            var configuration = builder.Build();
+            services.AddSingleton<IConfiguration>(configuration);
+            hostingContext.Configuration = configuration;
+
             // The configured ILoggerFactory is added as a singleton here. AddLogging below will not add an additional one.
-            if (_loggerFactory == null)
-            {
-                _loggerFactory = new LoggerFactory();
-                services.AddSingleton(provider => _loggerFactory);
-            }
-            else
-            {
-                services.AddSingleton(_loggerFactory);
-            }
+            var loggerFactory = _createLoggerFactoryDelegate?.Invoke(hostingContext) ?? new LoggerFactory();
+            services.AddSingleton(loggerFactory);
+            hostingContext.LoggerFactory = loggerFactory;
 
             var exceptions = new List<Exception>();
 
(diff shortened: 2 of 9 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -29,9 +29,10 @@ public class WebHostBuilder : IWebHostBuilder
         private readonly List<Action<ILoggerFactory>> _configureLoggingDelegates;
 
         private IConfiguration _config;
-        private ILoggerFactory _loggerFactory;
         private WebHostOptions _options;
         private bool _webHostBuilt;
+        private Func<IHostingEnvironment, IConfiguration, ILoggerFactory> _loggerFactoryFunc;
+        private Action<IConfigurationBuilder, IHostingEnvironment> _configureConfigurationBuilderDelegate;
```
Tratcher: ConfigureServices and ConfigureLogging allow multiple calls. I understand why creating the logger factory would not, but configuring configuration easily could.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -189,20 +239,32 @@ private IServiceCollection BuildCommonServices()
             var services = new ServiceCollection();
             services.AddSingleton(_hostingEnvironment);
 
+            var builder = new ConfigurationBuilder();
```
Tratcher: this should get seeded with _config.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -189,20 +239,32 @@ private IServiceCollection BuildCommonServices()
...
+            if (_loggerFactoryFunc == null)
             {
-                _loggerFactory = new LoggerFactory();
-                services.AddSingleton(provider => _loggerFactory);
+                loggerFactory = new LoggerFactory();
             }
             else
             {
-                services.AddSingleton(_loggerFactory);
+                loggerFactory = _loggerFactoryFunc(_hostingEnvironment, _config);
```
Tratcher: null check on user results? Or collapse the whole thing to:
`loggerFactory = _loggerFactoryFunc?(_hostingEnvironment, _config) ?? new LoggerFactory();`

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
glennc: I didn't do it because ordering of configuration providers is so important. Allowing multiples means that extension methods can add a config provider to the very end of the list of providers, overriding keys from any added before it. I am not against it completely, but it wasn't immediately obvious to me that it was a good thing.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
glennc: Yeah, it didn't appear that config had a way to do that already. So we need to either add a feature to config or manually do it here. It might be better to do it manually, add the scoped environment variables provider as the first thing then apply any SetSettings that have been used against it.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
glennc: I'll collapse it.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
Tratcher: https://github.com/aspnet/Configuration/issues/630

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -39,6 +40,13 @@ public interface IWebHostBuilder
         IWebHostBuilder ConfigureLogging(Action<ILoggerFactory> configureLogging);
 
         /// <summary>
+        /// Adds a delegate for configuring the provided <see cref="ILoggerFactory"/>. This may be called multiple times.
+        /// </summary>
+        /// <param name="configureLogging">The delegate that configures the <see cref="ILoggerFactory"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder ConfigureLogging<T>(Action<T> configureLogging) where T : ILoggerFactory;
```
davidfowl: This *does* bug me a bunch. The fact that your delegate may never run because T wasn't what you expected but maybe that's fine. We never figured out what to do here right?

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -52,5 +60,21 @@ public interface IWebHostBuilder
...
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder UseLoggerFactory(Func<IHostingEnvironment, IConfiguration, ILoggerFactory> createLoggerFactory);
+
+
+        /// <summary>
+        /// Adds a delegate for configuring the <see cref="IConfigurationBuilder"/> that will construct an <see cref="IConfiguration"/>.
+        /// </summary>
+        /// <param name="configureDelegate">The delegate for configuring the <see cref="IConfigurationBuilder" /> that will be used to construct an <see cref="IConfiguration" />.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder UseConfiguration(Action<IConfigurationBuilder, IHostingEnvironment> configureDelegate);
```
davidfowl: Did we decide to split app and host config?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -94,7 +95,7 @@ public IWebHostBuilder UseLoggerFactory(ILoggerFactory loggerFactory)
                 throw new ArgumentNullException(nameof(loggerFactory));
             }
 
-            _loggerFactory = loggerFactory;
+            _loggerFactoryFunc = (_, __) => loggerFactory;
```
davidfowl: c# 7 lets you have multiple`_`  arguments to mean ignore

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -132,6 +133,55 @@ public IWebHostBuilder ConfigureLogging(Action<ILoggerFactory> configureLogging)
...
+        /// <param name="configureLogging">The delegate that configures the <see cref="ILoggerFactory"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public IWebHostBuilder ConfigureLogging<T>(Action<T> configureLogging) where T: ILoggerFactory
+        {
+            if (configureLogging == null)
+            {
+                throw new ArgumentNullException(nameof(configureLogging));
+            }
+
+            _configureLoggingDelegates.Add(factory => configureLogging((T)factory));
```
davidfowl: This should only get called if the T matches

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -52,5 +60,21 @@ public interface IWebHostBuilder
...
         /// <returns>The value the setting currently contains.</returns>
         string GetSetting(string key);
+
+        /// <summary>
+        /// Adds a delegate to construct the <see cref="ILoggerFactory"/> that will be registered
+        /// as a singleton and used by the application.
+        /// </summary>
+        /// <param name="createLoggerFactory">The delegate that constructs an <see cref="IConfigurationBuilder" /></param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder UseLoggerFactory(Func<IHostingEnvironment, IConfiguration, ILoggerFactory> createLoggerFactory);
```
davidfowl: Let's make a `HostingContext` that has `IConfiguration` and `ILoggerFactory`.

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
glennc: @Tratcher Merge the configs here, making sure SetSettings aren't lost.

Also call SetBasePath or whatever that method is called on the builder before passing it to the funcs.

glennc:
We want to create a `WebHostBuilderContext` that has properties for the things we are consuming in most of these funcs. HostingEnv, Configuration, LoggerFactory are the main three I think.

JunTaoLuo:
:up::date: Will add doc comments

JunTaoLuo:
ping @glennc @davidfowl @Tratcher

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -92,5 +94,22 @@ public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hos
                 services.Replace(ServiceDescriptor.Singleton<IServiceProviderFactory<IServiceCollection>>(new DefaultServiceProviderFactory(options)));
             });
         }
+
+        public static IWebHostBuilder ConfigureLogging(this IWebHostBuilder builder, Action<HostingContext, LoggerFactory> configure)
```
JunTaoLuo: Should the configure delegate take `ILoggerFactory` instead of `LoggerFactory`?

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -92,5 +93,28 @@ public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hos
...
             });
         }
+
+        /// <summary>
+        /// Configures and use a <see cref="LoggerFactory"/> for the web host.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="configure">A callback used to configure the <see cref="LoggerFactory"/> that will be added as a singleton and used by the application.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder ConfigureLogging(this IWebHostBuilder hostBuilder, Action<HostingContext, LoggerFactory> configure)
```
JunTaoLuo: Should this be `ILoggerFactory` instead of `LoggerFactory`?

davidfowl: No, that's part of the change.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -132,6 +133,60 @@ public IWebHostBuilder ConfigureLogging(Action<ILoggerFactory> configureLogging)
...
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public IWebHostBuilder ConfigureLogging<T>(Action<T> configureLogging) where T : ILoggerFactory
+        {
+            if (configureLogging == null)
+            {
+                throw new ArgumentNullException(nameof(configureLogging));
+            }
+            _configureLoggingDelegates.Add(factory =>
+            {
+                if (factory is T)
```
davidfowl: C# 7 has

```
if (factory is T typedFactory)
{
    configureLogging(typedFactory);
}
```

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -29,9 +29,10 @@ public class WebHostBuilder : IWebHostBuilder
         private readonly List<Action<ILoggerFactory>> _configureLoggingDelegates;
 
         private IConfiguration _config;
-        private ILoggerFactory _loggerFactory;
         private WebHostOptions _options;
         private bool _webHostBuilt;
+        private Func<HostingContext, ILoggerFactory> _createLoggerFactoryDelegate;
+        private Action<HostingContext, IConfigurationBuilder> _configureConfigurationBuilderDelegate;
```
davidfowl: This should be a list

davidfowl: On second thought.... Lets not do that right now.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -132,6 +133,60 @@ public IWebHostBuilder ConfigureLogging(Action<ILoggerFactory> configureLogging)
...
+        }
+
+        /// <summary>
+        /// Adds a delegate for configuring the <see cref="IConfigurationBuilder"/> that will construct an <see cref="IConfiguration"/>.
+        /// </summary>
+        /// <param name="configureDelegate">The delegate for configuring the <see cref="IConfigurationBuilder" /> that will be used to construct an <see cref="IConfiguration" />.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public IWebHostBuilder UseConfiguration(Action<HostingContext, IConfigurationBuilder> configureDelegate)
+        {
+            if(configureDelegate == null)
```
davidfowl: nit: spacing

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
glennc: Because you want to do it later, before we ship, or because you are questioning whether we should do it at all?

src/Microsoft.AspNetCore.Hosting.Abstractions/HostingContext.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    /// <summary>
+    /// Context containing the common services on the <see cref="IWebHost" />. Some properties may be null until set by the <see cref="IWebHost" />.
+    /// </summary>
+    public class HostingContext
```
Tratcher: WebHostBuilderContext

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
Tratcher: Lets do multiple. Its weird for one call to silently overwrite another. This way different components can contribute to config.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -185,23 +240,29 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
...
             // Initialize the hosting environment
             _hostingEnvironment.Initialize(applicationName, contentRootPath, _options);
+            hostingContext.HostingEnvironment = _hostingEnvironment;
 
             var services = new ServiceCollection();
             services.AddSingleton(_hostingEnvironment);
 
+            var builder = new ConfigurationBuilder()
+                .SetBasePath(_hostingEnvironment.ContentRootPath)
+                .AddInMemoryCollection(_config.AsEnumerable());
```
Tratcher: We need to look into make this nested rather than copied.

glennc: Can a reload token fire on environment variable changes? Copying this way would mean that you can't get a reload token to fire based on a hosting setting being changed in the environment. Right?

It is probably OK for now since presumably most (all?) hosting config changes would require an app restart to apply.

I guess the alternative would be to add the environment variable provider to this config, the same as the hosting config does, and then apply any UseSettings calls that have been made.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -185,23 +240,29 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
...
+
+            _configureConfigurationBuilderDelegate?.Invoke(hostingContext, builder);
```
Tratcher: should we make the coriginal config available on the context before calling this?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -185,23 +240,29 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
             var appEnvironment = PlatformServices.Default.Application;
             var contentRootPath = ResolveContentRootPath(_options.ContentRootPath, appEnvironment.ApplicationBasePath);
             var applicationName = _options.ApplicationName ?? appEnvironment.ApplicationName;
+            var hostingContext = new HostingContext();
 
             // Initialize the hosting environment
             _hostingEnvironment.Initialize(applicationName, contentRootPath, _options);
+            hostingContext.HostingEnvironment = _hostingEnvironment;
```
Tratcher: amusingly the hosting environment is built from config, before we use it to build more config. That's going to be really confusing if you want to change the config source the environment builds from.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: That will depend on https://github.com/aspnet/Configuration/issues/630.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
JunTaoLuo: Spoke offline, this is not the intended functionality since it overrides previous LoggerFactory while being called ConfigureLogging. Will update.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: Won't be doing this.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: Would it be confusing that that configuration is the one that's about to be overwritten? Also, is it okay to just call build on the builder to get an `IConfiguration` that contains the original config since we just copied it over? Would be unintuitive but not sure how common this use case is.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: Chicken and egg?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: Filed https://github.com/aspnet/Hosting/issues/1004 for future work.

glennc:
From a design standpoint I think this captures everything. I don't think I can actually sign-off on a review because I started the PR, so :shipit:

PR #989, opened 2017-03-25
Remove SkipIfCurrentRuntimeIsCoreClrAttribute

JunTaoLuo:
Found this bug while working on https://github.com/aspnet/IISIntegration/pull/345. Updating to use RuntimeInformation.FrameworkDescription instead


src/Microsoft.AspNetCore.Server.IntegrationTesting/xunit/SkipIfCurrentRuntimeIsCoreClrAttribute.cs
```diff
@@ -17,7 +17,7 @@ public bool IsMet
         {
             get
             {
-                return !Process.GetCurrentProcess().ProcessName.ToLower().Contains("coreclr");
+                return !Process.GetCurrentProcess().ProcessName.ToLower().Contains("dotnet");
```
pranavkm: You could use `RuntimeInformation.FrameworkDescription`. I don't think looking for the process name is the correct check (for instance standalone apps don't have this name).

src/Microsoft.AspNetCore.Server.IntegrationTesting/xunit/SkipIfCurrentRuntimeIsClrAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// Skips a test if the runtime used to run the test is Clr.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public class SkipIfCurrentRuntimeIsClrAttribute : Attribute, ITestCondition
+    {
+        public bool IsMet
+        {
+            get
+            {
+                return !Process.GetCurrentProcess().ProcessName.ToLower().Contains("testhost");
```
pranavkm: Where's this used?

JunTaoLuo: I was going to use it in IISIntegration functional tests. Also I don't see why we only have the attribute for skipping CoreCLR but not CLR.

src/Microsoft.AspNetCore.Server.IntegrationTesting/xunit/SkipIfCurrentRuntimeIsCoreClrAttribute.cs
JunTaoLuo: Ah that would be better. I was wondering why this looks at the process name but I guess it's just really old code.

JunTaoLuo:
Apparently there is `FrameworkSkipCondition(RuntimeFrameworks.CLR)` so these attributes are redundant. Will remove

Hosting.sln
```diff
@@ -1,35 +1,34 @@
-
```
davidfowl: What happened here? can you revert this?

JunTaoLuo: VS did something... will revert.

PR #991, opened 2017-03-27
WIP: Simplify hosting configuration with some extension methods

davidfowl:
- Added WaitForShutdown method to IWebHost
- Added Run and RunApplication to IWebHostBuilder to Build and
Start the application in a single method.

After a discussion about what could be done to further simplify some of the startup logic with the `WebHostBuilder` a few things came up. These were some of the methods that we could add to make the startup logic more approachable while not completely diverging from the original API.

Further simplification requires picking kestrel to to be the default server but that won't happen as part of this PR and requires further discussion.

/cc @glennc @Tratcher @DamianEdwards

### DIFF
--- a/samples/SampleStartups/StartupHelloWorld.cs
+++ b/samples/SampleStartups/StartupHelloWorld.cs
@@ -22,7 +22,7 @@ namespace SampleStartups
         public static void MainWithMiddleware()
         {
             var host = new WebHostBuilder()
-                .RunWith(app =>
+                .RunApplication(app =>
                 {
                     // You can add middleware here
                     app.Run(async context =>
--- a/samples/SampleStartups/StartupHelloWorld.cs
+++ b/samples/SampleStartups/StartupHelloWorld.cs
@@ -19,6 +19,32 @@ namespace SampleStartups
             host.WaitForShutdown();
         }
 
+        public static void MainWithPort()
+        {
+            var host = new WebHostBuilder()
+                .Run(8080, async context =>
+                {
+                    await context.Response.WriteAsync("Hello World");
+                });
+
+            host.WaitForShutdown();
+        }
+
+        public static void MainWithMiddlewareWithPort()
+        {
+            var host = new WebHostBuilder()
+                .RunApplication(8080, app =>
+                {
+                    // You can add middleware here
+                    app.Run(async context =>
+                        {
+                            await context.Response.WriteAsync("Hello World");
+                        });
+                });
+
+            host.WaitForShutdown();
+        }
+
         public static void MainWithMiddleware()
         {
             var host = new WebHostBuilder()
--- a/samples/SampleStartups/StartupHelloWorld.cs
+++ b/samples/SampleStartups/StartupHelloWorld.cs
@@ -6,26 +6,32 @@ using Microsoft.AspNetCore.Http;
 
 namespace SampleStartups
 {
-    public class StartupHelloWorld : StartupBase
+    public class Program
     {
-        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public override void Configure(IApplicationBuilder app)
-        {
-            app.Run(async (context) =>
-            {
-                await context.Response.WriteAsync("Hello World!");
-            });
-        }
-
-        // Entry point for the application.
-        public static void Main(string[] args)
+        public static void Main()
         {
             var host = new WebHostBuilder()
-              //.UseKestrel()
-                .UseStartup<StartupHelloWorld>()
-                .Build();
+                .Run(async context =>
+                {
+                    await context.Response.WriteAsync("Hello World");
+                });
 
-            host.Run();
+            host.WaitForShutdown();
+        }
+
+        public static void MainWithMiddleware()
+        {
+            var host = new WebHostBuilder()
+                .RunWith(app =>
+                {
+                    // You can add middleware here
+                    app.Run(async context =>
+                    {
+                        await context.Response.WriteAsync("Hello World");
+                    });
+                });
+
+            host.WaitForShutdown();
         }
     }
 }
--- a/samples/SampleStartups/StartupHelloWorld.cs
+++ b/samples/SampleStartups/StartupHelloWorld.cs
@@ -22,7 +22,7 @@ namespace SampleStartups
         public static void MainWithPort()
         {
             var host = new WebHostBuilder()
-                .Run(8080, async context =>
+                .Run("localhost", 8080, async context =>
                 {
                     await context.Response.WriteAsync("Hello World");
                 });
@@ -33,13 +33,13 @@ namespace SampleStartups
         public static void MainWithMiddlewareWithPort()
         {
             var host = new WebHostBuilder()
-                .RunApplication(8080, app =>
+                .RunApplication("localhost", 8080, app =>
                 {
                     // You can add middleware here
                     app.Run(async context =>
-                        {
-                            await context.Response.WriteAsync("Hello World");
-                        });
+                    {
+                        await context.Response.WriteAsync("Hello World");
+                    });
                 });
 
             host.WaitForShutdown();
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
@@ -110,10 +110,40 @@ namespace Microsoft.AspNetCore.Hosting
         /// Runs a web application with the specified handler
         /// </summary>
         /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to run.</param>
+        /// <param name="port">The port to bind to.</param>
+        /// <param name="handler">A delegate that handles the request.</param>
+        public static IWebHost Run(this IWebHostBuilder hostBuilder, int port, RequestDelegate handler)
+        {
+            var host = hostBuilder.UseUrls($"http://*:{port}/")
+                                  .Configure(app => app.Run(handler))
+                                  .Build();
+            host.Start();
+            return host;
+        }
+
+        /// <summary>
+        /// Runs a web application with the specified handler
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to run.</param>
         /// <param name="configure">The delegate that configures the <see cref="IApplicationBuilder"/>.</param>
         public static IWebHost RunApplication(this IWebHostBuilder hostBuilder, Action<IApplicationBuilder> configure)
         {
             var host = hostBuilder.Configure(configure).Build();
+            host.Start();
+            return host;
+        }
+
+        /// <summary>
+        /// Runs a web application with the specified handler
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to run.</param>
+        /// <param name="port">The port to bind to.</param>
+        /// <param name="configure">The delegate that configures the <see cref="IApplicationBuilder"/>.</param>
+        public static IWebHost RunApplication(this IWebHostBuilder hostBuilder, int port, Action<IApplicationBuilder> configure)
+        {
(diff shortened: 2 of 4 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -92,5 +93,29 @@ public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hos
...
+            host.Start();
+            return host;
+        }
+
+        /// <summary>
+        /// Runs a web application with the specified handler
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to run.</param>
+        /// <param name="configure">The delegate that configures the <see cref="IApplicationBuilder"/>.</param>
+        public static IWebHost RunWith(this IWebHostBuilder hostBuilder, Action<IApplicationBuilder> configure)
```
davidfowl: This name isn't really good. RunAndConfigure? ConfigureAndRun? RunWithConfiguration?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -15,11 +15,86 @@ namespace Microsoft.AspNetCore.Hosting
...
+                host.WaitForShutdown(token, shutdownMessage);
+            }
+        }
+
+        private static void WaitForShutdown(this IWebHost host, CancellationToken token, string shutdownMessage)
+        {
+            var hostingEnvironment = host.Services.GetService<IHostingEnvironment>();
+            var applicationLifetime = host.Services.GetService<IApplicationLifetime>();
+
+            Console.WriteLine($"Hosting environment: {hostingEnvironment.EnvironmentName}");
```
lutzroeder: Can this be implemented without also writing to Console?

davidfowl: There's a whole discussion on this too. I think the default should use the console. Having an options to turn it off would be fine though.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
lutzroeder: Once an IApplicationBuilder is needed it's probably alright to call Configure and Build separately. If another Run API then maybe one that takes Run(handler, port) to avoid the extra UseUrls. Also like the idea of having Kestrel as a default.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -15,11 +15,86 @@ namespace Microsoft.AspNetCore.Hosting
...
+        }
+
+        /// <summary>
+        /// Blocks the calling thread until the web application shuts down.
+        /// </summary>
+        /// <param name="host">The <see cref="IWebHost"/> to wait for shutdown</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static void WaitForShutdown(this IWebHost host, CancellationToken token)
+        {
+            host.WaitForShutdown(token, shutdownMessage: null);
```
davidfowl: Should this dispose the host? Part of the API is about simplicity but it's also generally useful outside of that use case.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
davidfowl: Maybe, I'd like to play around with it a little before dumping it. Needing middleware isn't an advanced scenario, it's what you do after you get hello world working.

davidfowl: As for the port, overload, the assumption would be that it binds to `*` and not localhost (we've had these debates over and over). I'll add it and see what the feedback is.

davidfowl: I don't love these methods. I added port as the first argument but multiple arguments with one being a delegate is always ugly IMO.

```C#
foo.Run(app => { }, 8080);
foo.Run(8080, app=> { });
foo.Listen(8080).Run(app => { });
```

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -92,5 +93,59 @@ public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hos
...
+            return host;
+        }
+
+        /// <summary>
+        /// Runs a web application with the specified handler
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to run.</param>
+        /// <param name="port">The port to bind to.</param>
+        /// <param name="handler">A delegate that handles the request.</param>
+        public static IWebHost Run(this IWebHostBuilder hostBuilder, int port, RequestDelegate handler)
```
lutzroeder: Isn't there some FxCop rule that says keep adding new parameters at the end of the method if name stays the same (agree having the port first makes more sense otherwise).

davidfowl: We don't use FXCop 😄

samples/SampleStartups/StartupHelloWorld.cs
```diff
@@ -6,26 +6,58 @@
...
+                    await context.Response.WriteAsync("Hello World");
+                });
+
+            host.WaitForShutdown();
+        }
+
+        public static void MainWithPort()
+        {
+            var host = new WebHostBuilder()
+                .Run(8080, async context =>
```
Tratcher: I also don't see this extension being used outside of demoware, no real app is this simple. I'd rather have the RunApplication variant as Run.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -15,11 +15,86 @@ namespace Microsoft.AspNetCore.Hosting
...
+        }
+
+        /// <summary>
+        /// Runs a web application and block the calling thread until token is triggered or shutdown is triggered.
+        /// </summary>
+        /// <param name="host">The <see cref="IWebHost"/> to run.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static void Run(this IWebHost host, CancellationToken token)
+        {
+            host.Run(token, shutdownMessage: null);
```
Tratcher: No default shutdown message?

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -92,5 +93,59 @@ public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hos
...
+        {
+            var host = hostBuilder.UseUrls($"http://*:{port}/")
```
Tratcher: Yeah, @blowdart has killed this before. Too many localhost test servers being publicly exposed.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -92,5 +93,59 @@ public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hos
                 services.Replace(ServiceDescriptor.Singleton<IServiceProviderFactory<IServiceCollection>>(new DefaultServiceProviderFactory(options)));
             });
         }
+
+        /// <summary>
+        /// Runs a web application with the specified handler
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to run.</param>
+        /// <param name="handler">A delegate that handles the request.</param>
+        public static IWebHost Run(this IWebHostBuilder hostBuilder, RequestDelegate handler)
```
Tratcher: Why aren't these extensions named Start? the IWebHost.Run extensions all block but these call Start and return.

samples/SampleStartups/StartupHelloWorld.cs
```diff
@@ -6,26 +6,58 @@
...
-        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
-        public override void Configure(IApplicationBuilder app)
+        public static void Main()
         {
-            app.Run(async (context) =>
-            {
-                await context.Response.WriteAsync("Hello World!");
-            });
+            var host = new WebHostBuilder()
+                .Run(async context =>
```
davidfowl: Too ugly. It's really BuildRunStart

samples/SampleStartups/StartupHelloWorld.cs
davidfowl: > We've avoided specifying just a port because of the ambiguity over what the IP/Host is. Localhost? */Any? Loopack? IPv6Loopback? There's never been agreement for what the default host/ip should be.
  
I know, I was there, I'm forcing the discussion by picking `Any`.

> I also don't see this extension being used outside of demoware, no real app is this simple. I'd rather have the RunApplication variant as Run.

Agreed. Or very simple applications.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
davidfowl: This hasn't changed.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
blowdart: No no no no no. Don't make me cry.

davidfowl: I think it has to be this way for compat with other frameworks. 

golang:

> For TCP and UDP, the syntax of laddr is "host:port", like "127.0.0.1:8080". If host is omitted, as in ":8080", Listen listens on all available interfaces instead of just the interface with the given host address.


nodejs

> Begin accepting connections on the specified port and hostname. If the hostname is omitted, the server will accept connections directed to any IPv4 address (INADDR_ANY).

As a compromise, we could require both host and port.

samples/SampleStartups/StartupHelloWorld.cs
blowdart: Yea this is dangerous. to me it reads like "bind to any nic, on any IP I have". That's going to be a nasty surprise for someone. Be specific here.

davidfowl: I've added a required hostname.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
Tratcher: We already require host and port (and scheme). Unless you're going to accept them a separate parameters I don't see this extension adding much value over UseUrls.

davidfowl: You don't but others do. Everything here is doable already this isn't about functionality. If it was we wouldn't make new APIs. It's about ease of use, API design, the right nomenclature and common parameters used to Start, Listen, Run a simple server.

samples/SampleStartups/StartupHelloWorld.cs
lutzroeder: url instead of port is probably ok. UseUrls is a little bit more but having Run(...) or some similar method just take the handler instead of calling Configure is a great simplification.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
lutzroeder: Tend to agree. Taking url instead of just the port makes sense given it's consistent overall.

davidfowl:
Will revisit.

DamianEdwards:
We're looking at doing something in the meta-package: aspnet/MetaPackages#24

Issue #968, opened 2017-03-20
Server addresses configuration enhancements


JunTaoLuo:
- Add `PreferConfig/PreferHostingUrls` to `IServerAddressesFeature` with a default value of false
- Add extension to `WebHostBuilder` to set this flag

natemcmaster:
Followup: Kestrel https://github.com/aspnet/KestrelHttpServer/issues/1575

JunTaoLuo:
Followup: HttpSysServer https://github.com/aspnet/HttpSysServer/issues/336

PR #992, opened 2017-03-27
Add PreferHostingUrls to WebHostBuilder and IServerAddressesFeature

JunTaoLuo:
#968 

Let's see if we like this naming.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Features/IServerAddressesFeature.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Features/IServerAddressesFeature.cs
@@ -8,5 +8,7 @@ namespace Microsoft.AspNetCore.Hosting.Server.Features
     public interface IServerAddressesFeature
     {
         ICollection<string> Addresses { get; }
+
+        bool PreferHostingUrls { get; set; }
     }
 }
--- a/test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
+++ b/test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
@@ -67,15 +67,19 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         [Fact]
-        public void NoDefaultAddressesIfNotConfigured()
+        public void NoDefaultAddressesAndDoNotPreferHostingUrlsIfNotConfigured()
         {
-            var host = CreateBuilder().UseServer(this).Build();
-            host.Start();
-            Assert.Equal(false, host.ServerFeatures.Get<IServerAddressesFeature>().Addresses.Any());
+            using (var host = CreateBuilder().UseServer(this).Build())
+            {
+                host.Start();
+                var serverAddressesFeature = host.ServerFeatures.Get<IServerAddressesFeature>();
+                Assert.False(serverAddressesFeature.Addresses.Any());
+                Assert.False(serverAddressesFeature.PreferHostingUrls);
+            }
         }
 
         [Fact]
-        public void UsesLegacyConfigurationForAddresses()
+        public void UsesLegacyConfigurationForAddressesAndDoNotPreferHostingUrlsIfNotConfigured()
         {
             var data = new Dictionary<string, string>
             {
@@ -84,13 +88,17 @@ namespace Microsoft.AspNetCore.Hosting
 
             var config = new ConfigurationBuilder().AddInMemoryCollection(data).Build();
 
-            var host = CreateBuilder(config).UseServer(this).Build();
-            host.Start();
-            Assert.Equal("http://localhost:5002", host.ServerFeatures.Get<IServerAddressesFeature>().Addresses.First());
+            using (var host = CreateBuilder(config).UseServer(this).Build())
+            {
+                host.Start();
+                var serverAddressFeature = host.ServerFeatures.Get<IServerAddressesFeature>();
+                Assert.Equal("http://localhost:5002", serverAddressFeature.Addresses.First());
+                Assert.False(serverAddressFeature.PreferHostingUrls);
+            }
         }
 
         [Fact]
-        public void UsesConfigurationForAddresses()
+        public void UsesConfigurationForAddressesAndDoNotPreferHostingUrlsIfNotConfigured()
         {
             var data = new Dictionary<string, string>
             {
@@ -99,13 +107,17 @@ namespace Microsoft.AspNetCore.Hosting
 
             var config = new ConfigurationBuilder().AddInMemoryCollection(data).Build();
 
-            var host = CreateBuilder(config).UseServer(this).Build();
-            host.Start();
-            Assert.Equal("http://localhost:5003", host.ServerFeatures.Get<IServerAddressesFeature>().Addresses.First());
+            using (var host = CreateBuilder(config).UseServer(this).Build())
+            {
+                host.Start();
+                var serverAddressFeature = host.ServerFeatures.Get<IServerAddressesFeature>();
+                Assert.Equal("http://localhost:5003", serverAddressFeature.Addresses.First());
+                Assert.False(serverAddressFeature.PreferHostingUrls);
+            }
         }
 
         [Fact]
-        public void UsesNewConfigurationOverLegacyConfigForAddresses()
+        public void UsesNewConfigurationOverLegacyConfigForAddressesAndDoNotPreferHostingUrlsIfNotConfigured()
         {
             var data = new Dictionary<string, string>
             {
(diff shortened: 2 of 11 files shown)


src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -47,6 +48,12 @@ public interface IWebHostBuilder
         IWebHostBuilder UseSetting(string key, string value);
 
         /// <summary>
+        /// Indicate that the host should listen on the urls configured on the <see cref="IWebHostBuilder"/> instead of those configured on the <see cref="IServer"/>.
```
cesarbs: URLs

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Features/IServerAddressesFeature.cs
```diff
@@ -8,5 +8,7 @@ namespace Microsoft.AspNetCore.Hosting.Server.Features
     public interface IServerAddressesFeature
     {
         ICollection<string> Addresses { get; }
+
+        bool PreferHostingUrls { get; set; }
```
cesarbs: Why is it publicly settable?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -67,11 +67,13 @@ public void UseStartupThrowsWithNull()
         }
 
         [Fact]
-        public void NoDefaultAddressesIfNotConfigured()
+        public void NoDefaultAddressesAndDoNotPreferHostingUrlsIfNotConfigured()
```
cesarbs: There are orthogonal. Add a new test for `PreferHostingUrls`.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -121,6 +123,31 @@ public void UsesNewConfigurationOverLegacyConfigForAddresses()
         }
 
         [Fact]
+        public void DoNotPreferHostingUrlsWhenNoAddressConfigured()
```
cesarbs: Add test with addressed configured but `PreferHostingUrls()` not called.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -121,6 +123,31 @@ public void UsesNewConfigurationOverLegacyConfigForAddresses()
...
+        {
+            var host = CreateBuilder().UseServer(this).PreferHostingUrls().Build();
+            host.Start();
```
cesarbs: Shouldn't you dispose `host` in all these tests?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
JunTaoLuo: This is just making sure the default is false. I will add the check for the new flag in the other tests where only address is configured.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
JunTaoLuo: Probably, though we never had a problem with it since these are using a fake server, it doesn't consume any ips/ports. The hosts will be GC'ed and disposed automatically.

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Features/IServerAddressesFeature.cs
JunTaoLuo: Hosting needs to set it.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
cesarbs: > This is just making sure the default is false

I understand, but it should be in its own test.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
cesarbs: Ok, it's mostly for consistency - I see other tests disposing it. Not sure all of them do though.

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Features/IServerAddressesFeature.cs
cesarbs: As discussed in person, this is fine.

JunTaoLuo: Actually will try to make this an internal set on the concrete type and remove it from the interface. See if that works

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -47,6 +48,12 @@ public interface IWebHostBuilder
         IWebHostBuilder UseSetting(string key, string value);
 
         /// <summary>
+        /// Indicate that the host should listen on the URLs configured on the <see cref="IWebHostBuilder"/> instead of those configured on the <see cref="IServer"/>.
+        /// </summary>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder PreferHostingUrls();
```
davidfowl: Make it an extension method.

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/Features/IServerAddressesFeature.cs
davidfowl: On the concrete type not the interface.

cesarbs: We discussed about this. The problem is that the feature is set by the server. Hosting sets it on the interface. Would have to cast to do that.

davidfowl: I see

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
JunTaoLuo: Only if we flow this through configuration which exposes it to things like environment variables, ...

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -67,15 +67,19 @@ public void UseStartupThrowsWithNull()
         }
```
davidfowl: The reformatting makes these tests hard to review. Was it needed?

JunTaoLuo: We weren't disposing the hosts created during tests. It shouldn't be a problem but @cesarbs suggested to make it consistent and dispose. Hence the changes.

davidfowl: Those sorts of changes make it super hard to review... Next time do it later 😄

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -115,155 +127,190 @@ public void UsesNewConfigurationOverLegacyConfigForAddresses()
...
+                var serverAddressFeature = host.ServerFeatures.Get<IServerAddressesFeature>();
+                Assert.Equal("http://localhost:5009", serverAddressFeature.Addresses.First());
+                Assert.False(serverAddressFeature.PreferHostingUrls);
+            }
+        }
+
+        [Fact]
+        public void DoNotPreferHostingUrlsWhenNoAddressConfigured()
+        {
+            var host = CreateBuilder().UseServer(this).PreferHostingUrls(true).Build();
```
natemcmaster: Dispose?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
natemcmaster: https://github.com/aspnet/Hosting/pull/992/files?w=1

davidfowl: @natemcmaster That doesn't even help a little bit.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
JunTaoLuo: oops missed one

JunTaoLuo:
cc @glennc @DamianEdwards for naming approval.

natemcmaster:
@JunTaoLuo can you share samples of what users code might look like?

natemcmaster:
I'm not sure `PreferHostingUrls` makes it clear what the user is configuring.

That said, naming is hard. Been thinking about this and can't come up w/ anything better. And `PreferConfigurationFromIWebHostBuilderOverIServer` is certainly not good.

JunTaoLuo:
What would you like to see in a sample? I would imagine the default would look like the following
```c#
{
        .PreferHostingUrls(true)
        .UseKestrel(options => options.Listen(IPAddress.Loopback, 5001))
        .UseContentRoot(Directory.GetCurrentDirectory())
        .useStartup<Startup>()
        .Build();

    host.Run();
}
```
This way, `http://localhost:5001` configured on the server is used if no address is configured by the environment variable. However, if `ASPNETCORE_URLS` is set, then that will override the default configured on Kestrel (once the changes in Kestrel are made).

cesarbs:
@natemcmaster @JunTaoLuo How about `PreferWebHostBuilderUrls`?

davidfowl:
> @natemcmaster @JunTaoLuo How about PreferWebHostBuilderUrls?
 
Downvotes are so passive aggressive. So here's some text describing the downvote.

cesarbs:
Well, I tried :P

natemcmaster:
:shipit:

PR #993, opened 2017-03-27
Log status of port before starting server with SelfHostDeployer

natemcmaster:
I'm trying to make it easier to determine if EADDRINUSE errors in our integration tests are caused because the port truly is in use or because there is some libuv problem binding to addresses.

cref https://github.com/aspnet/Entropy/issues/209

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/PortHelper.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/PortHelper.cs
@@ -3,9 +3,7 @@
 
 using System;
 using System.Diagnostics;
-#if NETSTANDARD1_3
 using System.Runtime.InteropServices;
-#endif
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.IntegrationTesting
@@ -22,7 +20,7 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                 RedirectStandardError = true,
             };
 
-            if (UseNetstat())
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
                 psi.FileName = "cmd";
                 psi.Arguments = $"/C netstat -nq | find \"{port}\"";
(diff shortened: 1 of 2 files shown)


Tratcher:
Sample of the output?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/PortHelper.cs
```diff
@@ -0,0 +1,79 @@
...
+            catch (Exception ex)
+            {
+                logger.LogWarning("Failed to check port status. Executed: {0} {1}\nError: {2}", psi.FileName, psi.Arguments, ex.ToString());
+            }
+            return;
+        }
+
+        private static bool UseNetstat()
+        {
+#if NET46
```
davidfowl: Do we need this?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/PortHelper.cs
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+#if NETSTANDARD1_3
+using System.Runtime.InteropServices;
```
davidfowl: Do we need this?

natemcmaster:
macOS with and without port in use

image

Windows with and without port in use

image

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/PortHelper.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    linesLogged = true;
+                    logger.LogInformation("portstatus: {0}", data.Data);
+                }
+            };
+            process.ErrorDataReceived += (sender, data) =>
+            {
+                if (!string.IsNullOrWhiteSpace(data.Data))
+                {
+                    logger.LogWarning("portstatus: {0}", data.Data);
```
davidfowl: do you need to set `linesLogged` here too?

natemcmaster: stderr would indicate the lsof or netstat command failed to list port info or isn't available on the path.

mikeharder:
@natemcmaster: Looks like this is failing in our CI builds:

```
| ResponseFormats:WebListener:Clr:x64:Portable Warning: Failed to check port status. Executed: cmd /C netstat -nq | find "5087"
|                                                       Error: System.InvalidOperationException: The Process object must have the UseShellExecute property set to false in order to redirect IO streams.
|                                                          at System.Diagnostics.Process.StartWithShellExecuteEx(ProcessStartInfo startInfo)
|                                                          at System.Diagnostics.Process.Start()
|                                                          at Microsoft.AspNetCore.Server.IntegrationTesting.PortHelper.LogPortStatus(ILogger logger, Int32 port)
```

natemcmaster:
I'm going to revert this. It was an attempt to figure out what was causing port conflicts in tests, but it didn't help enough. I think we'll have more success if use dynamic ports.

Issue #947, opened 2017-02-22
Graceful Server Shutdown


Tratcher:
Today IServer is in the services container and a graceful shutdown is triggered by disposing the entire container. However, the container does not guarantee which order services get disposed in so services the server (or requests in flight) depend on may get disposed before the server stops, causing strange errors and making shutdown unstable.

Example scenario; the console logger got disposed before the server, causing exceptions when logging on shutdown:
```
Unhandled Exception: System.AggregateException: An error occurred while writing to logger(s). ---> System.InvalidOperationException: The collection has been marked as complete with regards to additions.
   at System.Collections.Concurrent.BlockingCollection`1.TryAddWithNoTimeValidation(T item, Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at Microsoft.Extensions.Logging.Console.Internal.ConsoleLoggerProcessor.EnqueueMessage(LogMessageEntry message)
   at Microsoft.Extensions.Logging.Console.ConsoleLogger.WriteMessage(LogLevel logLevel, String logName, Int32 eventId, String message, Exception exception)
   at Microsoft.Extensions.Logging.Console.ConsoleLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   --- End of inner exception stack trace ---
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.LoggerExtensions.LogError(ILogger logger, EventId eventId, Exception exception, String message, Object[] args)
   at Microsoft.AspNetCore.Server.HttpSys.LogHelper.LogException(ILogger logger, String location, Exception exception)
   at Microsoft.AspNetCore.Server.HttpSys.HttpSysListener.Dispose(Boolean disposing)
   at Microsoft.AspNetCore.Server.HttpSys.HttpSysListener.Dispose()
   at Microsoft.AspNetCore.Server.HttpSys.MessagePump.Dispose()
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.Dispose()
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.Dispose()
[...]

PR #995, opened 2017-03-28
WebHost Start/Run/StopAsync and IServer.Start/StopAsync

Tratcher:
#947 
Stopping the server before we dispose the container (and server) allows requests to drain gracefully without the chance of encountering already disposed services.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
@@ -166,7 +168,7 @@ namespace Microsoft.AspNetCore.Hosting
         public static IWebHost Start(this IWebHostBuilder hostBuilder, params string[] urls)
         {
             var host = hostBuilder.UseUrls(urls).Build();
-            host.Start();
+            host.StartAsync(CancellationToken.None).GetAwaiter().GetResult();
             return host;
         }
     }
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHost.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Hosting
@@ -24,6 +26,13 @@ namespace Microsoft.AspNetCore.Hosting
         /// <summary>
         /// Starts listening on the configured addresses.
         /// </summary>
-        void Start();
+        Task StartAsync(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Attempt to gracefully stop the host.
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task StopAsync(CancellationToken cancellationToken);
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/WebHostDefaults.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/WebHostDefaults.cs
@@ -16,5 +16,7 @@ namespace Microsoft.AspNetCore.Hosting
         public static readonly string ServerUrlsKey = "urls";
         public static readonly string ContentRootKey = "contentRoot";
         public static readonly string PreferHostingUrls = "preferHostingUrls";
+
+        public static readonly string ShutdownTimeoutKey = "shutdownTimeoutSeconds";
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/exceptions.net45.json
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/exceptions.net45.json
@@ -16,5 +16,24 @@
     "NewTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHostBuilder",
     "NewMemberId": "Microsoft.AspNetCore.Hosting.IWebHostBuilder ConfigureConfiguration(System.Action<Microsoft.AspNetCore.Hosting.WebHostBuilderContext, Microsoft.Extensions.Configuration.IConfigurationBuilder> configureDelegate)",
     "Kind": "Addition"
+  },
+  {
+    "OldTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "OldMemberId": "System.Void Start()",
+    "NewTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewMemberId": "System.Threading.Tasks.Task StartAsync(System.Threading.CancellationToken cancellationToken)",
+    "Kind": "Modification"
+  },
+  {
+    "OldTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewMemberId": "System.Threading.Tasks.Task StartAsync(System.Threading.CancellationToken cancellationToken)",
+    "Kind": "Addition"
+  },
+  {
+    "OldTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewMemberId": "System.Threading.Tasks.Task StopAsync(System.Threading.CancellationToken cancellationToken)",
+    "Kind": "Addition"
   }
 ]
--- a/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Hosting.Server
@@ -21,6 +23,13 @@ namespace Microsoft.AspNetCore.Hosting.Server
         /// </summary>
         /// <param name="application">An instance of <see cref="IHttpApplication{TContext}"/>.</param>
         /// <typeparam name="TContext">The context associated with the application.</typeparam>
-        void Start<TContext>(IHttpApplication<TContext> application);
+        /// <param name="cancellationToken">Indicates if the server startup should be aborted.</param>
+        Task StartAsync<TContext>(IHttpApplication<TContext> application, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Stop processing requests and shut down the server, gracefully if possible.
+        /// </summary>
+        /// <param name="cancellationToken">Indicates if the graceful shutdown should be aborted.</param>
+        Task StopAsync(CancellationToken cancellationToken);
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
@@ -82,6 +82,16 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             }
         }
 
+        public static void ServerShutdownException(this ILogger logger, Exception ex)
+        {
+            if (logger.IsEnabled(LogLevel.Debug))
+            {
+                logger.LogDebug(
+                    eventId: LoggerEventIds.ServerShutdownException,
+                    exception: ex,
+                    message: "Server shutdown exception");
+            }
+        }
 
         private class HostingLogScope : IReadOnlyList<KeyValuePair<string, object>>
         {
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -7,6 +7,8 @@ using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting.Builder;
 using Microsoft.AspNetCore.Hosting.Server;
@@ -38,6 +40,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         private IServiceProvider _applicationServices;
         private RequestDelegate _application;
         private ILogger<WebHost> _logger;
+
+        private bool _stopped;
 
         // Used for testing only
         internal WebHostOptions Options => _options;
@@ -100,7 +104,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             }
         }
 
-        public virtual void Start()
+        public virtual async Task StartAsync(CancellationToken cancellationToken)
         {
             HostingEventSource.Log.HostStart();
             _logger = _applicationServices.GetRequiredService<ILogger<WebHost>>();
@@ -112,7 +116,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             _hostedServiceExecutor = _applicationServices.GetRequiredService<HostedServiceExecutor>();
             var diagnosticSource = _applicationServices.GetRequiredService<DiagnosticListener>();
             var httpContextFactory = _applicationServices.GetRequiredService<IHttpContextFactory>();
-            Server.Start(new HostingApplication(_application, _logger, diagnosticSource, httpContextFactory));
+            var hostingApp = new HostingApplication(_application, _logger, diagnosticSource, httpContextFactory);
+            await Server.StartAsync(hostingApp, cancellationToken);
 
             // Fire IApplicationLifetime.Started
             _applicationLifetime?.NotifyStarted();
@@ -267,23 +272,51 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             }
         }
 
-        public void Dispose()
+        public async Task StopAsync(CancellationToken cancellationToken)
         {
+            if (_stopped)
+            {
+                return;
+            }
+            _stopped = true;
+
             _logger?.Shutdown();
+
+            if (!cancellationToken.CanBeCanceled)
+            {
+                cancellationToken = new CancellationTokenSource(Options.ShutdownTimeout).Token;
+            }
 
             // Fire IApplicationLifetime.Stopping
             _applicationLifetime?.StopApplication();
 
+            await Server?.StopAsync(cancellationToken);
+
             // Fire the IHostedService.Stop
             _hostedServiceExecutor?.Stop();
-
-            (_hostingServiceProvider as IDisposable)?.Dispose();
-            (_applicationServices as IDisposable)?.Dispose();
 
             // Fire IApplicationLifetime.Stopped
             _applicationLifetime?.NotifyStopped();
 
             HostingEventSource.Log.HostStop();
+        }
+
+        public void Dispose()
+        {
+            if (!_stopped)
+            {
+                try
+                {
+                    this.StopAsync().GetAwaiter().GetResult();
+                }
+                catch (Exception ex)
+                {
+                    _logger?.ServerShutdownException(ex);
+                }
+            }
+
+            (_applicationServices as IDisposable)?.Dispose();
+            (_hostingServiceProvider as IDisposable)?.Dispose();
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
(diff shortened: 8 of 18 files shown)


src/Microsoft.AspNetCore.TestHost/TestServer.cs
```diff
@@ -102,6 +102,10 @@ void IServer.Start<TContext>(IHttpApplication<TContext> application)
             });
         }
 
+        void IServer.Stop()
+        {
```
davidfowl: We may need to do something here no?

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
```diff
@@ -22,5 +22,10 @@ public interface IServer : IDisposable
         /// <param name="application">An instance of <see cref="IHttpApplication{TContext}"/>.</param>
         /// <typeparam name="TContext">The context associated with the application.</typeparam>
         void Start<TContext>(IHttpApplication<TContext> application);
+
+        /// <summary>
+        /// Stop processing requests and shut down the server, gracefully if possible.
+        /// </summary>
+        void Stop();
```
JunTaoLuo: Consider passing in a timeout? And make it configurable like all the other WebHostOptions?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -813,8 +824,15 @@ public DelegateHostedService(Action started, Action stopping)
 
         private class StartInstance : IDisposable
         {
+            public int StopCalls { get; set; }
```
JunTaoLuo: Where is this used at all?

davidfowl:
@Tratcher Can you add a test to make sure that the container can never dead lock the server shutdown.

Tratcher:
I updated Stop to take a cancellation token but I think we should revert it. It does not flow well up the stack. Stop is called from WebHost.Dispose so the app still can't pass their own token. To expose it I'd have to break WebHost.Dispose into Stop and Dispose, and even then Stop would be called by Run so the user still wouldn't see it. Users not calling Run would most likely overlook Stop and call Dispose, completely bypassing the graceful shutdown. Also, nothing else in the shutdown process can be cancelled. E.g. `_applicationLifetime?.StopApplication();` or `_hostedServiceExecutor?.Stop();`. Refactoring those is way beyond the scope of this issue.

Take a look and let me know.

src/Microsoft.AspNetCore.Hosting.Server.Abstractions/IServer.cs
```diff
@@ -22,5 +23,11 @@ public interface IServer : IDisposable
         /// <param name="application">An instance of <see cref="IHttpApplication{TContext}"/>.</param>
         /// <typeparam name="TContext">The context associated with the application.</typeparam>
         void Start<TContext>(IHttpApplication<TContext> application);
+
+        /// <summary>
+        /// Stop processing requests and shut down the server, gracefully if possible.
+        /// </summary>
+        /// <param name="cancellationToken">Indicates if the graceful shutdown should be aborted.</param>
+        void Stop(CancellationToken cancellationToken);
```
halter73: Can we make Start and Stop async?

@davidfowl

src/Microsoft.AspNetCore.Hosting.Abstractions/WebHostDefaults.cs
```diff
@@ -16,5 +16,7 @@ public static class WebHostDefaults
         public static readonly string ServerUrlsKey = "urls";
         public static readonly string ContentRootKey = "contentRoot";
         public static readonly string PreferHostingUrls = "preferHostingUrls";
+
+        public static readonly string ShutdownTimeoutKey = "shutdownTimeoutSeconds";
```
halter73: @Tratcher Should we move this config out of Kestrel so there is one true place to set this?

Tratcher:
Updated with IWebHost StartAsync, StopAsync, and RunAsync. I did not add any sync-over-async sugar yet so that it would be obvious what the affected code paths are.

TODO:
WebHostApplicationLifetimeEventsOrderedCorrectlyDuringShutdown is failing
RunAsync calls StopAsync without a timeout

samples/SampleStartups/StartupBlockingOnStart.cs
```diff
@@ -39,7 +40,7 @@ public static void Main(string[] args)
 
             using (host)
             {
-                host.Start();
+                host.StartAsync().Wait();
```
davidfowl: GetAwaiter().GetResult() all over

Tratcher:
I've added back the synchronous sugar so this should not break any existing apps.

@davidfowl @halter73 ready for a final review.
@glennc you should see this too.

src/Microsoft.AspNetCore.Hosting.Abstractions/exceptions.net45.json
```diff
@@ -16,5 +16,24 @@
...
+    "OldTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "OldMemberId": "System.Void Start()",
+    "NewTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewMemberId": "System.Threading.Tasks.Task StartAsync(System.Threading.CancellationToken cancellationToken)",
+    "Kind": "Modification"
+  },
+  {
+    "OldTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewTypeId": "public interface Microsoft.AspNetCore.Hosting.IWebHost : System.IDisposable",
+    "NewMemberId": "System.Threading.Tasks.Task StartAsync(System.Threading.CancellationToken cancellationToken)",
```
halter73: Why does this need to be marked as a modification *and* addition. @javiercn?

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -267,23 +272,51 @@ private void EnsureServer()
...
         }
+
+        public void Dispose()
+        {
+            if (!_stopped)
+            {
+                this.StopAsync().GetAwaiter().GetResult();
+            }
+
+            (_hostingServiceProvider as IDisposable)?.Dispose();
```
halter73: Shouldn't the _hostingServiceProvider be disposed last?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -15,11 +16,61 @@ namespace Microsoft.AspNetCore.Hosting
...
+        }
+
+        /// <summary>
+        /// Gracefully stops the host.
+        /// </summary>
+        /// <param name="host"></param>
+        /// <returns></returns>
+        public static Task StopAsync(this IWebHost host)
+        {
+            return host.StopAsync(CancellationToken.None);
```
halter73: Do we think it's weird that CancellationToken.None is really cancel after the configured shutdown timeout?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -15,11 +16,61 @@ namespace Microsoft.AspNetCore.Hosting
...
+        }
+
+        /// <summary>
+        /// Attempts to gracefully stop the host with the given timeout.
+        /// </summary>
+        /// <param name="host"></param>
+        /// <param name="timeout"></param>
```
halter73: Might be worth documenting that this is the timeout before active connections will start being aborted.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
```diff
@@ -23,40 +23,12 @@
 
 namespace Microsoft.AspNetCore.Hosting
 {
-    public class WebHostTests : IServer
+    public class WebHostTests
```
halter73: Do you have any tests that call into the synchronous extension methods? Did you run into any issues with the xunit sync context?

src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -82,6 +82,16 @@ public static void Shutdown(this ILogger logger)
             }
         }
 
+        public static void ServerShutdownException(this ILogger logger, Exception ex)
+        {
+            if (logger.IsEnabled(LogLevel.Debug))
+            {
+                logger.LogDebug(
```
halter73: Any tests verifying this actually logs when IServer.StopAsync throws?

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
```diff
@@ -27,6 +28,13 @@ public WebHostOptions(IConfiguration configuration)
             ContentRootPath = configuration[WebHostDefaults.ContentRootKey];
             HostingStartupAssemblies = configuration[WebHostDefaults.HostingStartupAssembliesKey]?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
             PreferHostingUrls = ParseBool(configuration, WebHostDefaults.PreferHostingUrls);
+
+            var timeout = configuration[WebHostDefaults.ShutdownTimeoutKey];
+            if (!string.IsNullOrEmpty(timeout)
+                && int.TryParse(timeout, NumberStyles.None, CultureInfo.InvariantCulture, out var seconds))
+            {
+                ShutdownTimeout = TimeSpan.FromSeconds(seconds);
```
halter73: Any tests for custom ShutdownTimeout config?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -15,11 +16,61 @@ namespace Microsoft.AspNetCore.Hosting
...
+        /// <returns></returns>
+        public static Task StopAsync(this IWebHost host, TimeSpan timeout)
```
halter73: It looks like this overrides `WebHostOptions.ShutdownTimeout`. Any tests verifying that?

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHost.cs
```diff
@@ -24,6 +26,13 @@ public interface IWebHost : IDisposable
...
         /// </summary>
-        void Start();
+        Task StartAsync(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Attempt to gracefully stop the host.
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task StopAsync(CancellationToken cancellationToken);
```
halter73: Do you have any tests that calls Stop or StopAsync instead of Dispose and verifies that server is restartable since the hosting and application service providers aren't disposed?

src/Microsoft.AspNetCore.Hosting.Abstractions/exceptions.net45.json
Tratcher: Too bug: https://github.com/aspnet/BuildTools/issues/154

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
Tratcher: A little, but there wasn't a way to expose the timeout to the extensions.  Really CancellationToken.None means you're not going to be cancelled externally, it doesn't prevent you from cancelling internally. HttpClient does something like this actually, it always chains its own token to yours with the default timeout.

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHost.cs
Tratcher: There's no plan to make the servers restartable.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -267,23 +272,51 @@ private void EnsureServer()
...
             // Fire IApplicationLifetime.Stopping
             _applicationLifetime?.StopApplication();
 
+            try
+            {
+                await Server?.StopAsync(cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                _logger?.ServerShutdownException(ex);
```
halter73: Are we OK with not rethrowing here? The sources of Kestrel crashing bugs currently get surfaced to Program.Main by throwing from IServer.Dispose, but this will likely change to throw from StopAsync. If the _logger isn't well configured, the app developer might never see why the process crashed.

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHost.cs
halter73: 👍 Still, having a test that calls StopAsync and verifies that it is in any way different than Dispose would probably be nice since it's a new method on WebHost.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
halter73: Sounds reasonable. Maybe its worth noting WebHostOptions.ShutdownTimeout in the doc comments for StopAsync(), but I understand CancellationToken.None meaning default timeout even if it's a little gross.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostTests.cs
Tratcher: A few, but I'll switch a few more back. Haven't seen any xunit issues.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
davidfowl: Stop should throw, but dispose shouldn't.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -28,7 +80,13 @@ public static void Run(this IWebHost host)
                     if (!cts.IsCancellationRequested)
                     {
                         Console.WriteLine("Application is shutting down...");
-                        cts.Cancel();
+                        try
+                        {
+                            cts.Cancel();
+                        }
+                        catch (ObjectDisposedException)
```
davidfowl: How does that happen?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -95,7 +153,16 @@ private static void Run(this IWebHost host, CancellationToken token, string shut
                 },
                 applicationLifetime);
 
-                applicationLifetime.ApplicationStopping.WaitHandle.WaitOne();
+                var waitForStop = new TaskCompletionSource<object>();
+                applicationLifetime.ApplicationStopping.Register(() =>
+                {
+                    // Prevent waitForStop from continuing on the Stopping thread to avoid deadlocks.
+                    Task.Run(() => { waitForStop.TrySetResult(null); });
```
davidfowl: Undo and use the brand new TCS API.`new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);`

davidfowl: You can also avoid the closure here but passing state to the RegisterCallback.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -48,26 +106,26 @@ public static void Run(this IWebHost host)
...
+            return host.RunAsync(token, shutdownMessage: null);
         }
 
-        private static void Run(this IWebHost host, CancellationToken token, string shutdownMessage)
+        private static async Task RunAsync(this IWebHost host, CancellationToken token, string shutdownMessage)
         {
             using (host)
             {
-                host.Start();
+                await host.StartAsync(token);
```
davidfowl: We may want to configure await in some places

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
Tratcher: If server stop throws then hosted services don't get stopped. Are you relying on IWebHost.Dispose to Dispose them as registered services?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
Tratcher: This has always been happening, I only noticed when I ran it under the debugger with all first chance exceptions. The CTS is disposed when the using statement exists on line 107, but it's also captured inside this callback closer the fires on assembly unload (which happens after Main exists).

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
halter73: While Dispose shouldn't throw in a perfect world, it's the only way to communicate errors back to Program.Main for any app that doesn't call the new Stop API. I assume there will be plenty of apps that only call Dispose and not Stop considering Dispose calls Stop internally.

Tratcher: Most apps call Run, which now calls Stop so that should report errors.

halter73: Do we think that WebHost.Dispose() should catch any errors from IServer.Stop without rethrowing? If IServer.StopStop is called first, then we don't have to worry about WebHost.Dispose() throwing anyway.

Tratcher: I'm going to move this try/catch/log to Dispose around the call to StopAsync

halter73: How about try/catch/log/dispose services/rethrow?

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -267,23 +272,51 @@ private void EnsureServer()
...
+            }
+
             // Fire IApplicationLifetime.Stopping
             _applicationLifetime?.StopApplication();
 
+            try
+            {
+                await Server?.StopAsync(cancellationToken);
+            }
+            catch (Exception ex)
```
halter73: At least move the catch to Dispose.

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
```diff
@@ -27,6 +28,13 @@ public WebHostOptions(IConfiguration configuration)
             ContentRootPath = configuration[WebHostDefaults.ContentRootKey];
             HostingStartupAssemblies = configuration[WebHostDefaults.HostingStartupAssembliesKey]?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
             PreferHostingUrls = ParseBool(configuration, WebHostDefaults.PreferHostingUrls);
+
+            var timeout = configuration[WebHostDefaults.ShutdownTimeoutKey];
+            if (!string.IsNullOrEmpty(timeout)
```
JunTaoLuo: nit: no need to special case IsNullOrEmpty, TryParse handles that

src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
```diff
@@ -166,7 +168,7 @@ public static IWebHostBuilder PreferHostingUrls(this IWebHostBuilder hostBuilder
         public static IWebHost Start(this IWebHostBuilder hostBuilder, params string[] urls)
```
davidfowl: Are you going to add `StartAsync` here?

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -267,23 +272,51 @@ private void EnsureServer()
...
+            {
+                return;
+            }
+            _stopped = true;
+
             _logger?.Shutdown();
 
+            if (!cancellationToken.CanBeCanceled)
+            {
+                cancellationToken = new CancellationTokenSource(Options.ShutdownTimeout).Token;
```
davidfowl: We should only do this if there is a timeout right? How do you disable the timeout?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
davidfowl: This is weird lets file an issue about this strange behavior and discuss alternatives on the issue.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
davidfowl: Seems wrong. Show me Maybe we shouldn't dispose the token.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -95,7 +153,17 @@ private static void Run(this IWebHost host, CancellationToken token, string shut
                 },
                 applicationLifetime);
 
-                applicationLifetime.ApplicationStopping.WaitHandle.WaitOne();
+                var waitForStop = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+                applicationLifetime.ApplicationStopping.Register(obj =>
+                {
+                    var tcs = (TaskCompletionSource<object>)obj;
```
davidfowl: 👍

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -95,7 +153,17 @@ private static void Run(this IWebHost host, CancellationToken token, string shut
...
+                    tcs.TrySetResult(null);
+                }, waitForStop);
+
+                await waitForStop.Task;
```
davidfowl: Does this men you can call StopApplication from a request thread now? Also see https://github.com/aspnet/KestrelHttpServer/issues/1501 for an example where somebody is trying to implement graceful shutdown. Does this make it any easier?

src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
Tratcher: Wasn't planning on it right now, this method doesn't get much use.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
Tratcher: Timeout.InfinateTimeSpan (-1s) will work here.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
Tratcher: https://github.com/aspnet/Hosting/issues/1016

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
Tratcher: We'll see what shows up in testing.

PR #996, opened 2017-03-29
more logging and more resiliant port selection

anurse:
* For Self-Host, we support starting with port `0` and sniffing the port number from stdout
* For IIS Express, if you ask the deploy to use port `0`, we pick a random port between `2000` and `65535`, try to start the server and if it fails, we try again up to 5 times. Then we sniff the resulting URL from stdout
* MORE LOGGING - We log all the things. This is to prepare for similar changes in aspnet/ServerTests which will write log messages out to a separate file for each individual test so we can more easily diagnose issues with hanging tests.

There is a breaking change here in that `ApplicationDeployer.Deploy` is now `DeployAsync` and `ApplicationDeployerFactory.Create` now takes an `ILoggerFactory` instead of an `ILogger`. I'll be going through repos that use this and submitting PRs to react shortly.


anurse:
@natemcmaster suggested changing the version prefix to help stage the change. There are other packages currently using the `0.3.0` number, but I don't want to rev all of them. This just means that we'll have to change the references to the IntegrationTesting package to use a separate MSBuild property in `dependencies.props` rather than the `AspNetCoreLabs` one they use now.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ProcessExtensions.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.Extensions.Logging;
```
cesarbs: Copyright header.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TaskExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace System.Threading.Tasks
```
cesarbs: Copyright header.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -9,6 +9,7 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting.Common
 {
     public static class TestUriHelper
     {
+        private static readonly Random Random = new Random();
```
cesarbs: nit: add a blank line after this

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -34,8 +35,15 @@ public static int FindFreePort()
 
         public static int FindFreePort(int initialPort)
         {
-            using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
+            Socket socket;
```
cesarbs: ?

anurse: Debugging something, I'll fix

anurse:
:up: :date:

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ProcessExtensions.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace System.Diagnostics
+{
+    public static class ProcessExtensions
```
Tratcher: Extension classes need to be named after their behavior, not the type being extended. Otherwise you can have conflicts.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -36,6 +36,12 @@ public static int FindFreePort(int initialPort)
         {
```
mikeharder: I think `FindFreePort(int initialPort)` should be removed, leaving only `FindFreePort()`.  There's no reason for a caller to specify an initial port.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TaskExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    internal static class TaskExtensions
+    {
+        public static async Task OrTimeout(this Task task, TimeSpan timeout)
+        {
+            var completed = await Task.WhenAny(task, Task.Delay(timeout));
+            if(completed == task)
+            {
+                // Manifest any exception
+                task.GetAwaiter().GetResult();
```
Tratcher: Couldn't you just await the task again?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -36,6 +36,12 @@ public static int FindFreePort(int initialPort)
...
             using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
             {
+                // Set linger to true, but the linger time to zero, which forces the socket to shut down immediately when closed
+                // The default behavior is that linger is off, which actually means that the socket may sit for a small amount
+                // of time before being fully closed and reset.
+                // More here: https://msdn.microsoft.com/en-us/library/windows/desktop/ms737582(v=vs.85).aspx
+                socket.LingerState = new LingerOption(enable: true, seconds: 0);
```
mikeharder: I wouldn't add this unless we know it will help.  I'm not sure it even matters if no client ever connected to the socket.  I'm testing this now.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISDeployer.cs
```diff
@@ -26,44 +27,46 @@ public IISDeployer(DeploymentParameters startParameters, ILogger logger)
         {
         }
 
-        public override DeploymentResult Deploy()
+        public override async Task<DeploymentResult> DeployAsync()
```
Tratcher: I thought you were just going to delete the unused IIS Deployer?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -16,6 +18,11 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
     /// </summary>
     public class IISExpressDeployer : ApplicationDeployer
     {
+        private const string IISExpressRunningMessage = "IIS Express is running.";
+        private const string FailedToInitializeBindingsMessage = "Failed to initialize site bindings";
+        private const string UnableToStartIISExpressMessage = "Unable to start iisexpress.";
+        private readonly static int MaximumAttempts = 5;
```
mikeharder: NIT: `private const` instead of `private readonly static`

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
anurse: The only reason it existed in the first place was because of port conflict issues in IIS, so yeah, as long as we're switching IIS over to always using a dynamic port we can remove it.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TaskExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace System.Threading.Tasks
+{
+    internal static class TaskExtensions
+    {
+        public static async Task OrTimeout(this Task task, TimeSpan timeout)
+        {
+            var completed = await Task.WhenAny(task, Task.Delay(timeout));
+            if(completed == task)
```
cesarbs: nit: `if (`

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISDeployer.cs
anurse: I can now that we have a smoother plan for larger breaking changes. I wasn't confident at the time that nobody uses it

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TaskExtensions.cs
anurse: Yeah, it just seemed unnecessary to use the sugar again. I can switch if you'd think it'd be clearer

anurse:
:up: :date:

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -16,136 +18,158 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
...
+                    Logger.LogError("Error occurred while starting the process. Exception: {exception}", ex.ToString());
+                }
 
-            try
-            {
-                HostProcess.Start();
-                HostProcess.BeginErrorReadLine();
-                HostProcess.BeginOutputReadLine();
-            }
-            catch (Exception ex)
```
kichalla: I recall adding this when the I couldn't get enough information from the logs as the process was crashing immediately. Could we add this in the new helper extension that you have in this PR?

kichalla: Oops. Ignore this. I see its already there. Not sure if we want to add it to the new extension itself.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -23,27 +23,22 @@ public static Uri BuildTestUri(string hint)
...
+                    return new UriBuilder(uriHint) { Port = FindFreePort() }.Uri;
+                }
+                else
+                {
+                    return uriHint;
+                }
             }
         }
 
         public static int FindFreePort()
```
mikeharder: I would copy the method from Kestrel for consistency, we're already using this method in two places.  The code is exactly the same as what you have here.

https://github.com/aspnet/KestrelHttpServer/blob/dev/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs#L508

anurse:
:up: :date:

anurse:
Since this is a change that requires associated changes to other repos, and I may have messed something up, **and** I'm not able to be attentive to the build this evening, I'm going to wait until tomorrow to merge this ;)

anurse:
There was a hang in the Linux Travis build caused by an incorrect expectation that we'd always print a URL. However there is a test in here that only runs on Ubuntu and uses `NoopServer`, which never actually binds to a port, so no URL is printed. I updated the condition we look for in stdout to fix.

PR #998, opened 2017-03-30
fix nginx deployer

anurse:
I verified that this fixes the issue in ServerTests. Going to merge once I have green builds because the change is minimal and there is a time sensitivity here (old builds of IntegrationTesting will fall off the feed soon and consumers will be broken and I'd like to avoid pinning them)


test/Microsoft.AspNetCore.Hosting.TestSites/Program.cs
```diff
@@ -6,6 +6,7 @@
 using System.Threading;
 using Microsoft.AspNetCore.Hosting.Server;
 using Microsoft.AspNetCore.Http.Features;
+using System;
```
Tratcher: ?

anurse: Not sure, will clean up

PR #1005, opened 2017-04-04
Add test logging helpers to Hosting

anurse:
These are based off the ones in ServerTests. Now that I've seen them at work, I want to add these to Entropy because we're seeing some hanging tests, but I don't want copy-pasta so I figured I'd make it common in Hosting.

I do plan to make this even more low-impact by adding some xunit extensibility to make it "magic", but I don't want to do that until I've worked out the logging process on a few tests.

There's a particular spot in here I'm not so proud of, but very open to suggestions ;). See if you can spot it!

Note: At some point I plan to pull this to somewhere even more common (Common?) in order to be able to use it in all our tests. Also, I plan to use some xunit hooks to avoid some of the boilerplate.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/LoggingHandler.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/LoggingHandler.cs
@@ -1,0 +1,25 @@
+﻿using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting
+{
+    internal class LoggingHandler : DelegatingHandler
+    {
+        private ILogger _logger;
+
+        public LoggingHandler(ILoggerFactory loggerFactory, HttpMessageHandler innerHandler) : base(innerHandler)
+        {
+            _logger = loggerFactory.CreateLogger<HttpClient>();
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            _logger.LogDebug("Sending {method} {url}", request.Method, request.RequestUri);
+            var response = await base.SendAsync(request, cancellationToken);
+            _logger.LogDebug("Received {statusCode} {reasonPhrase} {url}", response.StatusCode, response.ReasonPhrase, request.RequestUri);
+            return response;
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/LoggingHandler.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/LoggingHandler.cs
@@ -1,4 +1,5 @@
-﻿using System.Net.Http;
+﻿using System;
+using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
@@ -17,9 +18,17 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
             _logger.LogDebug("Sending {method} {url}", request.Method, request.RequestUri);
-            var response = await base.SendAsync(request, cancellationToken);
-            _logger.LogDebug("Received {statusCode} {reasonPhrase} {url}", response.StatusCode, response.ReasonPhrase, request.RequestUri);
-            return response;
+            try
+            {
+                var response = await base.SendAsync(request, cancellationToken);
+                _logger.LogDebug("Received {statusCode} {reasonPhrase} {url}", response.StatusCode, response.ReasonPhrase, request.RequestUri);
+                return response;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(0, ex, "Exception while sending '{method} {url}'", request.Method, request.RequestUri, request.RequestUri);
+                throw;
+            }
         }
     }
 }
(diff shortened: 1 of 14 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj
```diff
@@ -20,11 +20,14 @@
...
     <PackageReference Include="Microsoft.Extensions.FileProviders.Embedded" Version="$(AspNetCoreVersion)" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Testing" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.PlatformAbstractions" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Process.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Extensions.RuntimeEnvironment.Sources" Version="$(AspNetCoreVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.NETCore.Windows.ApiSets" Version="$(WindowsApiSetsVersion)" />
+    <PackageReference Include="Serilog.Extensions.Logging" Version="1.4.0" />
+    <PackageReference Include="Serilog.Sinks.File" Version="3.2.0" />
```
Tratcher: Does @eilon know you're using these versions

Tratcher: Even these dependency versions go in the dependencies.props

src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj
```diff
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <Description>ASP.NET Core helpers to deploy applications to IIS Express, IIS, WebListener and Kestrel for testing.</Description>
     <VersionPrefix>0.4.0</VersionPrefix>
-    <TargetFrameworks>net46;netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>net46;netstandard1.5</TargetFrameworks>
```
Tratcher: What did you need from 1.5?

anurse: `Assembly.GetEntryAssembly`

src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj
anurse: Yes, they were in the original ServerTests code that this is ported from.

anurse: @Eilon sorry if they need to be re-submitted when moved like this.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj
anurse: And since these are only used in tests which target `netcoreapp1.1` and we don't ship the package, this is a fully compatible change (though I will be confirming that before merging)

src/Microsoft.AspNetCore.Server.IntegrationTesting/xunit/TestLogging.cs
```diff
@@ -0,0 +1,153 @@
+﻿using System;
```
JunTaoLuo: Copyright

src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj
Eilon: From an OSS perspective if it's the same version it's fine.

From a product perspective, make sure the lead/PM.architect for this area agrees that adding these dependencies makes sense (because it affects everyone using the package, of course).

anurse: Makes total sense. This is an internal package, so the impact should be minimal.

anurse:
:up: :date: (and one that adds new reviewable code) - ServerTests also had an HttpClient handler that logged requests which seemed very useful for testing. I moved it to this repo and added a way to ensure that tests can always get an HttpClient with LoggingHandler, pointed at the deployed site.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/LoggingHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public LoggingHandler(ILoggerFactory loggerFactory, HttpMessageHandler innerHandler) : base(innerHandler)
+        {
+            _logger = loggerFactory.CreateLogger<HttpClient>();
+        }
+
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            _logger.LogDebug("Sending {method} {url}", request.Method, request.RequestUri);
+            var response = await base.SendAsync(request, cancellationToken);
```
Tratcher: try/catch/logexception/throw?

anurse: yeah, not a bad idea at all :)

PR #1008, opened 2017-04-06
fix hangs due to uncleared TCS in IntegrationTesting

anurse:
We wait for a message to indicate that the server has started, but don't properly clear the TCS if the server terminates without reaching that state. This PR fixes that, and adds some more timeouts to make sure we don't just hang when we miss things like this.


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -225,7 +229,10 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
                     }
 
                     // Wait for the app to start
-                    if (!await started.Task)
+                    // The timeout here is large, because we don't know how long the test could need
+                    // We cover a lot of error cases above, but I want to make sure we eventually give up and don't hang the build
+                    // just in case we missed one -anurse
+                    if (!await started.Task.OrTimeout(TimeSpan.FromMinutes(10)))
```
cesarbs: Consider replacing uses of `OrTimeout` with `TimeoutAfter`: https://github.com/aspnet/Testing/blob/dev/src/Microsoft.AspNetCore.Testing/TaskExtensions.cs

anurse: Missed that we have that helper. There are a bunch of places using `OrTimeout` that we should clear up. Anyway, I'll update it.

anurse:
Going to merge after addressing @cesarbs 's feedback, since this can cause some bad build hangs.

PR #1019, opened 2017-04-13
Add ConfigureAwait to prevent xunit hangs

Tratcher:
Pairs with https://github.com/aspnet/KestrelHttpServer/pull/1670


Tratcher:
Merging to unblock builds

Issue #1015, opened 2017-04-10
Add WebHostBuilderContext to ConfigureLogging<T> and remove UseLoggerFactory(this IWebHostBuilder hostBuilder, Action<WebHostBuilderContext, LoggerFactory> configure)


JunTaoLuo:
ConfigureLogging<T>(Action<T> configureLogging) is missing the WebHostBuilderContext and cannot handle logging configuration that may be dependent on configuration for example.

Remove the overload of UseLoggerFactory since it provides little additional benefit compared to just using UseLoggerFactory(Func<WebHostBuilderContext, ILoggerFactory> createLoggerFactory) 

cc @glennc

glennc:
Also `ConfigureLogging<LoggerFactory>(Action<LoggerFactory> configureLogging)` and `ConfigureLogging<LoggerFactory>(Action<Context, LoggerFactory> configureLogging)`  extension method, fill in the `T`. That way code like this works:

```csharp
builder.ConfigureLogging(loggerFactory => loggerFactory.AddConsole());
```
```csharp
builder.ConfigureLogging((context, factory) => factory.AddFileLogger(context.Configuration["LogFileName"]));
```

Right?

JunTaoLuo:
Thanks @glennc for the pr https://github.com/aspnet/Hosting/pull/1020

Issue #1014, opened 2017-04-09
Make the HostContext available to extension methods on IWebHostBuilder


davidfowl:
Today it's impossible to get at the `HostingContext` as an extension method on `IWebHostBuilder`, this means you can't flow the `IHostingEnvironment` in your extension methods that hang off the builder.

/cc @JunTaoLuo @glennc

glennc:
I was thinking of a couple of things around this:

1. promote WebHostBuilderContext to abstractions and have it as a property of the builder
1. provide a `UseServices` extension that has both `WebHostBuilderContext` and `ServiceCollection`. So that at least extensions that are just adding services can do it trivially.

What do you think of those?

davidfowl:
I don't think we need 2 types (maybe that was a typo?).  `UseServices` shouldn't be required to get access to the host context, and it should be called `ConfigureServices` right?

JunTaoLuo:
Thanks @glennc  for the PR https://github.com/aspnet/Hosting/pull/1020

PR #1020, opened 2017-04-13
More WebHostBuilderContext overloads

glennc:
Add Context to WebHostBuilder so that extension methods can get to it, and add some missing overloads.

I renamed `ConfigureConfiguration` here so I can get some more feedback on the name, since everyone seems to hate it but nobody has any better suggestions.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
@@ -14,6 +14,15 @@ namespace Microsoft.AspNetCore.Hosting
     public interface IWebHostBuilder
     {
         /// <summary>
+        /// The <see cref="WebHostBuilderContext"/> used during building.
+        /// </summary>
+        /// <remarks>
+        /// Some properties of this type will be null whilst it is being built, most noteably the <see cref="ILoggerFactory"/> will
+        /// be null inside the <see cref="ConfigureAppConfiguration(Action{WebHostBuilderContext, IConfigurationBuilder})"/> method.
+        /// </remarks>
+        WebHostBuilderContext Context { get; }
+
+        /// <summary>
         /// Builds an <see cref="IWebHost"/> which hosts a web application.
         /// </summary>
         IWebHost Build();
@@ -33,6 +42,13 @@ namespace Microsoft.AspNetCore.Hosting
         IWebHostBuilder ConfigureServices(Action<IServiceCollection> configureServices);
 
         /// <summary>
+        /// Specify the delegate that is used to configure the services of the web application.
+        /// </summary>
+        /// <param name="configureServices">The delegate that configures the <see cref="IServiceCollection"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        IWebHostBuilder ConfigureServices(Action<WebHostBuilderContext, IServiceCollection> configureServices);
+
+        /// <summary>
         /// Adds a delegate for configuring the provided <see cref="ILoggerFactory"/>. This may be called multiple times.
         /// </summary>
         /// <param name="configureLogging">The delegate that configures the <see cref="ILoggerFactory"/>.</param>
@@ -44,7 +60,7 @@ namespace Microsoft.AspNetCore.Hosting
         /// </summary>
         /// <param name="configureLogging">The delegate that configures the <see cref="ILoggerFactory"/>.</param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
-        IWebHostBuilder ConfigureLogging<T>(Action<T> configureLogging) where T : ILoggerFactory;
+        IWebHostBuilder ConfigureLogging<T>(Action<WebHostBuilderContext, T> configureLogging) where T : ILoggerFactory;
 
         /// <summary>
         /// Add or replace a setting in the configuration.
@@ -75,6 +91,6 @@ namespace Microsoft.AspNetCore.Hosting
         /// </summary>
         /// <param name="configureDelegate">The delegate for configuring the <see cref="IConfigurationBuilder" /> that will be used to construct an <see cref="IConfiguration" />.</param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
-        IWebHostBuilder ConfigureConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate);
+        IWebHostBuilder ConfigureAppConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate);
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
@@ -14,39 +14,20 @@ namespace Microsoft.AspNetCore.Hosting
     public interface IWebHostBuilder
     {
         /// <summary>
-        /// The <see cref="WebHostBuilderContext"/> used during building.
-        /// </summary>
-        /// <remarks>
-        /// Some properties of this type will be null whilst it is being built, most noteably the <see cref="ILoggerFactory"/> will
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -75,6 +91,6 @@ public interface IWebHostBuilder
         /// </summary>
         /// <param name="configureDelegate">The delegate for configuring the <see cref="IConfigurationBuilder" /> that will be used to construct an <see cref="IConfiguration" />.</param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
-        IWebHostBuilder ConfigureConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate);
+        IWebHostBuilder ConfigureAppConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate);
```
glennc: @DamianEdwards What do you think about this method? We wanted to keep the Configure part since it matches all the existing methods and is nice in intellisense, and using something other than Configuration as the description part seemed wrong. So breaking it up with a word so that it doesn't seem so circular was our best suggestion so far.

JunTaoLuo:
This addresses https://github.com/aspnet/Hosting/issues/1014 and https://github.com/aspnet/Hosting/issues/1015

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -33,6 +42,13 @@ public interface IWebHostBuilder
         IWebHostBuilder ConfigureServices(Action<IServiceCollection> configureServices);
```
JunTaoLuo: We should mark this as obsolete, same with `ConfigureLogging(Action<ILoggerFactory> configureLogging)`

Tratcher: Why? There's nothing wrong with this overload.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -25,23 +25,26 @@ namespace Microsoft.AspNetCore.Hosting
...
+        private readonly List<Action<WebHostBuilderContext, ILoggerFactory>> _configureLoggingDelegates;
 
         private IConfiguration _config;
         private WebHostOptions _options;
         private bool _webHostBuilt;
         private Func<WebHostBuilderContext, ILoggerFactory> _createLoggerFactoryDelegate;
         private List<Action<WebHostBuilderContext, IConfigurationBuilder>> _configureConfigurationBuilderDelegates;
+        private WebHostBuilderContext _hostingContext;
+
+        public WebHostBuilderContext Context { get { return _hostingContext; } }
```
Tratcher: Context => _hostingContext;

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -241,14 +261,15 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
             var appEnvironment = PlatformServices.Default.Application;
             var contentRootPath = ResolveContentRootPath(_options.ContentRootPath, appEnvironment.ApplicationBasePath);
             var applicationName = _options.ApplicationName ?? appEnvironment.ApplicationName;
-            var hostingContext = new WebHostBuilderContext
+
+            _hostingContext = new WebHostBuilderContext
```
Tratcher: _hostingContext should be set in the initial constructor so Context never returns null. You have the initial _config at that point too.

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
glennc: Yeah, I don't think we need to obsolete ConfigureServices, we aren't changing the IServiceCollection interface like ILoggerFactory. Most extension methods on IServiceCollection don't require anything from the Context today, so no reason to force everyone to accept the context that they don't need. You could make an argument for moving it to an extension method instead of being on the main WebHostBuilder though, if it isn't already.

JunTaoLuo: Yea let's at least move it to the extension.

Tratcher: Unnecessary breaking change, just leave it.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -95,26 +95,28 @@ public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hos
...
-                throw new ArgumentNullException(nameof(configure));
-            }
-
-            hostBuilder.UseLoggerFactory(context =>
-            {
-                var loggerFactory = new LoggerFactory();
-                configure(context, loggerFactory);
-                return loggerFactory;
-            });
+            hostBuilder.ConfigureLogging(configureLogging);
```
JunTaoLuo: Is this recursive right now?

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -14,23 +14,27 @@ namespace Microsoft.AspNetCore.Hosting
...
         /// <summary>
-        /// Builds an <see cref="IWebHost"/> which hosts a web application.
+        /// The <see cref="WebHostBuilderContext"/> used during building.
         /// </summary>
-        IWebHost Build();
+        /// <remarks>
+        /// Some properties of this type will be null whilst it is being built, most noteably the <see cref="ILoggerFactory"/> will
+        /// be null inside the <see cref="ConfigureAppConfiguration(Action{WebHostBuilderContext, IConfigurationBuilder})"/> method.
+        /// </remarks>
+        WebHostBuilderContext Context { get; }
```
JunTaoLuo: @Tratcher @glennc @davidfowl @DamianEdwards 

I feel that there is little benefit of having this context on the WebHostBuilder itself since its properties won't be initialized until Build(). We are providing the context in the other methods and extensions now and the WebHostBuilderContext will be resolvable from DI where needed. I plan to remove this, but just wanted to make sure everyone is on the same page first.

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
```diff
@@ -62,19 +72,26 @@ public interface IWebHostBuilder
...
         /// <summary>
-        /// Adds a delegate for configuring the <see cref="IConfigurationBuilder"/> that will construct an <see cref="IConfiguration"/>.
+        /// Adds a delegate to construct the <see cref="ILoggerFactory"/> that will be registered
+        /// as a singleton and used by the application.
         /// </summary>
-        /// <param name="configureDelegate">The delegate for configuring the <see cref="IConfigurationBuilder" /> that will be used to construct an <see cref="IConfiguration" />.</param>
+        /// <param name="createLoggerFactory">The delegate that constructs an <see cref="IConfigurationBuilder" /></param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
-        IWebHostBuilder ConfigureConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate);
+        IWebHostBuilder UseLoggerFactory(Func<WebHostBuilderContext, ILoggerFactory> createLoggerFactory);
```
JunTaoLuo: The churn is unfortunate but I wanted to sort the members alphabetically so overloads are next to each other.

JunTaoLuo:
@Tratcher This is ready for a final look.

src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHostBuilder.cs
davidfowl: Remove it

Tratcher: Agreed

JunTaoLuo:
:up::date:

PR #1021, opened 2017-04-14
[WIP] generic host abstractions

glennc:
This is the beginning of being able to build a more generic host builder and host, it is as far as @bradwilson got before he had to stop for a bit. I just got the api check passing.

Looks like it needs a bit of cleanup. If we are ok with merging it then I can do the cleanup for someone to merge. Though I suppose there is build work to do that I haven't thought about.

It seems like we should get the interface changes we want done sooner rather than later though.


src/Microsoft.Extensions.Hosting.Abstractions/Microsoft.Extensions.Hosting.Abstractions.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <Description>.NET Core hosting and startup abstractions for applications.</Description>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
davidfowl: 1.0

src/Microsoft.AspNetCore.Hosting.Abstractions/exceptions.net45.json
```diff
@@ -35,5 +35,40 @@
...
+  },
+  {
+    "OldTypeId": "public interface Microsoft.AspNetCore.Hosting.IHostingEnvironment",
+    "NewTypeId": "public interface Microsoft.AspNetCore.Hosting.IHostingEnvironment : Microsoft.Extensions.Hosting.IApplicationEnvironment",
+    "Kind": "Modification"
+  },
+  {
+    "OldTypeId": "public interface Microsoft.AspNetCore.Hosting.IApplicationLifetime",
+    "OldMemberId": "System.Void StopApplication()",
+    "Kind": "Removal"
```
glennc: @eilon not sure how this is supposed to work, it seems like I should've been able to say that it was modified and moved to another type in a different project but it wasn't obvious how I could do that.

src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Provides information about the hosting environment an application is running in.
+    /// </summary>
+    public interface IApplicationEnvironment
+    {
+        /// <summary>
+        /// Gets or sets the name of the environment. This property is automatically set by the host to the value
+        /// of the "DOTNETCORE_ENVIRONMENT" environment variable.
```
cwe1ss: DOTNET_ENVIRONMENT since it can also be used on the full framework!?

src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// </summary>
+        string EnvironmentName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the application. This property is automatically set by the host to the assembly containing
+        /// the application entry point.
+        /// </summary>
+        string ApplicationName { get; set; }
+    }
```
cwe1ss: Should `ContentRootPath` and `ContentRootFileProvider` also be added? Are they necessary for loading configuration files etc?

src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs
cwe1ss: also, the dotnet cli uses e.g. "DOTNET_CLI_TELEMETRY_OPTOUT" so I think using just DOTNET_ would be better.

glennc: I am OK with it being `DOTNET_ENVIRONMENT`, especially given the precedence.

src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs
glennc: You can set them on the ConfigurationBuilder if you need the base path for that. So config on its own isn't enough reason to add it. I am not convinced we should add it unless we have some scenarios in mind where you would want ContentRootPath instead of the alternatives. Do you have anything specific or just looking for gaps?

src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs
Tratcher: HOSTING_ENVIRONMENT? It has no effect on the framework, only your app.

src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs
```diff
@@ -1,22 +1,20 @@
...
     /// Defines methods for objects that are managed by the host.
     /// </summary>
     public interface IHostedService
     {
         /// <summary>
-        /// Triggered when the application host has fully started and the server is waiting
-        /// for requests.
+        /// Triggered when the application host has fully started.
         /// </summary>
         void Start();
```
Tratcher: Now that IWebHost has Start/StopAsync, we should consider making these async + cancellable as well.

bradwilson: There are plans for a non-web host and builder, but there will not be a shared abstraction because of the significant differences in the environments.

src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs
glennc: HOSTING_ENVIRONMENT would work as well. But it is very generic, more likely to clash.

src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs
Tratcher: It could be this generic:
```
public interface IHost : IDisposable
{
  IServiceProvider Services { get; }
  Task StartAsync(CancellationToken);
  Task StopAsync(CancellationToken);
}
```
The builder is a more interesting question. Most of it is about logging, config, and services, which seem generic beyond ASP.NET Core.

davidfowl: This seems fine.

cwe1ss: Isn't `IHostBuilder` one of the main reasons for this abstraction? This would allow 3rd party loggers and DI frameworks to build a single `UseSerilog()` / `UseStructureMap()` extension for all hosts.

davidfowl: @cwe1ss what would the return type of those methods? The problem with the builder abstraction is that it doesn't compose very well.

```
public interface IWebHostBuidler : IHostBuilder
{
}

public interface IHostBuilder
{
     IHostBuilder UseSerilog();
}

var webBuilder = new WebHostBuilder() 
                            .UseSerilog()
                            . // now what?
```

cwe1ss: I'm using something very hacky in one of my projects. But maybe it's an acceptable hack for the benefits we gain? 😉 

```csharp
public interface IHostBuilder<THostBuilder>
{
    THostBuilder Instance { get; }
}

public interface IWebHostBuilder : IHostBuilder<IWebHostBuilder>
{
}

public class WebHostBuilder : IWebHostBuilder
{
    public IWebHostBuilder Instance => this;
}

public static class LoggerHostBuilderExtensions
{
    // Only the low-level HostBuilder extensions (logging, DI, configuration) need to use this ugly code.

    public static THostBuilder UseSerilog<THostBuilder>(this IHostBuilder<THostBuilder> builder)
        where THostBuilder : IHostBuilder<THostBuilder>
    {
        // TODO: register Serilog in e.g. builder.Services

        return builder.Instance;
    }
}

public static class WebHostBuilderExtensions
{
    // Extensions for concrete host builders can use the simple code.

    public static IWebHostBuilder UseKestrel(this IWebHostBuilder builder)
    {
        return builder;
    }
}

class Program
{
    static void Main(string[] args)
    {
        // Compiles :)
        new WebHostBuilder()
            .UseSerilog()
            .UseKestrel();
    }
}```

File: src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs (hunk 3 above)
Comment (cwe1ss) [inline src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs]: Note that `Hosting:Environment` is currently marked as [legacy](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L54) in WebHostBuilder. I guess the main reason is the colon but maybe there already was some other discussion regarding the "Hosting" prefix?

File: src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs (hunk 5 above)
Comment (davidfowl) [inline src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs]: Not really, no. We'll leave this out for now and revisit later.

Comment (cwe1ss) [inline src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs]: ok! Out of curiosity, do you see other disadvantages than the ugly UseSerilog... extension method syntax?

Does this mean, there won't be a unified way to plug in logging, DI and configuration?

Comment (davidfowl) [inline src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs]: > Does this mean, there won't be a unified way to plug in logging, DI and configuration?

We're not merging the builders in 2.0. We have plans to take another look at this, but it's not going to land in time for 2.0. The goal of this change is:

- Make sure shared abstractions exist in a single place. Hosting.Extensions.Abstractions is the place where we add types that will be shared across the `WebHostBuilder` and `HostBuilder`.
- Move `IHostedService` to Hosting.Extensions.Abstractions. This is the new primitive which all things will build on top of. In an ideal world, http support is just an `IHostedService` and composes the http pipeline. Adding `IHostedService` in Hosting.Extensions.Abstractions also 
means other hosted services (like a background service `IHostedService`) can be shared in web and non-web scenarios.

The downside is that we need to duplicate a few extension methods (`ConfigureLogging`, `ConfigureServices`, `ConfigureAppConfiguration` etc) temporarily. The way I see it, if things work out, this host becomes the main host and we deprecate `WebHostBuilder`.

Here's the prototype I did a while back https://github.com/aspnet/Hosting/compare/davidfowl/uber-host.

Comment (cwe1ss) [inline src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs]: David, thx for the quick response! Would have been great to see this for 2.0! 

Won't your uber-host "completely break the world" by forcing a completely new pattern or did I understand it wrong?
```csharp
new HostBuilder()
  .UseLoggerFactory(...)
  .UseWebApplication(options => {
    // Would everyone currently targeting IWebHostBuilder have to target 'WebApplication' now?
    options.UseKestrel();
  })
```
Is your goal to allow people to register multiple services like the following?
```csharp
new HostBuilder()
  .UseLoggerFactory()
  .UseWebApplication()
  .UseABackgroundService()
  .UseAnotherBackgroundService();
```
I guess most applications will only host one service so this sounds like a very big breaking change for a very little used use case.

The common `IHostBuilder` approach sounds "less dramatic" and people who need multiple services could still do it by creating multiple hosts. The duplicated code wouldn't be too bad because the host-independent stuff should mostly be extension methods anyway. (e.g. UseSerilog(), UseStructureMap(), ...).

src/Microsoft.AspNetCore.Hosting.Abstractions/exceptions.net45.json
Eilon: This tool is being updated to just have add/remove, and no "modified". cc @smitpatel as FYI.

muratg:
Assigning to @davidfowl

src/Microsoft.Extensions.Hosting.Abstractions/IApplicationEnvironment.cs
Tratcher: `Hosting:Environment` was marked as legacy when we moved all of our keys to the ASPNETCORE_ prefix.

Tratcher:
FYI we're going to take a subset of these changes for preview1: https://github.com/aspnet/Hosting/pull/1051

PR #1029, opened 2017-04-20
Workaround: On xplat machines, provided executable permission to published standalone application executable

kichalla:
I have verified this fix on my Ubuntu VM

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
@@ -39,6 +39,9 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                 if (DeploymentParameters.PublishApplicationBeforeDeployment)
                 {
                     DotnetPublish();
+
+                    // Temporary workaround for https://github.com/dotnet/cli/issues/6286
+                    ProvideExecutePermission();
                 }
 
                 var hintUrl = TestUriHelper.BuildTestUri(DeploymentParameters.ApplicationBaseUriHint);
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -189,5 +192,52 @@ public override void Dispose()
...
+                        chmodProcessStartInfo.FileName,
+                        ex.ToString());
+                }
+
+                chmodProcess.WaitForExit(3 * 1000);
+
+                if (chmodProcess.HasExited && chmodProcess.ExitCode != 0)
+                {
+                    Logger.LogError(
+                        "Host process {processName} {pid} exited with code {exitCode} or failed to start.",
```
kichalla: will fix the error message

src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core helpers to deploy applications to IIS Express, IIS, WebListener and Kestrel for testing.</Description>
-    <VersionPrefix>0.4.0</VersionPrefix>
```
JunTaoLuo: Why the version bump? This isn't really a breaking change.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -189,5 +192,52 @@ public override void Dispose()
...
+                {
+                    Logger.LogError(
+                        "Error occurred while starting the process '{processName}'. Exception: {exception}",
+                        chmodProcessStartInfo.FileName,
+                        ex.ToString());
+                }
+
+                chmodProcess.WaitForExit(3 * 1000);
+
+                if (chmodProcess.HasExited && chmodProcess.ExitCode != 0)
```
JunTaoLuo: Also log if the chmod process did not exit (i.e. hang) though I don't even know if that's even possible.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Microsoft.AspNetCore.Server.IntegrationTesting.csproj
kichalla: Sure, I agree

kichalla:
🆙 📅

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -189,5 +192,62 @@ public override void Dispose()
...
+                }
+
+                var hasExited = chmodProcess.WaitForExit(3 * 1000);
+
+                if (hasExited)
+                {
+                    if (chmodProcess.ExitCode != 0)
+                    {
+                        Logger.LogError(
+                            "Process {processName} with id {pid} exited with code {exitCode} or failed to start.",
```
JunTaoLuo: nit: string interpolation

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -189,5 +192,62 @@ public override void Dispose()
...
+                            chmodProcessStartInfo.FileName,
+                            HostProcess.Id,
+                            HostProcess.ExitCode);
+                    }
+                }
+                else
+                {
+                    Logger.LogError(
+                            "Process {processName} with id {pid} did not exit in the given amount of time.",
```
JunTaoLuo: nit: string interpolation

anurse: So in this case, I don't mind either way because we control the logger, but in general when logging we should **avoid** string interpolation for cases like this. Logging has special support for these named format arguments and loggers that preserves the key-value mapping between the field name and value, so that structured logging sources can keep the data around.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -189,5 +192,62 @@ public override void Dispose()
...
+                        Logger.LogError(
+                            "Process {processName} with id {pid} exited with code {exitCode} or failed to start.",
+                            chmodProcessStartInfo.FileName,
+                            HostProcess.Id,
+                            HostProcess.ExitCode);
+                    }
+                }
+                else
+                {
+                    Logger.LogError(
```
anurse: Kill the process after logging. It's highly unlikely it'll still be around, but we don't want to leave any process alive :)

kichalla: Sure

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
JunTaoLuo: Ah that makes sense.

Issue #1000, opened 2017-03-31
Add support for executing IHostingStartup in the app assembly


danroth27:
There isn't a good way today for tooling to be able to enable features in an ASP.NET Core app (services, config providers, etc.) without having to parse user code. If we added support for executing IHostingStartup implementations that are within the app then tooling could add this kind of functionality by adding a code file with a class implementation to the project. For example, you could have a tooling gesture that sets up the Azure Key Vault config provider for you.

Tratcher:
What's missing, the discovery? It should work if VS sets the env variable to the app assembly name.

danroth27:
For IHostingStartup implementations in the app assembly I think it should work without needing the environment variable. For tooling scenarios you often want to light up features in the app that aren't environmental. For example, I still want to use the key vault config provider even when I'm not running in VS.

danroth27:
@DamianEdwards @davidfowl @muratg 

This could also be useful for configuring authentication given the changed coming to the authentication abstractions.

danroth27:
@DamianEdwards and @davidfowl chatted and agreed we should do this for 2.0.0. We would look for the assembly level attributes in the app by default with an option to turn it off.

@muratg Please assign

Tratcher:
The existence of a single app assembly is a poor architectural assumption that developers keep breaking. E.g. they move controllers or views to a second assembly and they no longer work. We need something that accounts for multi-assembly apps.

danroth27:
Some options:
1. Only scan the app assembly
2. Scan every assembly in the graph
3. Scan a list of assemblies specified by the user where the app assembly is in the list by default

If we did 2 then I think you might need a way to exclude stuff. 

I lean towards 3. Thoughts?

DamianEdwards:
3 seems reasonable to me

davidfowl:
We're not scanning every assembly in the graph by default in hosting 😄. 

> The existence of a single app assembly is a poor architectural assumption that developers keep breaking. E.g. they move controllers or views to a second assembly and they no longer work. We need something that accounts for multi-assembly apps.

I'm going to say no, this feature isn't for those people. It's not a plugin mechanism for libraries. Scanning only the app assembly IMO makes the most sense. We can wait to get more feedback to do anything else.

Tratcher:
Note: We'll do this by prepending IHostingEnvironment.ApplicationName to HostingStartupAssemblies

PR #1033, opened 2017-04-21
Execute IHostingStart instances in the primary assembly

Tratcher:
Also make the sample app runnable.
#1000
@danroth27 

ApplicationName is set when you call UseStartup or Configure (or you can set it explicitly). We take that name and search the same assembly for HostingStartup attributes/implementations and run them.

### DIFF
--- a/samples/SampleStartups/FakeServer.cs
+++ b/samples/SampleStartups/FakeServer.cs
@@ -1,0 +1,38 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Server;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace SampleStartups
+{
+    // We can't reference real servers in this sample without creating a circular repo dependency.
+    // This fake server lets us at least run the code.
+    public class FakeServer : IServer
+    {
+        public IFeatureCollection Features => new FeatureCollection();
+
+        public Task StartAsync<TContext>(IHttpApplication<TContext> application, CancellationToken cancellationToken)
+        {
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            return Task.CompletedTask;
+        }
+
+        public void Dispose()
+        {
+        }
+    }
+
+    public static class FakeServerWebHostBuliderExtensions
+    {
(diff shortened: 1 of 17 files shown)


davidfowl:
Didn't you tests mroe than one attribute in the same assembly?

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
```diff
@@ -26,7 +26,9 @@ public WebHostOptions(IConfiguration configuration)
             Environment = configuration[WebHostDefaults.EnvironmentKey];
             WebRoot = configuration[WebHostDefaults.WebRootKey];
             ContentRootPath = configuration[WebHostDefaults.ContentRootKey];
-            HostingStartupAssemblies = configuration[WebHostDefaults.HostingStartupAssembliesKey]?.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries) ?? new string[0];
+            // Search the primary assembly and configured assemblies.
+            HostingStartupAssemblies = (ApplicationName + ";" + configuration[WebHostDefaults.HostingStartupAssembliesKey] ?? "")
```
JunTaoLuo: `$"{ApplicationName};{configuration[WebHostDefaults.HostingStartupAssembliesKey]}"`

samples/SampleStartups/FakeServer.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System.Threading;
```
JunTaoLuo: I think we add copyright even in samples. Same for the other file.

samples/SampleStartups/FakeServer.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace SampleStartups
+{
+    // We can't reference real servers in this sample without creating a circular repo dependency.
+    // This fake server lets us at least run the code.
+    public class FakeServer : IServer
+    {
+        public IFeatureCollection Features => new FeatureCollection();
+
+        public Task StartAsync<TContext>(IHttpApplication<TContext> application, CancellationToken cancellationToken)
```
JunTaoLuo: nit: => Task.CompletedTask. Same below.

samples/SampleStartups/FakeServer.cs
Tratcher: No, but it's very inconsistent.

Tratcher:
@davidfowl Updated with an opt-out config switch and more tests.

PR #1043, opened 2017-04-25
Expose WebHostBuilderContext in UseDefaultServiceProvider

pakrym:
As part of https://github.com/aspnet/DependencyInjection/issues/481

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
@@ -86,12 +86,12 @@ namespace Microsoft.AspNetCore.Hosting
         /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
         /// <param name="configure">A callback used to configure the <see cref="ServiceProviderOptions"/> for the default <see cref="IServiceProvider"/>.</param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -86,12 +86,12 @@ public static IWebHostBuilder UseStartup(this IWebHostBuilder hostBuilder, Type
         /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
         /// <param name="configure">A callback used to configure the <see cref="ServiceProviderOptions"/> for the default <see cref="IServiceProvider"/>.</param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
-        public static IWebHostBuilder UseDefaultServiceProvider(this IWebHostBuilder hostBuilder, Action<ServiceProviderOptions> configure)
```
Tratcher: Add both overloads to avoid breaking people? Or is this a new 2.0 API?

pakrym: Was added in https://github.com/aspnet/Hosting/commit/d57d729d13c73750027944dda25082693d6b6d7f it's new in 2.0

davidfowl:
@pakrym Add both overloads

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
davidfowl: Add the overload either way.

Issue #1045, opened 2017-04-25
Remove PreferHostingUrls from WebHostOptions and read the value from config during EnsureServer()


JunTaoLuo:
Related to https://github.com/aspnet/IISIntegration/issues/361.

The issue here is that the options are created before configure services and hosting startup assemblies. Now that UseIISIntegration is planned to be called from the hosting startup assembly, this means that the options will be out of date when we read from it in EnsureServer(). The config is always up to date so we should use that to set the PreferHostingUrls property on the IServerAddressesFeature instead. This is currently what we do for server urls due to the same reason and it works.

cc @Tratcher

PR #1046, opened 2017-04-25
Remove PreferHostingUrls from options and read from config instead

JunTaoLuo:
https://github.com/aspnet/Hosting/issues/1045

From the issue:
Now that UseIISIntegration is planned to be called from the hosting startup assembly, this means that the options will be out of date when we read from it in EnsureServer(). The config is always up to date so we should use that to set the PreferHostingUrls property on the IServerAddressesFeature instead. This is currently what we do for server urls due to the same reason and it works.


davidfowl:
This looks bad

JunTaoLuo:
@davidfowl what concerns do you have?

PR #1051, opened 2017-04-26
Move IHostedService to new Microsoft.Extensions.Hosting.Abstractions package

Tratcher:
This takes a subset of https://github.com/aspnet/Hosting/pull/1021 so we can get it into preview1.


src/Microsoft.Extensions.Hosting.Abstractions/IHostedService.cs
```diff
@@ -1,22 +1,20 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-namespace Microsoft.AspNetCore.Hosting
+namespace Microsoft.Extensions.Hosting
```
davidfowl: 👏 I was going to suggest this. Hopefully you fixed SignalR as well 😄

Issue #1050, opened 2017-04-26
IHostingStartup should be called before we call ConfigureAppConfiguration callbacks


pakrym:
Callbacks registered via `builder.ConfigureAppConfiguration` from `IHostingStartup` are not executed right now.

Blocks: https://github.com/aspnet/AzureIntegration/issues/26

/cc @JunTaoLuo @davidfowl @Tratcher

PR #1052, opened 2017-04-26
Run IHostingStartup's before materializing IConfiguration

pakrym:
Fixes https://github.com/aspnet/Hosting/issues/1050

This change should be safe because we don't have any hosting startups that register configuration (or they would have found this bug before) and we are not passing ``IConfiguration into `IHostingStartup` in any way.

/cc @muratg -  approval required

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -280,14 +280,6 @@ namespace Microsoft.AspNetCore.Hosting
             _hostingEnvironment.Initialize(applicationName, contentRootPath, _options);
             _context.HostingEnvironment = _hostingEnvironment;
 
-            var services = new ServiceCollection();
-            services.AddSingleton(_hostingEnvironment);
-            services.AddSingleton(_context);
-
-            var builder = new ConfigurationBuilder()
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -288,20 +288,6 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
...
-            var configuration = builder.Build();
-            services.AddSingleton<IConfiguration>(configuration);
-            _context.Configuration = configuration;
-
-            // The configured ILoggerFactory is added as a singleton here. AddLogging below will not add an additional one.
-            var loggerFactory = _createLoggerFactoryDelegate?.Invoke(_context) ?? new LoggerFactory(configuration.GetSection("Logging"));
-            services.AddSingleton(loggerFactory);
-            _context.LoggerFactory = loggerFactory;
-
             if (!_options.PreventHostingStartup)
```
Tratcher: Move HostingStartupAssemblies all the way up, it only depends on _options

Tratcher: Keep going, all the way to line 275. E.g. the _hostingEnvironment construction feeds directly into the ServiceCollection construction, but now you've split them.

PR #1061, opened 2017-05-02
Fix ConfigureLogging doc comments

JunTaoLuo:
Fixing typos. Could consider updating this for preview1 but could also wait since only the doc comments are wrong.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
@@ -108,10 +108,10 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         /// <summary>
-        /// Adds a delegate for configuring the provided <see cref="ILoggerFactory"/>. This may be called multiple times.
+        /// Adds a delegate for configuring the provided <see cref="LoggerFactory"/>. This may be called multiple times.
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -134,6 +134,10 @@ public static IWebHostBuilder ConfigureLogging(this IWebHostBuilder hostBuilder,
         /// </summary>
         /// <param name="hostBuilder">The <see cref="IWebHostBuilder" /> to configure.</param>
         /// <param name="configureLogging">The delegate that configures the <see cref="ILoggerFactory"/>.</param>
+        /// <typeparam name="T">
+        /// The type of <see cref="ILoggerFactory"/> to configure.
+        /// The delegate will not execute if the type provided does not match the <see cref="ILoggerFactory"/> used by the <see cref="IWebHostBuilder"/>
```
Eilon: Sentences must end with periods 😦

JunTaoLuo: Fixed in: https://github.com/aspnet/Hosting/commit/c812e9ae6fef3a533e99c2bbf569d42134d3e1c1

PR #1064, opened 2017-05-05
Register IApplicationDiscriminator service

Tratcher:
Moving this functionality from DataProtection so we can reverse the dependency.
https://github.com/aspnet/DataProtection/pull/230
https://github.com/aspnet/Security/issues/1201#issuecomment-299541453


src/Microsoft.AspNetCore.Hosting/Internal/ApplicationDiscriminator.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using Microsoft.AspNetCore.DataProtection.Infrastructure;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    public class ApplicationDiscriminator : IApplicationDiscriminator
+    {
+        private readonly IHostingEnvironment _hostingEnvironment;
+
+        public ApplicationDiscriminator(IHostingEnvironment hostingEnv)
```
davidfowl: nit: hostingEnvironment

Issue #981, opened 2017-03-22
Track recent breaking changes in `exceptions.net*.json` files


dougbu:
This repo contains untracked breaking changes. To find and address them
1. Revert 387e2d8
2. Run `.\build.cmd /t:Pack` in projects 387e2d8 touched or `.\build.cmd /t:Package` in the repo root
3. Update `exceptions.net45.json` and `exceptions.netcore.json` to track the changes
4. If not already announced, suggest filing issues at https://github.com/aspnet/Announcements

Eilon:
Moving to preview2.

PR #1070, opened 2017-05-11
Update API check exceptions

JunTaoLuo:
Addresses #981 Re-enabling API checks for M.A.Hosting and M.A.Server.IntegrationTesting.


src/Microsoft.Extensions.Hosting.Abstractions/Microsoft.Extensions.Hosting.Abstractions.csproj
```diff
@@ -8,6 +8,7 @@
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>hosting</PackageTags>
+    <EnableApiCheck>false</EnableApiCheck>
```
JunTaoLuo: We will generate a baseline file for this when we RTM. This is a new package in 2.0.

src/Microsoft.AspNetCore.Server.IntegrationTesting/breakingchanges.netcore.json
```diff
@@ -0,0 +1,78 @@
+[
+    {
+      "TypeId": "public abstract class Microsoft.AspNetCore.Server.Testing.ApplicationDeployer : Microsoft.AspNetCore.Server.Testing.IApplicationDeployer",
```
JunTaoLuo: These removals are due to rename of the namespace from M.A.Server.Testing to M.A.Server.IntegrationTesting.

JunTaoLuo:
Strange that for renames, the API check tool doesn't require `"Kind": "Addition"`  entries

src/Microsoft.AspNetCore.Server.IntegrationTesting/breakingchanges.netcore.json
dougbu: Didn't the assembly change as well? If yes, the baseline file is invalid and should be removed instead.

dougbu:
> Strange that for renames, the API check tool doesn't require "Kind": "Addition" entries

The only additions that can be breaking changes are `abstract` methods added to known types. Every addition to an `interface` falls into this bucket. Otherwise, additions don't matter.

Besides, the tool doesn't have any information about a rename except that the old type or member doesn't exist. How would it figure out `Fred()` had been renamed to `Ginger()`; the new method may have nothing to do with the old? Why would it try?

src/Microsoft.AspNetCore.Server.IntegrationTesting/breakingchanges.netcore.json
JunTaoLuo: The assembly name change happened before the baselines were generated. I don't think we need to update the baseline.

JunTaoLuo:
@dougbu Ah yes, that makes sense.

src/Microsoft.AspNetCore.Server.IntegrationTesting/breakingchanges.netcore.json
dougbu: The baseline should be for 1.0.0, when this project / assembly were named Microsoft.AspNetCore.Server.Testing. Renaming the assembly was _the_ breaking change. The baseline file should have been removed then.

JunTaoLuo: The base lines were for 1.1.0. The timeline is as follows:
1. Assembly changed from M.A.S.Testing to M.A.S.IntegrationTesting in 1.1.0-preview1 (https://github.com/aspnet/Hosting/commit/f8189a2296e56b6b497afb98d49b004e283a93e5)
2. Baselines were created after the assemblies are renamed (https://github.com/aspnet/Hosting/commit/911da31476615a5ce4bfe2c9a59cffeddb8c2422)

I don't think we delete the baselines since it was created with the current assembly names. We should just update the breakingchanges.netcore.json with the removals.

dougbu: Went to confirm the timeline on https://www.nuget.org and realized we've never published either package. Breaking changes aren't relevant and API checks shouldn't be running here.

Revert changes in this project and, ideally, remove the baseline file.

JunTaoLuo:
:up::date:

Issue #994, opened 2017-03-28
Remove TestProjectHelpers from Server.IntegrationTesting


JunTaoLuo:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/xunit/TestProjectHelpers.cs this is only used in one functional test class in Hosting and does not belong in the IntegrationTesting package.

PR #1073, opened 2017-05-16
Remove TestProjectHelpers from Server.IntegrationTesting

JunTaoLuo:
Addresses #994.


Tratcher:
Almost exactly the same code exists in several functional test projects like https://github.com/aspnet/ServerTests/blob/dev/test/ServerComparison.FunctionalTests/Helpers.cs#L12. How about generalizing this rather than removing it?

JunTaoLuo:
Move to Testing? I think that makes more sense than Common.

Tratcher:
Makes sense

JunTaoLuo:
:up::date:

PR #1076, opened 2017-05-19
Target NETStandard2.0 TFM

JunTaoLuo:
Bringing back Microsoft.AspNetCore.Hosting.WindowsServices. (Will update coherence packages.csv)
Cross-compiling M.A.Hosting for `netcoreapp2.0` and `netstandard2.0` since it needs System.Runtime.Loader which is a netcoreapp only package.

cc @Eilon @DamianEdwards

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -16,8 +16,8 @@
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework' AND '$(OutputType)'=='library'">
-    <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework'">
+    <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETStandardLibraryNETFrameworkVersion)" PrivateAssets="All" />
   </ItemGroup>
 
 </Project>
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -3,6 +3,8 @@
     <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
     <InternalAspNetCoreSdkVersion>2.1.0-*</InternalAspNetCoreSdkVersion>
     <NETStandardImplicitPackageVersion>$(BundledNETStandardPackageVersion)</NETStandardImplicitPackageVersion>
+    <NETStandardLibraryNETFrameworkVersion>2.0.0-*</NETStandardLibraryNETFrameworkVersion>
+    <SystemDiagnosticsDiagnosticSourceVersion>4.4.0-*</SystemDiagnosticsDiagnosticSourceVersion>
     <TestSdkVersion>15.3.0-*</TestSdkVersion>
     <WindowsApiSetsVersion>1.0.1</WindowsApiSetsVersion>
     <XunitVersion>2.3.0-beta2-*</XunitVersion>
--- a/src/Microsoft.AspNetCore.Hosting.WindowsServices/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
+++ b/src/Microsoft.AspNetCore.Hosting.WindowsServices/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
@@ -1,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <Description>ASP.NET Core hosting infrastructure and startup logic for web applications running within a Windows service.</Description>
+    <TargetFramework>net461</TargetFramework>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;hosting</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.AspNetCore.Hosting\Microsoft.AspNetCore.Hosting.csproj" />
+    <PackageReference Include="NETStandard.Library.NETFramework" Version="$(NETStandardLibraryNETFrameworkVersion)" />
+    <Reference Include="System.ServiceProcess" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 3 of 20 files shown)


src/Microsoft.AspNetCore.Hosting.WindowsServices/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
pranavkm: Remember to update packages.csv

build/common.props
```diff
@@ -16,8 +16,8 @@
     <PackageReference Include="Internal.AspNetCore.Sdk" Version="$(InternalAspNetCoreSdkVersion)" PrivateAssets="All" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework' AND '$(OutputType)'=='library'">
-    <PackageReference Include="NETStandard.Library" Version="$(BundledNETStandardPackageVersion)" />
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)'=='.NETFramework'">
```
davidfowl: Yea, I'm not sure this is right. I believe the intent is that this is automatically added to the referencing project but it shouldn't be a package dependency that gets burnt into the package.

pranavkm: Does this sound better:

1. For `src` projects that target `.NETFramework`, add a reference to `NS.Library`. This keeps compat with 1.x
2. For samples and tests, we add a reference to `NS.Library.NETFramework`. We might require that they don't specifyy `PrivateAssets` to support publishing tests (plus it really doesn't matter for them).

natemcmaster: Actually, you might need NS.Library.NETFramework to compile projects in `src` as I did in DataProtection. I'm fine with PrivateAssets=All for now, though I'm getting mixed messages about whether this package will be automatic or not. See https://github.com/dotnet/sdk/issues/1220#issuecomment-301956539 and https://github.com/dotnet/standard/issues/328#issuecomment-299579532

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -90,8 +90,13 @@ public static void Run(this IWebHost host)
                     done.Wait();
                 };
 
+#if NETCOREAPP2_0
                 var assemblyLoadContext = AssemblyLoadContext.GetLoadContext(typeof(WebHostExtensions).GetTypeInfo().Assembly);
                 assemblyLoadContext.Unloading += context => shutdown();
```
natemcmaster: Have we tried the `AppDomain.CurrentDomain.ProcessExit` event? This is available in .NET Standard 2.0. On .NET Core 2.0, this event is analogous to AppContext.ProcessExit.

build/dependencies.props
```diff
@@ -1,8 +1,11 @@
 <Project>
   <PropertyGroup>
     <AspNetCoreVersion>2.0.0-*</AspNetCoreVersion>
+    <CoreFxVersion>4.3.0</CoreFxVersion>
```
natemcmaster: IIUC as a part of retargeting for ns2.0, we should start using the 4.4.0-\*. There may be some conflicts between corefx 4.3.0 and NS.Library 2.0.0

build/common.props
pranavkm: Huh. I thought it only complained at runtime that it couldn't find `netstandard.dll`. What happens when you compile a net461 targeting project that does not reference the package?

natemcmaster: It can happen at build time too. For example: Microsoft.AspNetCore.DataProtection.SystemWeb is net461 and references netstandard2.0. Without a PackageRef to NS.Library.NETFramework, it fails to compile

> DataProtectionStartup.cs(51,20): error CS0012: The type 'IServiceProvider' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.

build/dependencies.props
JunTaoLuo: We'll need to mirror more packages from coreclr in that case right?

build/common.props
JunTaoLuo: Without this reference I cannot compile WindowsServices which targets net461.

natemcmaster: Ok, looks like we need NS.Library.NETFramework. Let's mark as PrivateAssets=all until we figure out if it should be burnt into the package or not.

build/dependencies.props
natemcmaster: Yup. Any new packages you need can be added to this project: https://github.com/aspnet/CoreCLR/blob/dev/tools/MirrorLatestCoreCLR/restore-projects/CoreFx.proj

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
JunTaoLuo: I'm not sure @pakrym? I'll give it a try and see if the tests pass at least.

JunTaoLuo: Looks like it works, at least the functional test passes. This would simplify the PR a bit.

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -90,8 +89,7 @@ public static void Run(this IWebHost host)
                     done.Wait();
                 };
 
-                var assemblyLoadContext = AssemblyLoadContext.GetLoadContext(typeof(WebHostExtensions).GetTypeInfo().Assembly);
-                assemblyLoadContext.Unloading += context => shutdown();
+                AppDomain.CurrentDomain.ProcessExit += (sender, eventArgs) => shutdown();
```
pranavkm: Can we make this change in a follow up if we can't get a sign off? There wasn't a test written for this as part of the changeset this was introduced in so we might not be capturing regressions in behavior here.

pranavkm: Either that or have @pakrym \ @davidfowl sign off on this.

JunTaoLuo: It is being tested here: https://github.com/aspnet/Hosting/blob/dev/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs#L25. I asked @pakrym to do some more in depth manual testing to verify the behaviour later though. I would prefer getting this change in with this PR so we don't cross-compile.

JunTaoLuo: @davidfowl gave an offline :shipit:

Issue #1085, opened 2017-05-23
Make IHostedService async, and run before the Server is started


anurse:
In Hubbup, I'm writing an `IHostedService` to load data up from files into memory and reload them at regular intervals. There are two problems with this approach though:

1. `IHostedService.Start` is synchronous, thus I have to block the thread for any asynchronous code. Since `WebHost.StartAsync` is async, I should be able to flow continuation all the way up that call stack
2. `IHostedService.Start` is run AFTER the server is started (https://github.com/aspnet/Hosting/blob/c7bac3133f01aeea5c4e2c084159c5850440a9f8/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L120). This means requests could come in before my data is loaded (in `Start`).

Since this would be a breaking change in a new 2.0 API, I'd like to nominate this for preview 2. /cc @muratg @davidfowl

Tratcher:
@glennc

davidfowl:
I think we should do this for 2.0.0 but not the part where the server order is different from the `IHostedStartup` order. I have plans to move all of the ASP.NET Core logic into an IHostedService implementation.

anurse:
I agree that 2 needs consideration given the change to move the logic in to a Hosted Service. I'd certainly like to avoid making the services order-dependent.

What about adding a third event, `Initialize` which runs before any `Start` methods run? I realize that has could descend into a full lifecycle, but it seems really useful to have a phase of startup process that runs before we start listening to ports. In my case I'm trying to load data up and I don't want to accept client connections until I've loaded that data. If I want to do that in a place that is a) injected with dependencies from the container and b) is guaranteed to run before client's are accepted, my only option is Configure, which is synchronous, and seems like the wrong place for this.

muratg:
Assigning to @JunTaoLuo

JunTaoLuo:
We resolved one part of the discussion but the design for running IHostedService before server start was never finalized. Reopening to track that discussion.

natemcmaster:
+1 to run before server starts. FYI attempting to solve https://github.com/aspnet/DataProtection/issues/233.

anurse:
> I have plans to move all of the ASP.NET Core logic into an IHostedService implementation.

What's the actual plan for this? Is it in 2.0.0? Preview 2 or later? If we're doing it at all we should do it in a preview, but if we're not doing it in 2.0, we should consider moving Hosted Services before Server.

davidfowl:
preview3

[...]

PR #1090, opened 2017-05-26
Make IHostedServiceAsync

JunTaoLuo:
Addresses https://github.com/aspnet/Hosting/issues/1085.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -123,7 +123,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             _applicationLifetime?.NotifyStarted();
 
             // Fire IHostedService.Start
-            _hostedServiceExecutor.Start();
+            await _hostedServiceExecutor.StartAsync(cancellationToken).ConfigureAwait(false);
 
             _logger.Started();
 
@@ -296,7 +296,10 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             }
 
             // Fire the IHostedService.Stop
-            _hostedServiceExecutor?.Stop();
+            if (_hostedServiceExecutor != null)
+            {
+                await _hostedServiceExecutor.StopAsync(cancellationToken).ConfigureAwait(false);
+            }
 
             // Fire IApplicationLifetime.Stopped
             _applicationLifetime?.NotifyStopped();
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -296,7 +296,10 @@ private void EnsureServer()
             }
 
             // Fire the IHostedService.Stop
-            _hostedServiceExecutor?.Stop();
+            if (_hostedServiceExecutor != null)
```
JunTaoLuo: Because `await _hostedServiceExecutor?.StopAsync(cancellationToken)` didn't work. I'll see if I can make this look a little better.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -123,7 +123,7 @@ public void Initialize()
             _applicationLifetime?.NotifyStarted();
 
             // Fire IHostedService.Start
-            _hostedServiceExecutor.Start();
+            await _hostedServiceExecutor.StartAsync(cancellationToken);
```
Tratcher: `.ConfigureAwait(false);`? @halter73

davidfowl: Ah yes

JunTaoLuo: And on the .StopAsync() as well I suppose.

davidfowl: Ya everything ok the start and stop path needs this because of the xunit sync context.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
moozzyk: Do you need this check at all? You resolved _hostedServiceExecutor  as a required service so I don't think it can be null. Also, you don't check when you call StartAsync.

davidfowl: You can call dispose without calling start

moozzyk: For whatever reason I saw _hostedServiceExecutor being set in the ctor not in StartAsync. Need new glasses.

PR #1094, opened 2017-05-30
Moving CertificateLoader from MetaPackages to Hosting.

Tratcher:
Note the cross compile to netcoreapp2.0 and netstandard2.0.

I'll submit a matching PR to Metapackages to remove this code.


davidfowl:
Why is this in hosting? Repository convincence? Why is it AspNetCore instead of Extensions?

PR #1095, opened 2017-05-31
Change Certificates.Configuration to a Sources package

Tratcher:
This only has internal consumers.

Fixes https://github.com/aspnet/Hosting/issues/1098


davidfowl:
Shouldn't this go into Common with the rest of the source packages?

Tratcher:
It can't be tested in common, it needs Logging and Configuration.

davidfowl:
Ugh this is so arbitrary.

Tratcher:
The other idea was to put it in Kestrel's repo so it's at least next to one of its consumers. We've done something similar for StaticFiles & MVC

Tratcher:
Note types need to be made internal now.

Tratcher:
Will rebase onto rel/2.0.0-preview2 before committing

Tratcher:
Merged to rel and dev.

PR #1096, opened 2017-06-01
Call DataProtectionInitializer before server starts

natemcmaster:
Initialize data protection during WebHost.Initialize()

Resolves https://github.com/aspnet/DataProtection/issues/233


natemcmaster:
:watch: https://github.com/aspnet/DataProtection/pull/239

natemcmaster:
We reviewed the layering and dependencies between Hosting and DP. We'll revert some changes we made in 2.0.0-preview1 which will allow us to keep DP stuff in DP.

natemcmaster:
See https://github.com/aspnet/Hosting/pull/1097

PR #1106, opened 2017-06-18
Create a scoped service provider for the call to Configure

davidfowl:
- This allows scoped dependencies to be injected into the Configure method.
It means you can resolve the DbContext or any other scoped service without
the hassle of the CreateScope boiler plate. As a side effect, it also makes
Startup.Configure a bit more testable.

Inspired by issues like https://github.com/aspnet/Mvc/issues/6407 and code like in our template that needs to create a scope in order to run db init logic.

One downside of this change. If you hold onto scoped services past the `Configure` method, they'll explode later. While this would have exploded by default in the new templates, it now works.

### DIFF
--- a/test/Microsoft.AspNetCore.Hosting.Tests/StartupManagerTests.cs
+++ b/test/Microsoft.AspNetCore.Hosting.Tests/StartupManagerTests.cs
@@ -36,6 +36,26 @@ namespace Microsoft.AspNetCore.Hosting.Tests
             Assert.Equal(2, callbackStartup.MethodsCalled);
         }
 
+        [Fact]
+        public void StartupClassMayHaveScopedServicesInjected()
+        {
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddSingleton<IServiceProviderFactory<IServiceCollection>>(new DefaultServiceProviderFactory(new ServiceProviderOptions
+            {
+                ValidateScopes = true
+            }));
+
+            serviceCollection.AddScoped<DisposableService>();
+            var services = serviceCollection.BuildServiceProvider();
+
+            var type = StartupLoader.FindStartupType("Microsoft.AspNetCore.Hosting.Tests", "WithScopedServices");
+            var startup = StartupLoader.LoadMethods(services, type, "WithScopedServices");
+
+            var app = new ApplicationBuilder(services);
+            app.ApplicationServices = startup.ConfigureServicesDelegate(serviceCollection);
+            startup.ConfigureDelegate(app);
+        }
+
         [Theory]
         [InlineData(null)]
         [InlineData("Dev")]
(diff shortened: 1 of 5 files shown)


test/Microsoft.AspNetCore.Hosting.Tests/StartupManagerTests.cs
```diff
@@ -405,5 +425,15 @@ public void ConfigurationMethodCalled(object instance)
                 _configurationMethodCalledList.Add(instance);
             }
         }
+
+        public class DisposableService : IDisposable
+        {
+            public bool Disposed { get; set; }
```
pakrym: Were you going to use this?

davidfowl: Yea, but it's impossible to get to at the instance. I can remove it.

davidfowl: Fixing it.

davidfowl: 🆙 📅

davidfowl:
@pakrym one question I had, do we have a DI spec test that create "nested scopes". Scopes aren't nested in our container but will anything blow up if somebody was creating a scope on top of an existing scoped `IServiceProvider`?

pakrym:
@davidfowl  https://github.com/aspnet/DependencyInjection/blob/1b638a4419329b2ae1abd7a79ba46b520bd53f5e/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs#L348 ?

test/Microsoft.AspNetCore.Hosting.Tests/StartupManagerTests.cs
```diff
@@ -36,6 +36,31 @@ public void StartupClassMayHaveHostingServicesInjected()
...
+
+            var type = StartupLoader.FindStartupType("Microsoft.AspNetCore.Hosting.Tests", "WithScopedServices");
+            var startup = StartupLoader.LoadMethods(services, type, "WithScopedServices");
+            Assert.NotNull(startup.StartupInstance);
+
+            var app = new ApplicationBuilder(services);
+            app.ApplicationServices = startup.ConfigureServicesDelegate(serviceCollection);
+            startup.ConfigureDelegate(app);
+
+            var instance = (StartupWithScopedServices)startup.StartupInstance;
```
pakrym: You could've just registered factory and stored last created object.

divega:
Should we also have a DI spec test showing `IDisposable` services get disposed at the right level with nested scopes? E.g. a variation of `DisposingScopeDisposesService`

Issue #949, opened 2017-02-23
Re-enable 32-bit deployments


pranavkm:
Specifying RuntimeArchitecture.x86 fails due to https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs#L38-L41. The underlying issue might have been addressed since this check was added.

muratg:
@JunTaoLuo could you take a look?

muratg:
Note to self: check current status.

muratg:
@JunTaoLuo @Tratcher do we still not do 32 bit?

JunTaoLuo:
Right, we still are not testing with 32-bit deployments: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs#L40

muratg:
Thanks. Let's look into this after the previews.

PR #1111, opened 2017-06-21
Allow publishing rid specific desktop applications

pranavkm:
Related to #949


pranavkm:
Checking this in to unblock further work in MusicStore.

Issue #1022, opened 2017-04-15
Add WaitForShutdown method


davidfowl:
This makes it easy to start the server and wait for a signal (Ctrl+c/sigint by default).

Tratcher:
Do you mean Run without the console writeline output?

davidfowl:
And without calling start

muratg:
@davidfowl Are you doing this?

muratg:
@JunTaoLuo keeping this in 2.0.0

PR #1114, opened 2017-06-28
Add WaitForShutdown extensions to WebHostExtensions

JunTaoLuo:
Addresses #1022. I have been testing with Ctrl+C and cancellation tokens using the sample app in the MetaPackages repo. I will test SIGTERM as well. Will add unit tests.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
@@ -24,6 +24,32 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         /// <summary>
+        /// Block the calling thread until shutdown is triggered via Ctrl+C or SIGTERM.
+        /// </summary>
+        /// <param name="host">The running <see cref="IWebHost"/>.</param>
+        public static void WaitForShutdown(this IWebHost host)
+        {
+            host.WaitForShutdownAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Returns a Task that completes when shutdown is triggered via the given token, Ctrl+C or SIGTERM.
+        /// </summary>
+        /// <param name="host">The running <see cref="IWebHost"/>.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static async Task WaitForShutdownAsync(this IWebHost host, CancellationToken token = default(CancellationToken))
+        {
+            var done = new ManualResetEventSlim(false);
+            using (var cts = CancellationTokenSource.CreateLinkedTokenSource(token))
+            {
+                AttachCtrlcSigtermShutdown(cts, done, shutdownMessage: string.Empty);
+
+                await host.WaitForTokenShutdownAsync(cts.Token);
+                done.Set();
+            }
+        }
+
+        /// <summary>
         /// Runs a web application and block the calling thread until host shutdown.
         /// </summary>
         /// <param name="host">The <see cref="IWebHost"/> to run.</param>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -55,6 +54,78 @@ public static Task StopAsync(this IWebHost host, TimeSpan timeout)
...
+        {
+            return host.WaitForShutdownAsync(CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Returns a Task that completes when shutdown is triggered via the given token, Ctrl+C or SIGTERM.
+        /// </summary>
+        /// <param name="host">The running <see cref="IWebHost"/>.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static async Task WaitForShutdownAsync(this IWebHost host, CancellationToken token)
```
davidfowl: Why duplicate this? Why not call this in RunAsync?

JunTaoLuo: Not sure if that's worth the effort. There are a few differences:

1. RunAsync(CancellationToken) triggers shutdown only based on the token whereas WaitForShutdownAsync reacts to CTRL+C and SIGTERM by default. I'm not sure if this is a behaviour change we want.
2. RunAsync() matches the behaviour of WaitForShutdownAsync more but it calls RunAsync(CancellationToken) currently and takes advantage of the console logging there. If we were to use WaitForShutdownAsync instead, we need to duplicate a lot of the logging. 

For these reasons, I think it might be better to keep RunAsync and WaitForShutdown separate.

JunTaoLuo: Refactored

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -138,24 +159,61 @@ public static Task RunAsync(this IWebHost host, CancellationToken token)
...
+
+            await waitForStop.Task;
+
+            // WebHost will use its default ShutdownTimeout if none is specified.
+            await host.StopAsync();
+        }
+
+        private static void AttachCtrlCSIGTERMShutdown(CancellationTokenSource cts, ManualResetEventSlim resetEvent, string shutdownMessage)
+        {
+            Action shutdown = () =>
```
pakrym: local function

test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
```diff
@@ -43,6 +44,8 @@ public ShutdownTests(ITestOutputHelper output) : base(output)
                     PublishApplicationBeforeDeployment = true
                 };
 
+                deploymentParameters.EnvironmentVariables.Add(new KeyValuePair<string, string>("ASPNETCORE_STARTMECHANIC", "Run"));
```
davidfowl: nit: `deploymentParameters.EnvironmentVariables["ASPNETCORE_STARTMECHANIC] = "Run";`

JunTaoLuo: Hmmm not quite, this isn't a `IDictionary`, it's a `List<KVP<string,string>>`.

davidfowl: What

JunTaoLuo: Should I change it to an `IDictionary` as part of this change as well? I'll need to check if this breaks things downstream though.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -138,24 +159,61 @@ public static Task RunAsync(this IWebHost host, CancellationToken token)
...
+                tcs.TrySetResult(null);
+            }, waitForStop);
+
+            await waitForStop.Task;
+
+            // WebHost will use its default ShutdownTimeout if none is specified.
+            await host.StopAsync();
+        }
+
+        private static void AttachCtrlCSIGTERMShutdown(CancellationTokenSource cts, ManualResetEventSlim resetEvent, string shutdownMessage)
```
Tratcher: AttachCtrlcSigtermShutdown

test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
JunTaoLuo: Filed #1123 for future cleanup.

test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
```diff
@@ -68,6 +71,56 @@ public ShutdownTests(ITestOutputHelper output) : base(output)
...
+                };
+
+                deploymentParameters.EnvironmentVariables.Add(new KeyValuePair<string, string>("ASPNETCORE_STARTMECHANIC", "WaitForShutdown"));
+
+                using (var deployer = new SelfHostDeployer(deploymentParameters, loggerFactory))
+                {
+                    await deployer.DeployAsync();
+
+                    // Wait for application to start
+                    await Task.Delay(1000);
```
davidfowl: Don't we have a better way to do this?

davidfowl:
It seems like we still attach to ctrl+c and SIGTerm for graceful shutdown even if I pass a cancellation token. That wasn't the case before was it?

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -55,6 +54,51 @@ public static Task StopAsync(this IWebHost host, TimeSpan timeout)
...
+        {
+            host.WaitForShutdownAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Block the calling thread until shutdown is triggered via the given token, Ctrl+C or SIGTERM.
+        /// </summary>
+        /// <param name="host">The running <see cref="IWebHost"/>.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static void WaitForShutdown(this IWebHost host, CancellationToken token)
```
davidfowl: Remove the sync overload with the token.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -55,6 +54,51 @@ public static Task StopAsync(this IWebHost host, TimeSpan timeout)
...
+        {
+            return host.WaitForShutdownAsync(CancellationToken.None);
+        }
+
+        /// <summary>
+        /// Returns a Task that completes when shutdown is triggered via the given token, Ctrl+C or SIGTERM.
+        /// </summary>
+        /// <param name="host">The running <see cref="IWebHost"/>.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static async Task WaitForShutdownAsync(this IWebHost host, CancellationToken token)
```
davidfowl: Our APIs usually have a single overload that's async with a `default(CancellationToken)`. We should fix RunAsync as well.

test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
JunTaoLuo: Seems like these are no longer needed.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
JunTaoLuo: To simplify the rest of this: version A = RunAsync(), B = RunAsync(CancellationToken ct = default(CancellationToken))

It makes sense to make this change for `WaitForShutdownAsync` but why do we need to change `RunAsync`? The rationale for this pattern, to my understanding, is so that we don't need to create the overload without the cancellation token but for RunAsync, the behaviour is fundamentally different between A and B. Now, it is ambiguous what RunAsync() means. Was the intention to use A or B but with the default value?

I was surprised this compiles at all since RunAsync() could potentially resolve to either A or B but the compiler seems to prefer A. I also verified that if A is removed, RunAsync() will resolve to B but will use the default value. But given the compiler prefers A, B will never be invoked with the default value so I don't see why that's necessary to add.

If the goal is to combine the implementation of A and B into one then it will be hard to distinguish between RunAsync() and RunAsync(CancellationToken.None), which today behaves differently.

JunTaoLuo: I made the change since it compiles and doesn't seem to break anything but I think it's rather strange.

PR #1121, opened 2017-06-29
Remove dependency in HttpContextFactory for HostingApplicationTests

davidfowl:
See https://github.com/aspnet/HttpAbstractions/pull/881


pakrym:
You changed it to mock so build does not break?

pakrym:
LGTM, but I don't see why you cannot keep factory as is.

Issue #1124, opened 2017-06-30
IWebHost breaking change


grahamehorner:
the change 62f74d5be0b1f8091742d2d8a3e88f3fdbd05ab2 to IWebHost interface is breaking change; including service-fabric-aspnetcore, please consider re-adding Start() and Stop() to the interface to remain backwards compatible with service-fabric and other code; marking them as obsolete but with out error.

These could simply call the StartAsync, StopAsync under the hood  
https://github.com/Azure/service-fabric-aspnetcore/issues/21

Tratcher:
@DamianEdwards

davidfowl:
We've decided to fix this.

davidfowl:
@grahamehorner Have you tried service fabric with the latest build?

grahamehorner:
I have tried with
 
Service Fabric SDK 2.6.220
Service Fabric Runtime 5.6.220

but I'm still having issues, even the template generated code wont run inside service fabric when targeting ASP.NET Core 2.0

however the expectation with regard IWebHost.Start() is no longer shown?

grahamehorner:
Ok, now tried with both Service Fabric Stable nugget 2.6.220 & ASPNETCORE 2.0 preview 3
and Service Fabric 2.7.0-preview5718 and the service fails to start :'(  service shows the following

Error event: SourceId='System.FM', Property='State'.
Partition is below target replica or instance count.
fabric:/Application1/Web2 1 1 8b4e8ca4-a8ca-4f1b-9d49-6d7f102a79d1
  IB _Node_1 Up 131435483798573745
  (Showing 1 out of 1 replicas. Total available replicas: 0.)

grahamehorner:
still having issues, so I've created a GitHub repo so you can see my experience of hosting ASPNETCORE 2.0 preview3 inside of Service Fabric; not sure where the root issue is yet; 

https://github.com/grahamehorner/ServiceFabric.ASPNETCORE2

PR #1125, opened 2017-07-01
Add synchronous Start back to IWebHost

davidfowl:
- Fixed extension methods to have a default(CancellationToken)
- Removed extension methods for Start, StartAsync and StopAsync

#1124


muratg:
LGTM. @Tratcher?

davidfowl:
There was never a stop, the diff has a Stop but it's somewhere completely unrelated.

grahamehorner:
many thanks guys, I will pull down the nightly build and give this a try and report back 😎 @JunTaoLuo just to answer your query I realised after posting Stop() was never on the interface, but felt I would leave the issue as was posted as it seemed to align nicely with providing non asynchronous alternatives to the methods that were added. I'm happy either way, as the removal of Start() method on the interface was the root issue, and that has been reverted.  Thanks again guys

PR #1126, opened 2017-07-02
Use dictionary for deployment parameters

### LINKED ISSUE #1123: Deployment parameters should use IDictionary instead of IList for environment variables
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs#L106 and https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs#L111 should be changed to be an `IDictionary<string,string>`.

This came out of a discussion from https://github.com/aspnet/Hosting/pull/1114#discussion_r124685631

henkmollema:
Resolves #1123

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
@@ -44,7 +44,7 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
             ApplicationName = new DirectoryInfo(ApplicationPath).Name;
             ServerType = serverType;
             RuntimeFlavor = runtimeFlavor;
-            EnvironmentVariables.Add(new KeyValuePair<string, string>("ASPNETCORE_DETAILEDERRORS", "true"));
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -44,7 +44,7 @@ public class DeploymentParameters
             ApplicationName = new DirectoryInfo(ApplicationPath).Name;
             ServerType = serverType;
             RuntimeFlavor = runtimeFlavor;
-            EnvironmentVariables.Add(new KeyValuePair<string, string>("ASPNETCORE_DETAILEDERRORS", "true"));
+            EnvironmentVariables.Add("ASPNETCORE_DETAILEDERRORS", "true");
```
henkmollema: Should we use `Add` or the dictionary indexer syntax?

davidfowl: Indexer syntax

henkmollema:
@davidfowl updated

PR #1132, opened 2017-07-05
Update comments to reflect code behavior

jstallm:
Update comment so that its clear this value can be influenced by --Environment command line argument

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingEnvironment.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingEnvironment.cs
@@ -12,7 +12,8 @@ namespace Microsoft.AspNetCore.Hosting
     {
         /// <summary>
         /// Gets or sets the name of the environment. This property is automatically set by the host to the value
-        /// of the "ASPNETCORE_ENVIRONMENT" environment variable.
+        /// of the "ASPNETCORE_ENVIRONMENT" environment variable but can be overridden with the command line argument
+        /// --environment
         /// </summary>
         string EnvironmentName { get; set; }
         


jstallm:
I noticed that the value of env.EnvironmentName can be influenced by passing command line arg --environment upon startup. Adding this comment makes it clear that this value is typically the value of the environment variable, but not always.

src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingEnvironment.cs
```diff
@@ -12,7 +12,8 @@ public interface IHostingEnvironment
     {
         /// <summary>
         /// Gets or sets the name of the environment. This property is automatically set by the host to the value
-        /// of the "ASPNETCORE_ENVIRONMENT" environment variable.
+        /// of the "ASPNETCORE_ENVIRONMENT" environment variable but can be overridden with the command line argument
+        /// --environment
```
Tratcher: command line parameters aren't read by default, you have to opt into that.

jstallm: is that opt-in specified in comments anywhere? Whether its this repository or another? I would imagine the best place to put that is here since its not specified anywhere else.

Tratcher: That opt in is via the config system. ASPNETCORE_ app environment variables are added to config by default, but any other config source must be added manually. E.g. you can define environment via command line, config file, etc., but you have to tell WebHostBuilder to use those config sources.
If we were to provide any hint here, it would be "This property is automatically set by the host to the value of the "ASPNETCORE_ENVIRONMENT" variable, or "environment" as specified in any other configuration source".

jstallm: "or 'environment' as specified in any other configuration source"
This is exactly what I was after. Nice revision.  thank you for the clarity.

muratg:
@Tratcher Could you review further and take in or close?  Thanks.

Tratcher:
The current PR is not ready to merge. Justin, can you make the recommended comment changes?

jkotalik:
Closing, see #1182

PR #1138, opened 2017-07-11
Attach the correlation id to the request scope if logging is on

davidfowl:
- Add the correlation id to the request scope as well as setting
it as the parent id of the activity

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
@@ -46,12 +46,22 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 RecordRequestStartEventLog(httpContext);
             }
 
+            var diagnosticListenerEnabled = _diagnosticListener.IsEnabled();
+            var loggingEnabled = _logger.IsEnabled(LogLevel.Critical);
 
-            if (_diagnosticListener.IsEnabled())
+            // If logging is enabled or the diagnostic listener is enabled, try to get the correlation
+            // id from the header
+            StringValues correlationId;
+            if (diagnosticListenerEnabled || loggingEnabled)
+            {
+                httpContext.Request.Headers.TryGetValue(RequestIdHeaderName, out correlationId);
+            }
+
+            if (diagnosticListenerEnabled)
             {
                 if (_diagnosticListener.IsEnabled(ActivityName, httpContext))
                 {
-                    context.Activity = StartActivity(httpContext);
+                    context.Activity = StartActivity(httpContext, correlationId);
                 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -46,12 +46,22 @@ public void BeginRequest(HttpContext httpContext, ref HostingApplication.Context
...
+            var diagnosticListenerEnabled = _diagnosticListener.IsEnabled();
+            var loggingEnabled = _logger.IsEnabled(LogLevel.Critical);
 
-            if (_diagnosticListener.IsEnabled())
+            // If logging is enabled or the diagnostic listener is enabled, try to get the correlation
+            // id from the header
+            StringValues correlationId;
+            if (diagnosticListenerEnabled || loggingEnabled)
+            {
+                httpContext.Request.Headers.TryGetValue(RequestIdHeaderName, out correlationId);
```
davidfowl: We might want to add this as a known header to kestrel since we'll be doing it all the time (if logging is enabled).

src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs
```diff
@@ -119,13 +119,19 @@ public int Count
                     {
                         return new KeyValuePair<string, object>("RequestPath", _httpContext.Request.Path.ToString());
                     }
-                    throw new IndexOutOfRangeException(nameof(index));
+                    else if (index == 2)
+                    {
+                        return new KeyValuePair<string, object>("CorrelationId", _correlationId);
```
ngbrown: If the correlation id in the logs is going to tie requests that travel through any number of services in series, then it needs to be set even at the origin service.

It seems like the `CorrelationId` should be set to `_httpContext.TraceIdentifier` if `_correlationId` is `null`.

davidfowl: No, this never overwrites the request id. That's not right.

ngbrown: I think then that this `CorrelationId` is really a mis-labeled `ParentRequestId` and that an actual `CorrelationId` should be added to reflect the root request id shared across all services that come after the root service that is handling a request made by a user.

This is my understanding of the spec at [FlatRequestId.md](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/FlatRequestId.md#non-hierarchical-request-id-example).  In that documentation, the correlation id is set in Service A and remains the same in all chained services.

I opened a new issue #1384 to this effect.

Issue #1038, opened 2017-04-24
WebSocketClient.ConnectAsync does not throw when the provided CancellationToken is cancelled.


ericwj:
The test written as below fails, though it should succeed.

It attempts to connect to a `TestServer` that is configured to never respond to a web socket request, using its associated `WebSocketClient` with a `CancellationToken` that is set to cancel after one second.

The test waits 2 seconds for an `OperationCanceledException` or a `TaskCancelledException` to be thrown, but no exception is thrown.

```C#
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Http;
using System;
using System.Net.WebSockets;
using System.Threading;
using System.Threading.Tasks;
using Xunit;

namespace WebSocketClientTests
{
	public class TestServerTests
	{
		private static Task Forever => new TaskCompletionSource<object>().Task;
		private WebSocket serverSocket;

		public async Task AcceptAllWebSocketConnections(HttpContext context, Func<Task> next) {
			if (!context.WebSockets.IsWebSocketRequest) {
				context.Response.StatusCode = 400;
				return;
			}
			serverSocket = await context.WebSockets.AcceptWebSocketAsync();
			await Forever;
		}
		public Task NoResponseToAllWebSocketConnections(HttpContext context, Func<Task> next)
			=> Forever;

		[Fact]
		public async Task TestServer_WebSocketClient_ConnectAsync_ThrowsWhenCancelled() {
			var builder = new WebHostBuilder()
				.Configure(app => {
					app.UseWebSockets();
					app.Use(NoResponseToAllWebSocketConnections);
				});
			var uri = new Uri("ws://localhost");
			var ts = new Microsoft.AspNetCore.TestHost.TestServer(builder);
			var wsclient = ts.CreateWebSocketClient();

			var timeout = TimeSpan.FromSeconds(1);
			var gracePeriod = TimeSpan.FromSeconds(1);
			var source = new CancellationTokenSource();

			source.CancelAfter(timeout);
			var delay = Task.Delay(timeout + gracePeriod);
			var connect = wsclient.ConnectAsync(uri, source.Token);
			var either = Task.WhenAny(delay, connect).Unwrap();
			await Assert.ThrowsAnyAsync<OperationCanceledException>(() => either);
		}
	}
}
```

ericwj:
The `CancellationToken` passed to the `RequestState` constructor is never used. 

[...]

PR #1141, opened 2017-07-13
WebSocketClient.ConnectAsync now throws when cancelled

ericwj:
#1038

### DIFF
--- a/samples/SampleStartups/FakeServer.cs
+++ b/samples/SampleStartups/FakeServer.cs
@@ -22,7 +22,7 @@ namespace SampleStartups
         }
     }
 
-    public static class FakeServerWebHostBuliderExtensions
+    public static class FakeServerWebHostBuilderExtensions
     {
         public static IWebHostBuilder UseFakeServer(this IWebHostBuilder builder)
         {
--- a/src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
+++ b/src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
@@ -87,7 +87,6 @@ namespace Microsoft.AspNetCore.TestHost
         {
             private readonly IHttpApplication<Context> _application;
             private TaskCompletionSource<WebSocket> _clientWebSocketTcs;
-            private CancellationToken _cancellationToken;
             private CancellationTokenRegistration _cancellationTokenRegistration;
             private WebSocket _serverWebSocket;
 
@@ -97,8 +96,8 @@ namespace Microsoft.AspNetCore.TestHost
             public RequestState(Uri uri, PathString pathBase, CancellationToken cancellationToken, IHttpApplication<Context> application)
             {
                 _clientWebSocketTcs = new TaskCompletionSource<WebSocket>();
-                _cancellationToken = cancellationToken;
-                _cancellationTokenRegistration = cancellationToken.Register(RequestCancelled);
+                _cancellationTokenRegistration = cancellationToken.Register(
+                    () => _clientWebSocketTcs.TrySetCanceled(cancellationToken));
                 _application = application;
 
                 // HttpContext
@@ -167,11 +166,6 @@ namespace Microsoft.AspNetCore.TestHost
                 _application.DisposeContext(Context, exception);
             }
 
-            private void RequestCancelled()
-            {
-                _clientWebSocketTcs.TrySetCanceled(_cancellationToken);
-            }
-
             private string CreateRequestKey()
             {
                 byte[] data = new byte[16];
@@ -193,16 +187,16 @@ namespace Microsoft.AspNetCore.TestHost
                 var websockets = TestWebSocket.CreatePair(context.SubProtocol);
                 if (_clientWebSocketTcs.TrySetResult(websockets.Item1))
                 {
-                    Context.HttpContext.Response.StatusCode = 101; // Switching Protocols
+                    Context.HttpContext.Response.StatusCode = StatusCodes.Status101SwitchingProtocols;
                     _serverWebSocket = websockets.Item2;
-                    return Task.FromResult<WebSocket>(_serverWebSocket);
+                    return Task.FromResult(_serverWebSocket);
                 }
                 else
                 {
-                    Context.HttpContext.Response.StatusCode = 500; // Internal Server Error
+                    Context.HttpContext.Response.StatusCode = StatusCodes.Status500InternalServerError;
                     websockets.Item1.Dispose();
                     websockets.Item2.Dispose();
-                    return _clientWebSocketTcs.Task; // Cancelled or Faulted - no result
+                    return _clientWebSocketTcs.Task; // Canceled or Faulted - no result
                 }
             }
         }
--- a/src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
+++ b/src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
@@ -87,6 +87,8 @@ namespace Microsoft.AspNetCore.TestHost
         {
             private readonly IHttpApplication<Context> _application;
             private TaskCompletionSource<WebSocket> _clientWebSocketTcs;
+            private CancellationToken _cancellationToken;
+            private CancellationTokenRegistration _cancellationTokenRegistration;
             private WebSocket _serverWebSocket;
 
             public Context Context { get; private set; }
@@ -95,6 +97,8 @@ namespace Microsoft.AspNetCore.TestHost
             public RequestState(Uri uri, PathString pathBase, CancellationToken cancellationToken, IHttpApplication<Context> application)
             {
                 _clientWebSocketTcs = new TaskCompletionSource<WebSocket>();
+                _cancellationToken = cancellationToken;
+                _cancellationTokenRegistration = cancellationToken.Register(RequestCancelled);
                 _application = application;
 
                 // HttpContext
@@ -163,6 +167,11 @@ namespace Microsoft.AspNetCore.TestHost
                 _application.DisposeContext(Context, exception);
             }
 
+            private void RequestCancelled()
+            {
+                _clientWebSocketTcs.TrySetCanceled(_cancellationToken);
+            }
+
             private string CreateRequestKey()
             {
                 byte[] data = new byte[16];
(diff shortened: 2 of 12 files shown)


davidfowl:
I haven't reviewed the code yet but the formatting here is totally jacked. Are you using tabs? If so, change the formatting to spaces to make it match the rest of the file. See the https://github.com/aspnet/Home/wiki/Engineering-guidelines guide for more details on style.

ericwj:
That should do it.

davidfowl:
@ericwj The brace style is also wrong.

ericwj:
Dreadful. Hopefully there is a better way than how I did it.. Probably.

muratg:
@JunTaoLuo Could you take a look please?

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -181,12 +190,20 @@ bool IHttpWebSocketFeature.IsWebSocketRequest
...
-                return Task.FromResult<WebSocket>(_serverWebSocket);
+                if (_clientWebSocketTcs.TrySetResult(websockets.Item1))
+                {
+                    Context.HttpContext.Response.StatusCode = 101; // Switching Protocols
+                    _serverWebSocket = websockets.Item2;
+                    return Task.FromResult<WebSocket>(_serverWebSocket);
+                }
+                else
+                {
+                    Context.HttpContext.Response.StatusCode = 500; // Internal Server Error
```
jkotalik: @JunTaoLuo is 500 correct here?

test/Microsoft.AspNetCore.TestHost.Tests/TestClientTests.cs
```diff
@@ -239,6 +239,54 @@ public void Dispose()
...
+                });
+            var server = new TestServer(builder);
+
+            // Act
+            var client = server.CreateWebSocketClient();
+            var tokenSource = new CancellationTokenSource();
+            tokenSource.Cancel();
+
+            // Assert
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(async () => await client.ConnectAsync(new System.Uri("http://localhost"), tokenSource.Token));
```
jkotalik: Can you also verify the response here has the correct status code?

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -181,12 +190,20 @@ bool IHttpWebSocketFeature.IsWebSocketRequest
...
             {
-                Context.HttpContext.Response.StatusCode = 101; // Switching Protocols
-
                 var websockets = TestWebSocket.CreatePair(context.SubProtocol);
-                _clientWebSocketTcs.SetResult(websockets.Item1);
-                _serverWebSocket = websockets.Item2;
-                return Task.FromResult<WebSocket>(_serverWebSocket);
+                if (_clientWebSocketTcs.TrySetResult(websockets.Item1))
+                {
+                    Context.HttpContext.Response.StatusCode = 101; // Switching Protocols
```
JunTaoLuo: Use `StatusCode.Status101SwitchingProtocols` instead.

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
JunTaoLuo: Any non-101 response is considered a failed handshake. Given that we do not know the reason that the cancellation token is canceled, I think 500 is appropriate here.

test/Microsoft.AspNetCore.TestHost.Tests/TestClientTests.cs
```diff
@@ -239,6 +239,54 @@ public void Dispose()
             clientSocket.Dispose();
         }
 
+        [ConditionalFact]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono, SkipReason = "Hangs randomly (issue #507)")]
```
JunTaoLuo: We no longer plan to test on Mono, we can remove all instance of this skip condition.

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -163,6 +167,11 @@ internal void ServerCleanup(Exception exception)
                 _application.DisposeContext(Context, exception);
             }
 
+            private void RequestCancelled()
```
JunTaoLuo: The method doesn't really need to be extracted out here. Inline this with the `cancellationToken.Register(...)` call. That way you also do not need the `_cancellationToken` field.

test/Microsoft.AspNetCore.TestHost.Tests/TestClientTests.cs
```diff
@@ -239,6 +239,54 @@ public void Dispose()
...
+        public async Task WebSocketAcceptThrowsWhenCancelled()
+        {
+            // Arrange
+            // This logger will attempt to access information from HttpRequest once the HttpContext is createds
```
JunTaoLuo: nit: createds => created

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -181,12 +190,20 @@ bool IHttpWebSocketFeature.IsWebSocketRequest
...
+                    Context.HttpContext.Response.StatusCode = 101; // Switching Protocols
+                    _serverWebSocket = websockets.Item2;
+                    return Task.FromResult<WebSocket>(_serverWebSocket);
+                }
+                else
+                {
+                    Context.HttpContext.Response.StatusCode = 500; // Internal Server Error
+                    websockets.Item1.Dispose();
+                    websockets.Item2.Dispose();
+                    return _clientWebSocketTcs.Task; // Cancelled or Faulted - no result
```
JunTaoLuo: nit: we prefer using the American English spelling of words so Cancelled => Canceled.

test/Microsoft.AspNetCore.TestHost.Tests/TestClientTests.cs
jkotalik: I think there are many in that file, we should make a separate issue to remove mono test checks.

JunTaoLuo: Sure we can do a clean up, but there's no point in adding more code we are going to remove later.

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
ericwj: I'll do the same for the 500.

Would you also like me to fix the `IDE0001 Name can be simplified` information message?

`returnTask.FromResult(_serverWebSocket)` instead of  
`return Task.FromResult<WebSocket>(_serverWebSocket)`  
as the result type is an exact match with the declared task type.

JunTaoLuo: Sure that would be great

samples/SampleStartups/FakeServer.cs
```diff
@@ -22,7 +22,7 @@ public void Dispose()
         }
     }
 
-    public static class FakeServerWebHostBuliderExtensions
+    public static class FakeServerWebHostBuilderExtensions
```
jkotalik: Not sure why there is a diff here, but please remove changes here.

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -181,12 +184,20 @@ bool IHttpWebSocketFeature.IsWebSocketRequest
...
-
                 var websockets = TestWebSocket.CreatePair(context.SubProtocol);
-                _clientWebSocketTcs.SetResult(websockets.Item1);
-                _serverWebSocket = websockets.Item2;
-                return Task.FromResult<WebSocket>(_serverWebSocket);
+                if (_clientWebSocketTcs.TrySetResult(websockets.Item1))
+                {
+                    Context.HttpContext.Response.StatusCode = StatusCodes.Status101SwitchingProtocols;
+                    _serverWebSocket = websockets.Item2;
+                    return Task.FromResult(_serverWebSocket);
```
jkotalik: Can the Task.FromResult be of type WebSocket?

samples/SampleStartups/FakeServer.cs
JunTaoLuo: no keep it, there was an old typo. Bulider => Builder

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
JunTaoLuo: Yea, it's a generic `FromResult<T>(T result)`

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -87,6 +87,7 @@ private class RequestState : IDisposable, IHttpWebSocketFeature
         {
             private readonly IHttpApplication<Context> _application;
             private TaskCompletionSource<WebSocket> _clientWebSocketTcs;
+            private CancellationTokenRegistration _cancellationTokenRegistration;
```
JunTaoLuo: Not sure what _cancellationTokenRegistration is used for. I don't see anywhere in this PR where the callback is deregistered. I think it's safe to remove?

ericwj: It isn't used but it is referenced preventing GC. I wasn't sure whether I would call Dispose on it, but that would possibly happen before all references to `_clientWebSocketTcs.Task` and the `CancellationTokenSource` the client might have are gone and hence break this fix. Notice `_clientWebSocketTcs` isn't being disposed neither.

ericwj: Well 'break this fix'... the difference is it might stay around until the next GC instead of until whenever RequestState gets disposed. Certainly whenever it gets disposed cancellation will break. I'm just not sure exactly how GC would work; if the client has a reference to a task linked to `_clientWebSocketTcs` would that mean RequestState would stay around? That'd mean it'd be OK to properly clean up both the registration and the tcs in RequestState.Dispose. Not that it matters much because no one will have an explicit reference to RequestState...it'll be disposed at GC time anyway.

JunTaoLuo:
Merged in https://github.com/aspnet/Hosting/commit/9b1cbedffcc48154c7c52655efe67927ae933e32. Thanks for the contribution @ericwj !

Issue #877, opened 2016-11-02
Remove case-sensitivity from Configure{IHostingEnvironment.EnvironmentName}Services Lookup


carusology:
**Problem**

There is inconsistent behavior around case-sensitivity with respect to `IHostingEnvironment.EnvironmentName`. If I set `ASPNETCORE_ENVIRONMENT` to `development`, `Development`, or `DEVELOPMENT`, `IHostingEnvironment.IsDevelopment()` will return `true` because it uses `StringComparison.OrdinalIgnoreCase` [here](https://github.com/aspnet/Hosting/blob/b6da89f54cff11474f17486cdc55c2f21f2bbd6b/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingEnvironmentExtensions.cs#L77). This means in your `ConfigureServices` code in Startup.cs, as suggested in #252, you can do the following:

```
public class Startup {

    private IHostingEnvironment environment { get; }

    public Startup(IHostingEnvironment environment) {
        this.environment = environment;
    }

    public void ConfigureServices(IServiceCollection services) {
        if (this.environment.IsDevelopment()) {
            // development-specific services
        } else {
            // non-development-specific services
        }
    }

}
```

However, an alternative approach can be done by putting the environment name in the method name in the format of `Configure{IHostingEnvironment.EnvironmentName}Services`. Here's an example:

```
public class Startup {

    private IHostingEnvironment environment { get; }

    public Startup(IHostingEnvironment environment) {
        this.environment = environment;
    }

    public void ConfigureDevelopmentServices(IServiceCollection services) {
        // development-specific services
    }

    public void ConfigureServices(IServiceCollection services) {
        // non-development-specific services
    }

}
```

In order to use it this way, I must set `ASPNETCORE_ENVIRONMENT` to *exactly* `Development` for `ConfigureDevelopmentServices` to be ran because [the method name comparison is case-sensitive](https://github.com/aspnet/Hosting/blob/b6da89f54cff11474f17486cdc55c2f21f2bbd6b/src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs#L143). If you set it to any other case (`development`, `DEVELOPMENT`) the `ConfigureServices` method will be ran instead.

**Suggested Fix**

[...]

PR #1166, opened 2017-08-11
Fixes case sensitivity on Startup names

jkotalik:
Fix for #877 
I now check when there is an environmentName that
1. The string starts with methodNameStart
2. The string has the environmentName right after methodNameStart
3. The string has the methodNameEnd right after environmentName 

Also we should add StringSegments?

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
@@ -94,8 +94,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 // Full scan
                 var definedTypes = assembly.DefinedTypes.ToList();
 
-                var startupType1 = definedTypes.Where(info => info.Name.Equals(startupNameWithEnv, StringComparison.Ordinal));
-                var startupType2 = definedTypes.Where(info => info.Name.Equals(startupNameWithoutEnv, StringComparison.Ordinal));
+                var startupType1 = definedTypes.Where(info => info.Name.Equals(startupNameWithEnv, StringComparison.OrdinalIgnoreCase));
+                var startupType2 = definedTypes.Where(info => info.Name.Equals(startupNameWithoutEnv, StringComparison.OrdinalIgnoreCase));
 
                 var typeInfo = startupType1.Concat(startupType2).FirstOrDefault();
                 if (typeInfo != null)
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -65,7 +69,7 @@ public static StartupMethods LoadMethods(IServiceProvider hostingServiceProvider
 
         public static Type FindStartupType(string startupAssemblyName, string environmentName)
         {
-            if (string.IsNullOrEmpty(startupAssemblyName))
+           if (string.IsNullOrEmpty(startupAssemblyName))
```
Tratcher: formatting?

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -117,51 +121,71 @@ public static Type FindStartupType(string startupAssemblyName, string environmen
...
             var methods = startupType.GetMethods(BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static);
-            var selectedMethods = methods.Where(method => method.Name.Equals(methodNameWithEnv)).ToList();
-            if (selectedMethods.Count > 1)
+
+            if (environmentName != null)
             {
-                throw new InvalidOperationException(string.Format("Having multiple overloads of method '{0}' is not supported.", methodNameWithEnv));
+                var selectedMethods = methods.Where(method =>
+                {
+                    if (!method.Name.StartsWith(methodNameStart, StringComparison.Ordinal))
```
Tratcher: Why case sensitive for any part of the method name? It's weird to just do it for one part.

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -117,51 +121,71 @@ public static Type FindStartupType(string startupAssemblyName, string environmen
...
+                }
+                methodInfo = selectedMethods.FirstOrDefault();
             }
-            if (selectedMethods.Count == 0)
+
+            if (methodInfo == null)
             {
-                selectedMethods = methods.Where(method => method.Name.Equals(methodNameWithNoEnv)).ToList();
+                var methodNameWithNoEnv = methodNameStart + methodNameEnd;
+                var selectedMethods = methods.Where(method => method.Name.Equals(methodNameWithNoEnv)).ToList();
```
Tratcher: specify compairison

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
jkotalik: I thought we didn't want to check against Configure, Services, and Container for case sensitivity. Hence the check on only the env name.

test/Microsoft.AspNetCore.Hosting.Tests/StartupManagerTests.cs
```diff
@@ -88,6 +88,29 @@ public void StartupClassAddsConfigureServicesToApplicationServices(string enviro
             Assert.Equal(environment, options.Environment);
         }
 
+        [Theory]
+        [InlineData("caseinsensitive")]
+        [InlineData("CaseInsensitive")]
+        [InlineData("CASEINSENSITIVE")]
+        [InlineData("CaSEiNSENsitiVE")]
+        public void StartupClassAddsConfigureServicesToApplicationServicesCaseInsensitive(string environment)
```
davidfowl: Does this test configure as well?

jkotalik:
I've realized now that there are two concerns here, both in class and method finding. I'll update it to handle class finding.

jkotalik:
@davidfowl Should this change have case insensitivity for only the Environment Name, or the whole method/class name? If we did the latter, we would break people that had method names with different cases. ex startupEnvName and StartupEnvName would both be found.

jkotalik:
@Tratcher too

PR #1169, opened 2017-08-14
Adds .Value to logger calls to avoid calling .ToUriComponents on a bad header.

jkotalik:
@Tratcher Fix for #1048


davidfowl:
Test?

jkotalik:
I believe it is non-trivial to test this. The problem is by accessing the setter of HttpContext, it will throw then by providing an invalid HostString. It would require creating two test HttpRequest and HttpContext objects to set the header. For such a minor change, I'm not sure if it is worth the effort.

PR #1178, opened 2017-08-19
Add fallback when resolving application name

### LINKED ISSUE #1137: Bad error when booting webhost without calling UseStartup or Configure
```
Unhandled Exception: System.ArgumentException: A valid non-empty application name must be provided.
Parameter name: applicationName
   at Microsoft.AspNetCore.Hosting.Internal.HostingEnvironmentExtensions.Initialize(IHostingEnvironment hostingEnvironment, String applicationName, String contentRootPath, WebHostOptions options)
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.BuildCommonServices(AggregateException& hostingStartupErrors)
   at ConsoleApp11.Program.Main(String[] args) in C:\Users\dfowler\documents\visual studio 2017\Projects\ConsoleApp11\ConsoleApp11\Program.cs:line 11
```

JunTaoLuo:
Resolves #1137


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -221,7 +221,7 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
             }
 
             var contentRootPath = ResolveContentRootPath(_options.ContentRootPath, AppContext.BaseDirectory);
-            var applicationName = _options.ApplicationName;
+            var applicationName = _options.ApplicationName ?? Assembly.GetEntryAssembly()?.GetName().Name;
```
JunTaoLuo: This does not guarantee to be not null. GetEntryAssembly() returns null when launching a net461 test project via xunit. Launching an net461 app by itself or launching a netcoreapp2.0 test project via xunit returns a valid value. The suspicion here is that launching via a different AppDomain in net461 doesn't resolve this correctly.

test/Microsoft.AspNetCore.Hosting.Tests/Fakes/Startup.cs
```diff
@@ -7,16 +7,17 @@
...
 {
-    public class Startup : StartupBase
+    public class Startup : StartupBase, IStartup
     {
         public Startup()
         {
         }
 
-        public void ConfigureServices(IServiceCollection services)
+        public IServiceProvider ConfigureServices(IServiceCollection services)
```
davidfowl: Why?

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -570,17 +570,31 @@ public void DefaultContentRootIsApplicationBasePath()
...
+        public void DefaultWebHostBuilderWithNoStartupThrows()
         {
-            var builder = new ConfigurationBuilder();
             var host = new WebHostBuilder()
                 .UseServer(new TestServer());
 
-            var ex = Assert.Throws<ArgumentException>(() => host.Build());
+            var ex = Assert.Throws<InvalidOperationException>(() => host.Build());
+
+            Assert.Equal("No service for type 'Microsoft.AspNetCore.Hosting.IStartup' has been registered.", ex.Message);
```
JunTaoLuo: This is the exception thrown at https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L169. I think we might be able to do a little better by indicating that A startup need to be specified via UseStartup, Configure, or injecting IStartup.

khellang:
Have you considered what happens to `HostingStartupAssemblies` when `ApplicationName` ends up being `null`?

https://github.com/aspnet/Hosting/blob/312e0da556e491a1a5b120a3c6d5626f5b5e2edc/src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs#L31-L32

I think it might be a good idea to fall back to the entry assembly _inside_ `WebHostOptions` instead of outside.

test/Microsoft.AspNetCore.Hosting.Tests/Fakes/Startup.cs
JunTaoLuo: So that I can use it in the test at https://github.com/aspnet/Hosting/blob/57634a8dd91852a9f62d49f078260b604ed389ec/test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs#L589

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
JunTaoLuo: I tried to create a better error message but didn't come up with anything nice. The best I was able to come up with was 
```c#
throw new InvalidOperationException($"No startup methods configured. Please specify startup methods via WebHostBuilder.UseStartup, WebHostBuilder.Configure, injecting {nameof(IStartup)} or specifying the {nameof(WebHostOptions.StartupAssembly)}.");
```
Which is not concise either.

jkotalik: I think that is fine.

Tratcher: I like the new one.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -570,17 +570,31 @@ public void DefaultContentRootIsApplicationBasePath()
...
+                })
+                .UseServer(new TestServer());
 
-            // ArgumentException adds "Parameter name" to the message and this is the cleanest way to make sure we get the right
-            // expected string
-            Assert.Equal(new ArgumentException("A valid non-empty application name must be provided.", "applicationName").Message , ex.Message);
+            using (var host = builder.Build())
+            {
+                var hostingEnv = host.Services.GetService<IHostingEnvironment>();
+                Assert.Equal(Assembly.GetEntryAssembly()?.GetName().Name, hostingEnv.ApplicationName);
```
Tratcher: It would be clearer to hard code it for the expected results from .NET vs Core.

test/Microsoft.AspNetCore.Hosting.Tests/Fakes/Startup.cs
```diff
@@ -7,16 +7,17 @@
 
 namespace Microsoft.AspNetCore.Hosting.Fakes
 {
-    public class Startup : StartupBase
+    public class Startup : StartupBase, IStartup
```
khellang: `StartupBase` already implements `IStartup`? 😕 

https://github.com/aspnet/Hosting/blob/1ea0647ae2395252e6e8423691a544ca34491a73/src/Microsoft.AspNetCore.Hosting/Startup/StartupBase.cs#L10

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -570,17 +570,31 @@ public void DefaultContentRootIsApplicationBasePath()
...
+
+            Assert.Equal("No service for type 'Microsoft.AspNetCore.Hosting.IStartup' has been registered.", ex.Message);
+        }
+
+        [Fact]
+        public void DefaultApplicationNameToStartupAssemblyName()
+        {
+            var builder = new WebHostBuilder()
+                .ConfigureServices(services =>
+                {
```
khellang: Nit: expression?

test/Microsoft.AspNetCore.Hosting.Tests/Fakes/Startup.cs
JunTaoLuo: This `StartupBase` is https://github.com/aspnet/Hosting/blob/dev/test/Microsoft.AspNetCore.Hosting.Tests/Fakes/StartupBase.cs#L8-L19

I preferred the more surgical fix to implement IStartup in this class but there may be better alternatives.

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
JunTaoLuo: I'm going to change this to a functional test so the actual assembly name is resolved instead of the xunit's testhost.

jkotalik:
@JunTaoLuo Seems to be some odd diffs here (renaming). I'm not sure if it is intentional.

JunTaoLuo:
I moved two projects that were test apps into a different test assets directory. They are not tests themselves but are used in other tests.

JunTaoLuo:
Fixed test failures, ready for a final 👀

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -166,7 +166,12 @@ private void EnsureStartup()
                 return;
             }
 
-            _startup = _hostingServiceProvider.GetRequiredService<IStartup>();
+            _startup = _hostingServiceProvider.GetService<IStartup>();
+
+            if (_startup == null)
+            {
+                throw new InvalidOperationException($"No startup configured. Please specify startup via WebHostBuilder.UseStartup, WebHostBuilder.Configure, injecting {nameof(IStartup)} or specifying the startup assembly via {nameof(WebHostDefaults.StartupAssemblyKey)} in the web host configuration.");
```
davidfowl: 👍

test/TestAssets/IStartupInjectionAssemblyName/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace IStartupInjectionAssemblyName
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var server = new TestServer(new WebHostBuilder()
+                .ConfigureServices(services => services.AddSingleton<IStartup, Startup>()));
+
+            // Mimic application startup messages so application deployer knows that the application has started
```
davidfowl: That's super hacky 😄

PR #1180, opened 2017-08-20
Decouple ApplicationName from any assembly names

### LINKED ISSUE #1137: Bad error when booting webhost without calling UseStartup or Configure

### LINKED ISSUE #1179: Using a custom "ApplicationName" and UseStartup<Startup>() doesn't work
I was trying to set my own `IHostingEnvironment.ApplicationName` on my WebHostBuilder. However, if I set it before I use `UseStartup`, [it will be overwritten](https://github.com/aspnet/Hosting/blob/4cdc970055ef2ad77887109c4acde89278e0ce18/src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs#L51) and the output is `Hello from 'AppName'!`:

```csharp
namespace AppName
{
    public class Program
    {
        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            app.Run(async (context) =>
            {
                await context.Response.WriteAsync($"Hello from '{env.ApplicationName}'!");
            });
        }

        {
            BuildWebHost(args).Run();
        }

        public static IWebHost BuildWebHost(string[] args) =>
            WebHost.CreateDefaultBuilder(args)
                .UseSetting(WebHostDefaults.ApplicationKey, "MyCustomAppName")
                .UseStartup<Program>()
                .Build();
    }
}
```
So I thought I'm smart and just call `UseSetting` after `UseStartup` but this results in the following exception:
```
Hosting startup assembly exception
System.InvalidOperationException: Startup assembly MyCustomAppName failed to execute. See the inner exception for more details. ---> System.IO.FileNotFoundException: Could not load file or assembly 'MyCustomAppName, Culture=neutral, PublicKeyToken=null'. Das System kann die angegebene Datei nicht finden.
   at System.Reflection.RuntimeAssembly._nLoad(AssemblyName fileName, String codeBase, Evidence assemblySecurity, RuntimeAssembly locationHint, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, Boolean suppressSecurityChecks, IntPtr ptrLoadContextBinder)
   at System.Reflection.RuntimeAssembly.InternalLoadAssemblyName(AssemblyName assemblyRef, Evidence assemblySecurity, RuntimeAssembly reqAssembly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean throwOnFileNotFound, Boolean forIntrospection, IntPtr ptrLoadContextBinder)
   at System.Reflection.Assembly.Load(AssemblyName assemblyRef)
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.BuildCommonServices(AggregateException& hostingStartupErrors)
```

Is this a feature (meaning, custom application names can not be used with Startup-types) or a bug?

khellang:
Here's a proposal to fix a couple of issues...

This PR effectively stops using `StartupAssembly` to check whether to scan for an `IStartup` type. Instead it always sets `StartupAssembly` when calling `Configure`, `UseStartup` or `UseStartupAssembly` (new), and in addition, it sets a new `FindStartupType` key when you need to scan for a startup type.

If `StartupAssembly` hasn't been set (no method above has been called), it falls back to the entry assembly name. This fixes #1137.

Instead of using `ApplicationName` in `HostingStartupAssemblies`, it now uses `StartupAssembly`.

This means that we no longer use `ApplicationName` for any assembly loading or scanning, it's simply just a string that you can set to whatever you want. This fixes #1179.

If an `ApplicationName` hasn't been provided (by the user), it falls back to `StartupAssembly` (which could've already fallen back to the entry assembly name).

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
@@ -58,7 +58,7 @@ namespace Microsoft.AspNetCore.Hosting
 
 
             return hostBuilder
-                .UseSetting(WebHostDefaults.ApplicationKey, startupAssemblyName)
+                .UseSetting(WebHostDefaults.FindStartupTypeKey, "true")
                 .UseSetting(WebHostDefaults.StartupAssemblyKey, startupAssemblyName);
         }
 
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
@@ -90,6 +90,27 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         /// <summary>
+        /// Specify the application name. This can be retrieved from <see cref="IHostingEnvironment.ApplicationName"/>.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="applicationName">The name of the application.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder UseApplicationName(this IWebHostBuilder hostBuilder, string applicationName)
+        {
+            if (applicationName == null)
+            {
+                throw new ArgumentNullException(nameof(applicationName));
+            }
+
+            if (applicationName.Length == 0)
+            {
+                throw new ArgumentException("Application name must not be empty.", nameof(applicationName));
+            }
+
+            return hostBuilder.UseSetting(WebHostDefaults.ApplicationKey, applicationName);
+        }
+
+        /// <summary>
         /// Specify the server to be used by the web host.
         /// </summary>
         /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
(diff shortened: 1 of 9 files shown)


khellang:
> I don't see the point of this change if consumers of `IHostingEnvironment` still assume ApplicationName is an assembly name. Save this for 3.0 when we can change both things together.

Haven't we already established that you can't confidently assume that `ApplicationName` is an assembly name? We've already seen several reports that people are using `ApplicationName` for other things? And you can't really blame them given the unfortunate naming. Especially when there's already an (internal) configuration value called `StartupAssembly`.

davidfowl:
> Haven't we already established that you can't confidently assume that ApplicationName is an assembly name? 

No, if it's set, it's an assembly name. Anyone that changes it to something else will soon learn that this assumption is squarely baked in at multiple levels.

> We've already seen several reports that people are using ApplicationName for other things. And you can't really blame them given the unfortunate naming and documentation. Since the IHostingEnvironment property is settable, there's no guarantee that it'll be an assembly name at all. IMO, any assumption that it's an assembly name is a bug.

The problem is this fixes nothing. Several components use ApplicationName as an assembly name like MVC. This change would have been great to do before 2.0 but now I don't see what you've done besides given people false hope that they can now use ApplicationName when infact it will still break.

davidfowl:
I'd prefer if we introduced a new property in the next major version that was something like `FriendlyName` (like appdomain has)

khellang:
Alright. You've convinced me. I guess it's too late to do anything about it 😢

davidfowl:
v3!

Issue #1108, opened 2017-06-19
If the same Startup assembly is specified multiple times it is being executed multiple times


moozzyk:
If the same Startup assembly is specified more than once I think it would be better to execute it once and print a warning.

muratg:
We should consider making loading idempotent.

khellang:
The same would be true if you have multiple `HostingStartupAttribute` defined (with the same type). Is there a reason why multiple attributes in the same assembly is allowed?

halter73:
I tested using multiple HostingStartupAttributes defined with the same time. `Assembly.GetCustomAttributes<HostingStartupAttribute>()` already dedupes identical attributes with identical arguments, so that isn't an issue.

Specifying the same assembly multiple times in ASPNETCORE_HOSTINGSTARTUPASSEMBLIES do cause that assembly's hosting startup classes to be executed multiple times though.

khellang:
> `Assembly.GetCustomAttributes<HostingStartupAttribute>()` already dedupes identical attributes with identical arguments, so that isn't an issue.

Ah, interesting. Didn't know about that behavior. Is that related to `TypeId`? (I've always wondered what that property is for...)

jkotalik:
So from what I can tell, there are two scenarios we need to check here:
- in UseStartup<Type>()
- in ASPNETCORE_HOSTINGSTARTUPASSEMBLIES 
@moozzyk were there any other places where you saw duplicate startup assemblies being specified?

moozzyk:
@jkotalik - yes, I think this is it. 

(Not sure how you are going to identify duplicates but on Windows paths/file names are not case sensitive.)

jkotalik:
Right now I think we have to check in [WebHostBuilder.UseSetting](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L83-L87), as UseStartup is an extension method on WebHostBuilder. It would be odd to expose a set containing the assemblies being used.

Tratcher:
Deduplicate at the point of execution, not where the properties are set.

Tratcher:
Here: https://github.com/aspnet/Hosting/blob/a2962d54f18dece7ec57ee42f3e565f3727168e1/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L198

jkotalik:
[...]

PR #1183, opened 2017-08-24
Detect duplicate startups in HostingStartupAssemblies

jkotalik:
Fix for #1108

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.Hosting
         private bool _webHostBuilt;
         private List<Action<WebHostBuilderContext, IConfigurationBuilder>> _configureAppConfigurationBuilderDelegates;
 
-        private ISet<string> assemblies;
+        private IDictionary<string, string> assemblies;
         /// <summary>
         /// Initializes a new instance of the <see cref="WebHostBuilder"/> class.
         /// </summary>
@@ -41,7 +41,7 @@ namespace Microsoft.AspNetCore.Hosting
             _hostingEnvironment = new HostingEnvironment();
             _configureServicesDelegates = new List<Action<WebHostBuilderContext, IServiceCollection>>();
             _configureAppConfigurationBuilderDelegates = new List<Action<WebHostBuilderContext, IConfigurationBuilder>>();
-            assemblies = new HashSet<string>();
+            assemblies = new Dictionary<string, string>();
 
             _config = new ConfigurationBuilder()
                 .AddEnvironmentVariables(prefix: "ASPNETCORE_")
@@ -201,12 +201,6 @@ namespace Microsoft.AspNetCore.Hosting
                 {
                     try
                     {
-                        if (assemblies.Contains(assemblyName))
-                        {
-                            // TODO print a warning
-                            continue;
-                        }
-                        assemblies.Add(assemblyName);
                         var assembly = Assembly.Load(new AssemblyName(assemblyName));
 
                         foreach (var attribute in assembly.GetCustomAttributes<HostingStartupAttribute>())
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -193,12 +194,18 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
...
             {
                 var exceptions = new List<Exception>();
+                var assemblies = new HashSet<StringSegment>(StringSegmentComparer.OrdinalIgnoreCase);
 
                 // Execute the hosting startup assemblies
                 foreach (var assemblyName in _options.HostingStartupAssemblies)
                 {
                     try
                     {
+                        if (assemblies.Contains(assemblyName))
```
campersau: `if (!assemblies.Add(assemblyName))`

Also you are inside of a try catch. Maybe just add the exception to `exceptions`?

khellang: Why even throw at all? Does it matter to anyone that it was skipped and not added twice?

davidfowl: Agreed. This shouldn't throw.

Tratcher: Yes, just ignore the duplicate.

Tratcher: `_options.HostingStartupAssemblies.Distinct(StringSegmentComparer.OrdinalIgnoreCase)`?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -195,7 +197,7 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
                 var exceptions = new List<Exception>();
 
                 // Execute the hosting startup assemblies
-                foreach (var assemblyName in _options.HostingStartupAssemblies)
+                foreach (var assemblyName in _options.HostingStartupAssemblies.Distinct(StringComparer.OrdinalIgnoreCase))
```
JunTaoLuo: I think it's worth logging a warning. We shouldn't silently ignore when a startup assembly is listed twice since that is likely wrong.

jkotalik: I'd assume we probably just Console.WriteLine the error?

JunTaoLuo: Yea the logger isn't configured at this point so just write to console.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -15,6 +16,7 @@
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.ObjectPool;
+using Microsoft.Extensions.Primitives;
```
JunTaoLuo: nit: this isn't needed anymore.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: Let's not do that. You can't turn it off. Let's not do more things you can't turn off.

jkotalik: So are we just going to silently ignore duplicate assemblies?

davidfowl: Yes

JunTaoLuo: The alternative is saving this information somewhere and pass it to WebHost and log a warning there. We have a similar pattern with hosting startup assembly errors. Is it worth it?

jkotalik: Hence why I added it as an exception instead of a warning before. Not sure if it is worth it here just for this message.

jkotalik:
@Tratcher @davidfowl @JunTaoLuo What's our consensus here? Silently ignore?

Tratcher:
yes

jkotalik:
@Tratcher _ping_

Issue #1160, opened 2017-08-07
Port logging startup exceptions in 1.1.x 


jkotalik:
See #1153 
This should be ported for future 1.1.x patches.

muratg:
@Eilon, 1.1.3 candidate.

Eilon:
This patch is *approved* for 1.1.x. Please follow the patch creation instructions that were sent by email, and submit a PR to the correct branch.

Eilon:
@muratg - please assign.

jkotalik:
Merged.

Eilon:
@jkotalik - mark as done?

jkotalik:
@Eilon Sorry I'm still an intern. pls no hate

PR #1184, opened 2017-08-24
Porting "Always log startup exceptions to 1.1.4"

jkotalik:
Fix for https://github.com/aspnet/Hosting/issues/1160
Had to modify tests to work with 1.1.x

### DIFF
--- a/test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
+++ b/test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
@@ -16,6 +17,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Extensions.ObjectPool;
 using Microsoft.Extensions.PlatformAbstractions;
 using Xunit;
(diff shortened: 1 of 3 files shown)


muratg:
Why is the branch 1.1.4?  Shouldn't it be 1.1.3?  

cc @pranavkm

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -573,6 +575,47 @@ public void Build_DoesNotOverrideILoggerFactorySetByConfigureServices()
...
+        [Fact]
+        public void StartupErrorsAreLoggedIfCaptureStartupErrorsIsFalse()
+        {
+            var testSink = new TestSink();
+            var factory = new TestLoggerFactory(testSink, enabled: true);
+
+            var builder = CreateWebHostBuilder()
+                .CaptureStartupErrors(false)
+                .ConfigureServices(collection => collection.AddSingleton<ILoggerFactory>(factory))
+
```
pranavkm: Remove empty line

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -573,6 +575,47 @@ public void Build_DoesNotOverrideILoggerFactorySetByConfigureServices()
...
+                .Configure(app =>
+                {
+                    throw new InvalidOperationException("Startup exception");
+                })
+                .UseServer(new TestServer());
+
+            Assert.Throws<InvalidOperationException>(() => builder.Build());
+
+            Assert.True(testSink.Writes.Any(w => w.Exception?.Message == "Startup exception"));
```
pranavkm: `Assert.Contains("Startup exception", testSink.Writes.Select(w => w.Exception?.Message).Where(m => m != null));` is slightly better because it'll cause xunit to print the 2nd argument (the haystack). Helps when the test fails. With the current approach all you get is a `Assert.True(false)` which is less useful.

pranavkm:
@muratg yup, the branch needs to `rel/1.1.3`. We use `patch/*` branches to make infrastructural \ test \ non-src changes to produces (for .e.g flaky tests), but release branches always follow the package version.

@jkotalik you'll have to 

a) Create a `rel/1.1.3` branch (`git checkout patch/1.1.4 -b rel/1.1.3`
b) Bump up the version of Hosting to `1.1.3` and update all dependencies (see @JunTaoLuo's email)

JunTaoLuo:
FYI, don't bump the version of Hosting to 1.1.3 in this PR, use Coherence-Patch instead according to the instruction in the email instead.

jkotalik:
Closing to merge with rel/1.1.3

muratg:
Thanks folks! We should ensure the other patch check-ins follow these guidelines as well.

PR #1186, opened 2017-08-24
Ports "Always log startup exceptions" to 1.1.3

jkotalik:
Fix for #1184


pranavkm:
Changes look good. Lots of version changes incoming. :shipit:

PR #1189, opened 2017-08-28
Cross-compile Windows Services hosting package

### LINKED ISSUE #904: Microsoft.AspNetCore.Hosting.WindowsServices  .NET Core Support
Is there any plan to upgrade the above assembly to support .NET Core ? 

When trying to build a .NET Core Windows service I'm experiencing problems when inheriting from WebHostService which in turn Inherits from ServiceBase and receive the following error: 

The type 'ServiceBase' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.ServiceProcess, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.	ConsoleApp1..NETCoreApp,Version=v1.1

henkmollema:
Cross-compile `Microsoft.AspNetCore.Hosting.WindowsServices` for .NET 4.6.1 and .NET Core 2.0.

I had the use the .NET Core MyGet feed as the `ServiceBase` API is only available in version `4.5.0-*`. Perhaps   the package in the ASP.NET Core MyGet feed can be updated as well?

Resolves #904 

/cc @muratg @JunTaoLuo

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -40,6 +40,7 @@
     <SerilogSinksFilePackageVersion>3.2.0</SerilogSinksFilePackageVersion>
     <SystemDiagnosticsDiagnosticSourcePackageVersion>4.4.1</SystemDiagnosticsDiagnosticSourcePackageVersion>
     <SystemReflectionMetadataPackageVersion>1.5.0</SystemReflectionMetadataPackageVersion>
+    <SystemServiceProcessVersion>4.5.0-preview1-25914-04</SystemServiceProcessVersion>
     <XunitPackageVersion>2.3.0</XunitPackageVersion>
     <XunitRunnerVisualStudioPackageVersion>2.3.0</XunitRunnerVisualStudioPackageVersion>
   </PropertyGroup>
--- a/src/Microsoft.AspNetCore.Hosting.WindowsServices/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
+++ b/src/Microsoft.AspNetCore.Hosting.WindowsServices/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core hosting infrastructure and startup logic for web applications running within a Windows service.</Description>
-    <TargetFramework>net461</TargetFramework>
+    <TargetFrameworks>net461;netstandard2.0</TargetFrameworks>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;hosting</PackageTags>
@@ -10,7 +10,12 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.AspNetCore.Hosting\Microsoft.AspNetCore.Hosting.csproj" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' ">
     <Reference Include="System.ServiceProcess" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.ServiceProcess.ServiceController" Version="$(SystemServiceProcessVersion)" />
   </ItemGroup>
 
 </Project>


build/dependencies.props
```diff
@@ -13,5 +13,6 @@
     <SerilogExtensionsLoggingVersion>1.4.0</SerilogExtensionsLoggingVersion>
     <SerilogFileSinkVersion>3.2.0</SerilogFileSinkVersion>
     <SystemReflectionMetadataVersion>1.5.0-*</SystemReflectionMetadataVersion>
+    <SystemServiceProcessVersion>4.5.0-*</SystemServiceProcessVersion>
```
henkmollema: Should we specify the package name completely? E.g. `SystemServiceProcessServiceControllerVersion`. If this setting is desired/necessary at all.

NuGet.config
```diff
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
+    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
```
Tratcher: We don't add these feeds directly, we mirror the packages to our feed. @ryanbrandenburg

build/dependencies.props
natemcmaster: > Should we specify the package name completely?

Either way is fine. I'll be deleting this file soon anyways as we move to a system called "lineups" to manage our package versions.

Any reason you need 4.5.0-* specifically? Or will 4.4.0 work?

cref https://github.com/aspnet/Universe/blob/dev/build/dependencies.props#L114

NuGet.config
ryanbrandenburg: Yeah, `System.ServiceProcess.ServiceController` is already being mirrored to `aspnetcore-ci-dev`.

natemcmaster:
Talked with @Tratcher. It sounds like .NET Core 2.0 doesn't have an implementation of this API yet. We haven't started building for .NET Core 2.1 yet, so this may need to wait.

build/dependencies.props
henkmollema: Yeah, I saw a PR of you coming by where you implemented this. Figured it would be fine.

I've used the 4.5.0 packages because the API was added in .NET Core 2.1: https://github.com/dotnet/corefx/pull/22920. I thought this maps to 4.5.0-* packages of the runtime.

ryanbrandenburg: Seconded, it's easier if this can be 4.4.0, then we'll update to 4.5.0 at a later date in all repos at once.

henkmollema:
I see. Feel free to ping me if we target .NET Core 2.1 and you want me to update this PR. 😄

build/dependencies.props
natemcmaster: The 4.4.0 package only has a few types from System.ServiceProcess.* namespace. 4.5.0 adds more. Which version we need will depends which API is actually referenced from hosting. I haven't looked closely at what we use. If we need 4.5.0 that's no problem, we just need to setup our internal build processes to start using new packages from the corefx team.

natemcmaster:
Ok. I'm going to assign this PR to @JunTaoLuo for now so we can follow up on it later. We'll hold off merging till we can build against .NET Core 2.1.

build/dependencies.props
henkmollema: This package [uses `ServiceBase`](https://github.com/aspnet/Hosting/blob/9b1cbedffcc48154c7c52655efe67927ae933e32/src/Microsoft.AspNetCore.Hosting.WindowsServices/WebHostService.cs#L12), which was added in .NET Core 2.1.

natemcmaster: Good to know. Then yes, we'll need the 4.5.0 version of the package.

JunTaoLuo:
We shouldn't be cross-compiling here. We should wait for 2.1 where the ServiceBase API is available. This was why I was asking about when we are taking the new runtime last week.

henkmollema:
Will `ServiceBase` be part of .NET Standard 2.1?

natemcmaster:
> .NET Standard 2.1

There are no plans for this AFAIK, however the API surface should be available when you target `netcoreapp2.1`

henkmollema:
I see, does that mean we drop .NET 4.6.1?

JunTaoLuo:
Ah I mis-spoke, I meant we will need use netcoreapp2.1, not netcoreapp2.0. We plan to have it available for both core and framework.

henkmollema:
That makes sense, figured we need that. I was already looking for `netcoreapp2.1` but could not find it in the ASP.NET repos 😄

Tratcher:
We should also update the ServiceBase sample for the new Generic Host.
https://github.com/aspnet/Hosting/blob/ae9da9290eb825f4145aaaa23603186d51211e53/samples/GenericHostSample/ServiceBaseLifetime.cs#L1

And add a sample for the current RunAsService.

Tratcher:
Waiting on https://github.com/dotnet/designs/blob/d8c7b4a7b0a88fe83db1e7778df43574103f0a06/compat-pack/compat-pack.md

JunTaoLuo:
FYI The current eta is for middle of November.

Tratcher:
https://www.nuget.org/packages/System.ServiceProcess.ServiceController/4.5.0-preview1-25914-04

henkmollema:
@Tratcher I've updated with the new preview package of `
System.ServiceProcess.ServiceController`. I've cross-compiled with .NET Standard 2.0 instead of .NET Core. Does that makes sense? Or should we target .NET Core instead.

Tratcher:
netstandard2.0 is preferred wherever possible.

JunTaoLuo:
Merged in https://github.com/aspnet/Hosting/commit/e03c6a7908b338235d6619d4959c4760114a642b. Thanks @henkmollema for the contribution!

henkmollema:
:tada:

Issue #1135, opened 2017-07-07
[TestServer] HttpContext.Connection containing wrong/default values


peterwurzinger:
Hi!
I don't know if it's actually an issue, so I'll just describe the behavior I've discovered.

I'm using Microsoft.AspNetCore.Hosting.TestHost.TestServer for integration testing purposes for a custom authentication middleware/handler.
The configuration is as follows
```cs
_server = new TestServer(new WebHostBuilder()
                .UseStartup<Startup>());
//Startup only contains configuration stuff for my middleware

_client = _server.CreateClient();
_client.GetAsync("/");
//...
```

When executing this, the request hits my middleware and I'm accessing `HttpContext.Connection`, where all exposed properties are set to their default values (Local/RemotePort = 0, other properties are set to `null`).
Am I missing a configuration step or something? We cannot get around testing whether the middleware is checking the clients IP-Address or not.

I tried to debug that based on the sources of Microsoft.AspNetCore.Http(Abstractions), but I didn't get too far with that.

EDIT: I'm using the preview2-final - packages of Microsoft.AspNetCore.*, if that makes any difference.

davidfowl:
What are you expecting those to be set to?

peterwurzinger:
Well my expectation was, to give a real example, that the property RemoteIPAddress would contain an instance of IPAddress containing the client's IP - e.g in an integration test scenario 127.0.0.1 or ::1.
Maybe my expectation is wrong, but since ALL properties seem to be unset, the ConnectionInfo instance is unusable in every use case I can imagine atm.

Tratcher:
It's not even loopback, it's in memory.
What we really need is an easy way to provide values. Most test apps don't care.

peterwurzinger:
So the client's request doesn't take all the way through the server's request pipeline and will therefore not trigger all modules that would normally handle the request - did I get that right @Tratcher?

davidfowl:
The client is fake, the request is fake, it's an in memory pipeline, there is no networking involved. It *does* however, run the entire server pipeline so everything gets invoked like normal, just without network IO.

peterwurzinger:
Okay, and due to the fact that there is no networking involved such OSI 3/4 artifacts like IP addresses won't be available..?

davidfowl:
@peterwurzinger correct. All of that to say, we *could* set the IPAddress to localhost and it wouldn't be incorrect. It's just that we don't today so it's null.

muratg:
[...]

Issue #816, opened 2016-07-14
Consider adding a RequestHandler property in TestServer to represent RequestDelegate


jijiechen:
TestServer is great for most scenarios, but sometimes people really want to reference an HttpContext instance to confirm it is correctly processed.

Right now, we can only create a client object and invoke TestServer in a "integration" way. Could you add some property to TestServer class to add ability to invoke the RequestDelegate directly?
With such reference, we can check state of HttpContext, can assert exceptions which correctly handled, etc.
We can easily to implement another IServer to make things happen but if you make it a builtin feature, it would be very nice.

Refer to issue https://github.com/aspnet/HttpAbstractions/issues/669

Tratcher:
Yes, sometimes HttpClient is too abstracted and I'd like to directly build and send an HttpContext and check the results. This would be nice for customizing server side structures like the IFeature interfaces.

MCGPPeters:
Working on a port a the [OWINHttpMessageHandler](https://github.com/damianh/OwinHttpMessageHandler) Damian Hickey created. Maybe it can help : [AspNetCoreHttpMessageHandler](https://github.com/MCGPPeters/AspNetCoreHttpMessageHandler) ....

Tratcher:
Bumping out of backlog for re-triage. We get lots of requests for higher fidelity control over the request. People want to set things like a Windows Auth user, the client certificate, remote IP, etc..
See:
https://github.com/aspnet/Hosting/issues/1135
https://github.com/aspnet/Hosting/issues/924#issuecomment-334813951

jkotalik:
@Tratcher I think it is best to expose the feature collection for modification before we send the request. It may be hard to expose everything in the TestServer as a property.

Tratcher:
@jkotalik this wouldn't be for modification, this would be for full construction. You wouldn't use HttpClient as an intermediary with this approach.

jkotalik:
Gotcha. Just so we are on the same page, we would have a new code path that accepts an IFeatureCollection or HttpContext and allows you to effectively call "SendAsync" on the context. Then when you receive the response, we can expose the modified HttpContext and or the RequestDelegate.

fletchsod-developer:
Any idea what version will this be available in NuGet?

davidfowl:
When we ship the next version of asp.net Core. You can use the myget feed if you want to use bleeding edge stuff. We don’t ship individual packages.

PR #1196, opened 2017-09-02
Sets base address of TestServer to 127.0.0.1:80

jkotalik:
Fix for #1135

### DIFF
--- a/test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
+++ b/test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
@@ -43,6 +43,13 @@ namespace Microsoft.AspNetCore.TestHost
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
```diff
@@ -43,6 +43,13 @@ public void DoesNotCaptureStartupErrorsByDefault()
         }
 
         [Fact]
+        public void TestServerDefaultsToCorrectIpAndPort()
+        {
+            var host = new TestServer(new WebHostBuilder().Configure(app => { }));
+            Assert.Equal(new Uri("http://127.0.0.1:80"), host.BaseAddress);
```
davidfowl: We should check the HttpContext.Request.Connection.RemoteIpAddress has the right address.

jkotalik:
Closing as we may want to consolidate #1135 and #816 together.

PR #1203, opened 2017-09-11
Add extension methods in our test host package to streamline setup of apps.

javiercn:
* Create a sources package to encode the convention followed in our templates to create a WebHost.
* Add an extension method to setup the content root relative to the solution folder.
* Add a factory method to create a WebHostBuilder based on the pattern followed in our template.


shared/Microsoft.AspNetCore.Hosting.WebHostBuilderFactory.Sources/WebHostFactoryResolver.cs
```diff
@@ -0,0 +1,114 @@
...
+            var factory = programType?.GetTypeInfo().GetDeclaredMethod(CreateWebHostBuilder);
+            if (factory == null)
+            {
+                return FactoryResolutionResult.NoCreateWebHostBuilder(programType);
+            }
+
+            return FactoryResolutionResult.Succeded(args => (IWebHostBuilder)factory.Invoke(null, new object[] { args }), programType);
+        }
+
+        public static FactoryResolutionResult ResolveWebHostFactory(Assembly assembly)
```
Tratcher: No tests for this.

shared/Microsoft.AspNetCore.Hosting.WebHostBuilderFactory.Sources/WebHostFactoryResolver.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Reflection;
+
+namespace Microsoft.AspNetCore.Hosting.WebHostBuilderFactory
+{
+    internal class WebHostFactoryResolver
+    {
+        public static readonly string CreateWebHostBuilder = nameof(CreateWebHostBuilder);
+        public static readonly string BuildWebHost = nameof(BuildWebHost);
+
+        public static FactoryResolutionResult ResolveWebHostBuilderFactory(Assembly assembly)
```
Tratcher: I would expect EF wants wants a combine method they can call that supports both the 2.0 or 2.1 pattern.

javiercn: That is https://github.com/aspnet/Hosting/pull/1203/files#diff-572ec7cf75c906228766114371468f11R31

shared/Microsoft.AspNetCore.Hosting.WebHostBuilderFactory.Sources/WebHostFactoryResolver.cs
javiercn: I didn't want to create a test for this in Hosting as it would require another project. My strategy was to have the tests on the MVC package (that consumes this) and on EF that already has tests for this on the dotnet tool. I can add all the tests here if you prefer, but it would be very verbose IMO and it's already going to be tested at higher levels of the stack.

Tratcher: Add the project, that's not significant overhead. It's better for the tests to live adjacent to the code.

shared/Microsoft.AspNetCore.Hosting.WebHostBuilderFactory.Sources/WebHostFactoryResolver.cs
```diff
@@ -0,0 +1,114 @@
...
+                        return FactoryResolutionResult.Succeded(args => (IWebHost)buildWebHostMethod.Invoke(null, new object[] { args }), findResult.ProgramType);
+                    }
+                case FactoryResolutionResultKind.NoBuildWebHost:
+                default:
+                    throw new InvalidOperationException();
+            }
+        }
+    }
+
+    internal class FactoryResolutionResult
```
Tratcher: Move these to their own file.

src/Microsoft.AspNetCore.TestHost/WebHostBuilderFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.WebHostBuilderFactory;
+
+namespace Microsoft.AspNetCore.TestHost
+{
+    public static class WebHostBuilderFactory
+    {
+        public static IWebHostBuilder CreateFromStartup<TStartup>(string [] args)
```
Tratcher: This is misleading. There's no direct link between TStartup and what gets run, it's only used to look up an assembly. Why not `CreateFromProgram<TProgram>`? That's what you're actually resolving and running. Then you could even bypass the entrypoint check and go directly to the method resolver.

davidfowl: `CreateFromDefaultEntryPoint<TProgram>`

javiercn: I've been thinking about this for a bit. I think we should keep it this way (maybe find a better name like CreateFromStartupEntryPoint<TStartup> or something similar). The reason for this is that all our tools rely on the concept of Startup not the entry point (dotnet ef, etc.) and this method is meant to support the convention we follow in our templates (in which the startup class lives in the same assembly as the application).

javiercn:
Addressed feedback @Tratcher

src/Microsoft.AspNetCore.TestHost/WebHostBuilderFactory.cs
davidfowl: It's not right though. The entry point isn't necessarily where your startup code is. This is specifically about finding code in Program.cs, not the Startup class.

Tratcher: Agreed, Startup hasn't been the primary entry point for a while now.

javiercn:
Addressed feedback @Tratcher

src/Microsoft.AspNetCore.TestHost/WebHostBuilderFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+            var result = WebHostFactoryResolver.ResolveWebHostBuilderFactory(assembly);
+            if (result.ResultKind != FactoryResolutionResultKind.Success)
+            {
+                return null;
+            }
+
+            return result.WebHostBuilderFactory(args);
+        }
+
+        public static IWebHostBuilder CreateFromTypeAssemblyEntryPoint<T>(string[] args) =>
```
Tratcher: `TypesAssembly`?

test/TestAssets/BuildWebHostPatternTestSite/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace BuildWebHostPatternTestSite
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+        }
+
+        public static IWebHost BuildWebHost(string[] args) => null;
```
Tratcher: null? Your tests should execute the resolved factory to make sure it returns the expected result.

javiercn: I don't have a server so I can't build a WebHost. The logic is only about finding the method, what it does when called is the responsibility of the CreateWebHostBuilder method.

src/Microsoft.AspNetCore.TestHost/WebHostBuilderFactory.cs
javiercn: will change before merging

test/TestAssets/BuildWebHostPatternTestSite/Program.cs
Tratcher: You're creating a factory, but you don't test that your factory works correctly. Returning a fake IWebHost implementation would be adequate.

Issue #1163, opened 2017-08-11
Generic Host


davidfowl:
## Generic Host

As part of ASP.NET Core, we introduced a set of cross cutting concerns that are fundamental to building many applications.
These include things like configuration, dependency injection, and logging. The `WebHostBuilder` is the glue that brings those common cross cutting
concerns together but it also couples the building of the HTTP pipeline into the same component. The goal of the generic host is to de-couple
the HTTP pipeline building from the Host API to enable other scenarios like messaging, background tasks, and supporting other non HTTP workloads that would benefit
from the same cross-cutting libraries that HTTP benefits from.

### Introducing Microsoft.Extensions.Hosting

Microsoft.Extensions.Hosting is the generic hosting library that will offer a similar API to `WebHostBuilder` but won't be specific to HTTP.
The eventual goal is to completely replace the `WebHostBuilder` API with this new one. The generic host will offer a few building blocks
for application models:

#### IHostedService

```C#
public interface IHostedService
{
    Task StartAsync(CancellationToken cancellationToken);

    Task StopAsync(CancellationToken cancellationToken);
}
```

This is the entry point to code execution. Many of these may be registered in as services. They will be resolved and executed **in order of registration**. StartAsync will be called when the host starts, and StopAsync will be called in **reverse registration order** when the host shuts down gracefully. IHostedServices are registered as part of container registration and resolved by the host as an `IEnumerable<IHostedService>`.

#### IHostBuilder

```C#
public interface IHostBuilder 
{
    IHost Build();
    
    IHostBuilder ConfigureAppConfiguration(Action<HostBuilderContext, IConfigurationBuilder> configureDelegate);

    IHostBuilder ConfigureHostConfiguration(Action<IConfigurationBuilder> configureDelegate);
    
    IHostBuilder ConfigureServices(Action<HostBuilderContext, IServiceCollection> configureServices);

    IHostBuilder ConfigureContainer<TContainerBuilder>(Action<HostBuilderContext, TContainerBuilder> configureContainer);

    IHostBuilder UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory);
    
    IDictionary<object, object> Properties { get; }
}
```

This is the main component that libraries and applications will extend to light up functionality.

#### IHost

```C#
public interface IHost : IDisposable
{
   IServiceProvider Services { get; }
   Task StartAsync(CancellationToken cancellationToken = default(CancellationToken));
[...]

PR #1204, opened 2017-09-11
Generic Host implementation

Tratcher:
#1163 This is a functional, tested API and implementation for generic hosts. It may not yet have enough sugar, but we can iterate on that beyond in followup issues.

- Should I file a followup issue to include a default timeout for IHost.StopAsync like we have for IWebHost? - https://github.com/aspnet/Hosting/issues/1208

### DIFF
--- a/samples/GenericHostSample/MyServiceA.cs
+++ b/samples/GenericHostSample/MyServiceA.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace GenericHostSample
+{
+    public class MyServiceA : IHostedService
+    {
+        private bool _stopping;
+        private Task _backgroundTask;
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            Console.WriteLine("MyServiceA is starting.");
+            _backgroundTask = BackgroundTask();
+            return Task.CompletedTask;
+        }
+
+        private async Task BackgroundTask()
+        {
+            while (!_stopping)
+            {
+                await Task.Delay(TimeSpan.FromSeconds(5));
+                Console.WriteLine("MyServiceA is doing background work.");
+            }
+
+            Console.WriteLine("MyServiceA background task is stopping.");
+        }
+
+        public async Task StopAsync(CancellationToken cancellationToken)
+        {
+            Console.WriteLine("MyServiceA is stopping.");
+            _stopping = true;
+            if (_backgroundTask != null)
+            {
+                // TODO: cancellation
+                await _backgroundTask;
+            }
+        }
+    }
+}
--- a/samples/GenericHostSample/MyServiceB.cs
+++ b/samples/GenericHostSample/MyServiceB.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace GenericHostSample
+{
+    public class MyServiceB : IHostedService
+    {
+        private bool _stopping;
+        private Task _backgroundTask;
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            Console.WriteLine("MyServiceB is starting.");
+            _backgroundTask = BackgroundTask();
+            return Task.CompletedTask;
+        }
+
+        private async Task BackgroundTask()
+        {
+            while (!_stopping)
+            {
+                await Task.Delay(TimeSpan.FromSeconds(7));
+                Console.WriteLine("MyServiceB is doing background work.");
+            }
+
+            Console.WriteLine("MyServiceB background task is stopping.");
+        }
+
+        public async Task StopAsync(CancellationToken cancellationToken)
+        {
+            Console.WriteLine("MyServiceB is stopping.");
+            _stopping = true;
+            if (_backgroundTask != null)
+            {
+                // TODO: cancellation
+                await _backgroundTask;
+            }
+        }
+    }
+}
--- a/samples/GenericHostSample/ProgramFullControl.cs
+++ b/samples/GenericHostSample/ProgramFullControl.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace GenericHostSample
+{
+    public class ProgramFullControl
+    {
+        public static async Task Main(string[] args)
+        {
+            var host = new HostBuilder()
+                .UseServiceProviderFactory<MyContainer>(new MyContainerFactory())
+                .ConfigureContainer<MyContainer>((hostContext, container) =>
+                {
+                })
+                .ConfigureAppConfiguration((hostContext, config) =>
+                {
+                    config.AddEnvironmentVariables();
+                    config.AddJsonFile("appsettings.json", optional: true);
+                    config.AddCommandLine(args);
+                })
+                .ConfigureServices((hostContext, services) =>
+                {
+                    services.AddScoped<IHostedService, MyServiceA>();
+                    services.AddScoped<IHostedService, MyServiceB>();
+                })
+                .Build();
+
+            var s = host.Services;
+
+            using (host)
+            {
+                Console.WriteLine("Starting!");
+
+                await host.StartAsync();
+
+                Console.WriteLine("Started! Press <enter> to stop.");
+
+                Console.ReadLine();
+
+                Console.WriteLine("Stopping!");
+
+                await host.StopAsync();
+
+                Console.WriteLine("Stopped!");
+            }
+        }
+    }
+}
--- a/samples/GenericHostSample/ServiceBaseLifetime.cs
+++ b/samples/GenericHostSample/ServiceBaseLifetime.cs
@@ -1,0 +1,64 @@
+﻿#if NET461
+using System;
+using System.ServiceProcess;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace GenericHostSample
+{
+    public static class ServiceBaseLifetimeHostExtensions
+    {
+        public static IHostBuilder UseServiceBaseLifetime(this IHostBuilder hostBuilder)
+        {
+            return hostBuilder.ConfigureServices((hostContext, services) => services.AddSingleton<IHostLifetime, ServiceBaseLifetime>());
+        }
+
+        public static Task RunAsServiceAsync(this IHostBuilder hostBuilder, CancellationToken cancellationToken = default)
+        {
+            return hostBuilder.UseServiceBaseLifetime().Build().RunAsync(cancellationToken);
+        }
+    }
+
+    public class ServiceBaseLifetime : ServiceBase, IHostLifetime
+    {
+        private Action<object> _startCallback;
+        private Action<object> _stopCallback;
+        private object _startState;
+        private object _stopState;
+
+        public void RegisterDelayStartCallback(Action<object> callback, object state)
+        {
+            _startCallback = callback ?? throw new ArgumentNullException(nameof(callback));
+            _startState = state ?? throw new ArgumentNullException(nameof(state));
+
+            Run(this);
+        }
+
+        public void RegisterStopCallback(Action<object> callback, object state)
+        {
+            _stopCallback = callback ?? throw new ArgumentNullException(nameof(callback));
+            _stopState = state ?? throw new ArgumentNullException(nameof(state));
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken)
+        {
+            Stop();
+            return Task.CompletedTask;
+        }
+
+        protected override void OnStart(string[] args)
+        {
+            _startCallback(_startState);
+            base.OnStart(args);
+        }
+
+        protected override void OnStop()
+        {
+            _stopCallback(_stopState);
+            base.OnStop();
+        }
+    }
+}
+#endif
--- a/samples/GenericWebHost/WebHostExtensions.cs
+++ b/samples/GenericWebHost/WebHostExtensions.cs
@@ -1,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Hosting.Internal;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.ObjectPool;
+
+namespace GenericWebHost
+{
+    public static class WebHostExtensions
+    {
+        public static IHostBuilder ConfigureWebHost(this IHostBuilder builder, Action<HostBuilderContext, IApplicationBuilder> configureApp)
+        {
+            return builder.ConfigureServices((bulderContext, services) =>
+            {
+                services.Configure<WebHostServiceOptions>(options =>
+                {
+                    options.ConfigureApp = configureApp;
+                });
+                services.AddSingleton<IHostedService, WebHostService>();
+                
+                var listener = new DiagnosticListener("Microsoft.AspNetCore");
+                services.AddSingleton<DiagnosticListener>(listener);
+                services.AddSingleton<DiagnosticSource>(listener);
+                
+                services.AddTransient<IHttpContextFactory, HttpContextFactory>();
+                services.AddScoped<IMiddlewareFactory, MiddlewareFactory>();
+
+                // Conjure up a RequestServices
+                services.AddTransient<IStartupFilter, AutoRequestServicesStartupFilter>();
+                services.AddTransient<IServiceProviderFactory<IServiceCollection>, DefaultServiceProviderFactory>();
+
+                // Ensure object pooling is available everywhere.
+                services.AddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
+            });
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Hosting.Abstractions/HostBuilderContext.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/HostBuilderContext.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Context containing the common services on the <see cref="IHost" />. Some properties may be null until set by the <see cref="IHost" />.
+    /// </summary>
+    public class HostBuilderContext
+    {
+        public HostBuilderContext(IDictionary<object, object> properties)
+        {
+            Properties = properties ?? throw new System.ArgumentNullException(nameof(properties));
+        }
+
+        /// <summary>
+        /// The <see cref="IHostingEnvironment" /> initialized by the <see cref="IHost" />.
+        /// </summary>
+        public IHostingEnvironment HostingEnvironment { get; set; }
+
+        /// <summary>
+        /// The <see cref="IConfiguration" /> containing the merged configuration of the application and the <see cref="IHost" />.
+        /// </summary>
+        public IConfiguration Configuration { get; set; }
+        
+        /// <summary>
+        /// A central location for sharing state between components during the host building process.
+        /// </summary>
+        public IDictionary<object, object> Properties { get; }
+    }
+}
--- a/src/Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostBuilderExtensions.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostBuilderExtensions.cs
@@ -1,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public static class HostingAbstractionsHostBuilderExtensions
+    {
+        /// <summary>
+        /// Start the web host and listen on the specified urls.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IHostBuilder"/> to start.</param>
+        /// <returns>The <see cref="IHostBuilder"/>.</returns>
+        public static IHost Start(this IHostBuilder hostBuilder)
+        {
+            var host = hostBuilder.Build();
+            host.StartAsync(CancellationToken.None).GetAwaiter().GetResult();
+            return host;
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostExtensions.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostExtensions.cs
@@ -1,0 +1,95 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public static class HostingAbstractionsHostExtensions
+    {
+        /// <summary>
+        /// Starts the host synchronously.
+        /// </summary>
+        /// <param name="host"></param>
+        public static void Start(this IHost host)
+        {
+            host.StartAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Attempts to gracefully stop the host with the given timeout.
+        /// </summary>
+        /// <param name="host"></param>
+        /// <param name="timeout">The timeout for stopping gracefully. Once expired the
+        /// server may terminate any remaining active connections.</param>
+        /// <returns></returns>
+        public static Task StopAsync(this IHost host, TimeSpan timeout)
+        {
+            return host.StopAsync(new CancellationTokenSource(timeout).Token);
+        }
+
+        /// <summary>
+        /// Block the calling thread until shutdown is triggered via Ctrl+C or SIGTERM.
+        /// </summary>
+        /// <param name="host">The running <see cref="IHost"/>.</param>
+        public static void WaitForShutdown(this IHost host)
+        {
+            host.WaitForShutdownAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Runs a web application and block the calling thread until host shutdown.
+        /// </summary>
+        /// <param name="host">The <see cref="IHost"/> to run.</param>
+        public static void Run(this IHost host)
+        {
+            host.RunAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Runs a web application and returns a Task that only completes when the token is triggered or shutdown is triggered.
+        /// </summary>
+        /// <param name="host">The <see cref="IHost"/> to run.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static async Task RunAsync(this IHost host, CancellationToken token = default(CancellationToken))
+        {
+            using (host)
+            {
+                await host.StartAsync(token);
+
+                await host.WaitForShutdownAsync(token);
+            }
+        }
+
+        /// <summary>
+        /// Returns a Task that completes when shutdown is triggered via the given token.
+        /// </summary>
+        /// <param name="host">The running <see cref="IHost"/>.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static async Task WaitForShutdownAsync(this IHost host, CancellationToken token = default(CancellationToken))
+        {
+            var applicationLifetime = host.Services.GetService<IApplicationLifetime>();
+
+            token.Register(state =>
+            {
+                ((IApplicationLifetime)state).StopApplication();
+            },
+            applicationLifetime);
+
+            var waitForStop = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+            applicationLifetime.ApplicationStopping.Register(obj =>
+            {
+                var tcs = (TaskCompletionSource<object>)obj;
+                tcs.TrySetResult(null);
+            }, waitForStop);
+
+            await waitForStop.Task;
+
+            // WebHost will use its default ShutdownTimeout if none is specified.
+            await host.StopAsync();
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Hosting.Abstractions/HostingEnvironmentExtensions.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/HostingEnvironmentExtensions.cs
@@ -1,0 +1,79 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Extension methods for <see cref="IHostingEnvironment"/>.
+    /// </summary>
+    public static class HostingEnvironmentExtensions
+    {
+        /// <summary>
+        /// Checks if the current hosting environment name is <see cref="EnvironmentName.Development"/>.
+        /// </summary>
+        /// <param name="hostingEnvironment">An instance of <see cref="IHostingEnvironment"/>.</param>
+        /// <returns>True if the environment name is <see cref="EnvironmentName.Development"/>, otherwise false.</returns>
+        public static bool IsDevelopment(this IHostingEnvironment hostingEnvironment)
+        {
+            if (hostingEnvironment == null)
+            {
+                throw new ArgumentNullException(nameof(hostingEnvironment));
+            }
+
+            return hostingEnvironment.IsEnvironment(EnvironmentName.Development);
+        }
+
+        /// <summary>
+        /// Checks if the current hosting environment name is <see cref="EnvironmentName.Staging"/>.
+        /// </summary>
+        /// <param name="hostingEnvironment">An instance of <see cref="IHostingEnvironment"/>.</param>
+        /// <returns>True if the environment name is <see cref="EnvironmentName.Staging"/>, otherwise false.</returns>
+        public static bool IsStaging(this IHostingEnvironment hostingEnvironment)
+        {
+            if (hostingEnvironment == null)
+            {
+                throw new ArgumentNullException(nameof(hostingEnvironment));
+            }
+
+            return hostingEnvironment.IsEnvironment(EnvironmentName.Staging);
+        }
+
+        /// <summary>
+        /// Checks if the current hosting environment name is <see cref="EnvironmentName.Production"/>.
+        /// </summary>
+        /// <param name="hostingEnvironment">An instance of <see cref="IHostingEnvironment"/>.</param>
+        /// <returns>True if the environment name is <see cref="EnvironmentName.Production"/>, otherwise false.</returns>
+        public static bool IsProduction(this IHostingEnvironment hostingEnvironment)
+        {
+            if (hostingEnvironment == null)
+            {
+                throw new ArgumentNullException(nameof(hostingEnvironment));
+            }
+
+            return hostingEnvironment.IsEnvironment(EnvironmentName.Production);
+        }
+
+        /// <summary>
+        /// Compares the current hosting environment name against the specified value.
+        /// </summary>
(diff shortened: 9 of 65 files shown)


src/Microsoft.Extensions.Hosting.Abstractions/HostBuilderContext.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// The <see cref="IHostingEnvironment" /> initialized by the <see cref="IHost" />.
+        /// </summary>
+        public IHostingEnvironment HostingEnvironment { get; set; }
+
+        /// <summary>
+        /// The <see cref="IConfiguration" /> containing the merged configuration of the application and the <see cref="IHost" />.
+        /// </summary>
+        public IConfiguration Configuration { get; set; }
+        
+        IDictionary<object, object> Properties { get; }
```
Tratcher: This is not used anywhere. Should it be?

src/Microsoft.Extensions.Hosting.Abstractions/IHostBuilder.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        IHostBuilder ConfigureAppConfiguration(Action<HostBuilderContext, IConfigurationBuilder> configureDelegate);
+
+        IHostBuilder ConfigureServices(Action<HostBuilderContext, IServiceCollection> configureDelegate);
+
+        IHostBuilder ConfigureContainer<TContainerBuilder>(Action<HostBuilderContext, TContainerBuilder> configureDelegate);
+
+        IHostBuilder UseServiceProviderFactory<TContainerBuilder>(IServiceProviderFactory<TContainerBuilder> factory);
+
+        IDictionary<object, object> Properties { get; }
```
Tratcher: This is not used anywhere. Should it be?

src/Microsoft.Extensions.Hosting.Abstractions/IHostLifetime.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public interface IHostLifetime
```
Tratcher: These method names need work. OnStarted blocks the start process, so it's more like OnStarting.

OnStopping initiates shutdown indirectly, chaining things like CTL+C to `IApplicationLifetime.StopApplication()`. It's not clear why it needs a callback do to that, it could Inject IApplicationLifetime directly.

StopAsync informs the HostLifetime instance that the host was stopped. This seems more like Dispose.

davidfowl: > StopAsync informs the HostLifetime instance that the host was stopped. This seems more like Dispose.

Lets not conflate StopAsync and Dispose again 😄 . It'll already automagically get disposed if it implements IDisposable.

src/Microsoft.Extensions.Hosting/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+            return hostBuilder.ConfigureServices((context, collection) => collection.AddSingleton<IHostLifetime, ConsoleLifetime>());
+        }
+
+        /// <summary>
+        /// Enables console support, builds and starts the host, and waits for Ctrl+C or SIGTERM to shut down.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IHostBuilder" /> to configure.</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public static Task RunConsoleAsync(this IHostBuilder hostBuilder, CancellationToken cancellationToken = default(CancellationToken))
```
davidfowl: This is nice

src/Microsoft.Extensions.Hosting/Internal/ConsoleLifetime.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public class ConsoleLifetime : IHostLifetime
+    {
+        public ConsoleLifetime(IHostingEnvironment environment, IApplicationLifetime applicationLifetime)
+        {
+            Environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            ApplicationLifetime = applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime));
+            ApplicationLifetime.ApplicationStarted.Register(() =>
+            {
+                Console.WriteLine("Application starting. Press Ctrl+C to shut down.");
```
davidfowl: I dunno if we should repeat the same mistake here... Is this the only reason we wire up the event 😄

Tratcher: At least now it's contained in the ConsoleLifetime instance. It can now have its own options, or just be replaced.

src/Microsoft.Extensions.Hosting.Abstractions/IHostingEnvironment.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Provides information about the web hosting environment an application is running in.
+    /// </summary>
+    public interface IHostingEnvironment
+    {
+        /// <summary>
+        /// Gets or sets the name of the environment. This property is automatically set by the host to the value
+        /// of the "ASPNETCORE_ENVIRONMENT" environment variable, or "environment" as specified in any other configuration source.
```
davidfowl: comment isn't right

src/Microsoft.Extensions.Hosting.Abstractions/Microsoft.Extensions.Hosting.Abstractions.csproj
```diff
@@ -7,6 +7,14 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>hosting</PackageTags>
     <EnableApiCheck>false</EnableApiCheck>
+    <RootNamespace>Microsoft.Extensions.Hosting</RootNamespace>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" />
```
davidfowl: We should think about using options as well.

src/Microsoft.Extensions.Hosting/Internal/ConsoleLifetime.cs
davidfowl: We need to add this back as the default lifetime implementation:

https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs#L137

Process exit.

src/Microsoft.Extensions.Hosting.Abstractions/Microsoft.Extensions.Hosting.Abstractions.csproj
Tratcher: There's no Options.Abstractions package. Options is pulled in from Extensions.Hosting via Logging.

samples/GenericWebHost/WebHostExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.ObjectPool;
+
+namespace GenericWebHost
+{
+    public static class WebHostExtensions
+    {
+        public static IHostBuilder ConfigureWebHost(this IHostBuilder builder, Action<HostBuilderContext, IApplicationBuilder> configureApp)
```
davidfowl: Do you think we should ship this in the Hosting package?

samples/GenericWebHost/WebHostExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            return builder.ConfigureServices((bulderContext, services) =>
+            {
+                services.Configure<WebHostServiceOptions>(options =>
+                {
+                    options.ConfigureApp = configureApp;
```
davidfowl: Options hmm.

samples/GenericWebHost/WebHostExtensions.cs
Tratcher: Not yet. This was just a sample for how we might redesign WebHost in the future.

src/Microsoft.Extensions.Hosting/Internal/Host.cs
```diff
@@ -0,0 +1,98 @@
...
+        private IHostLifetime _hostLifetime;
+        private ApplicationLifetime _applicationLifetime;
+        private IEnumerable<IHostedService> _hostedServices;
+
+        internal Host(IServiceProvider services)
+        {
+            Services = services ?? throw new ArgumentNullException(nameof(services));
+            _applicationLifetime = Services.GetRequiredService<IApplicationLifetime>() as ApplicationLifetime;
+            _logger = Services.GetRequiredService<ILogger<Host>>();
+            _hostLifetime = Services.GetService<IHostLifetime>() ?? new NoOpHostLifetime();
```
davidfowl: Why is this optional?

src/Microsoft.Extensions.Hosting/Internal/Host.cs
```diff
@@ -0,0 +1,98 @@
...
+                    catch (Exception ex)
+                    {
+                        exceptions.Add(ex);
+                    }
+                }
+            }
+
+            // Fire IApplicationLifetime.Stopped
+            _applicationLifetime?.NotifyStopped();
+            await _hostLifetime.StopAsync(cancellationToken);
```
davidfowl: Fix this bug, this needs to be stopping, not stopped.

src/Microsoft.Extensions.Hosting/Internal/Host.cs
Tratcher: IHost.Start/StopAsync can function fine without it.

Tratcher: Or should I pre-register NoOpHostLifetime?

davidfowl: I think the default would be the AppDomain.ProcessExit one.

src/Microsoft.Extensions.Hosting/Internal/Host.cs
Tratcher: IApplicationLifetime.Stopped? IApplicationLifetime.Stopping is triggered before you stop the services, and Stopped is afterwards.

davidfowl: Stopping. IHostedService.Stop should map to IApplicationLifetime.Stopping,  it shouldn't be called after.

Tratcher: So when does IApplicationLifetime.Stopped get called? or do you plan to eliminate it?
The current flow:
IHost.StopAsync
- IApplicationLifetime.Stopping
- IHostedService(s).StopAsync
- IApplicationLifetime.Stopped
- IHostLifetime.StopAsync

Switching IApplicationLifetime.Stopped and  IHostLifetime.StopAsync might make sense, but Stopping belongs before the calls to Stop.

davidfowl: We can consider that, I'm not sure we need Stopped, (I'm not sure why it's used today) Dispose kinda covers it (maybe Stopped maps to Dispose).

src/Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostExtensions.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        /// <summary>
+        /// Returns a Task that completes when shutdown is triggered via the given token.
+        /// </summary>
+        /// <param name="host">The running <see cref="IHost"/>.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static async Task WaitForShutdownAsync(this IHost host, CancellationToken token = default(CancellationToken))
+        {
+            var done = new ManualResetEventSlim(false);
```
JunTaoLuo: This isn't needed

src/Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostExtensions.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            host.WaitForShutdownAsync().GetAwaiter().GetResult();
+        }
+
+        /// <summary>
+        /// Returns a Task that completes when shutdown is triggered via the given token.
+        /// </summary>
+        /// <param name="host">The running <see cref="IHost"/>.</param>
+        /// <param name="token">The token to trigger shutdown.</param>
+        public static async Task WaitForShutdownAsync(this IHost host, CancellationToken token = default(CancellationToken))
```
JunTaoLuo: Ad discussed WaitForTokenShutdownAsync can be renamed to WaitForShutdownAsyn in this scenario.

src/Microsoft.Extensions.Hosting.Abstractions/HostingEnvironmentExtensions.cs
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
```
JunTaoLuo: What's being used from System.IO?

src/Microsoft.Extensions.Hosting.Abstractions/IHostBuilder.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    /// <summary>
+    /// A program initialization abstraction.
+    /// </summary>
+    public interface IHostBuilder
+    {
+        /// <summary>
+        /// A central location for sharing state between components during the host building process.
+        /// </summary>
+        IDictionary<object, object> Properties { get; }
```
JunTaoLuo: Is this something we want to add to WebHostBuilder as well (in the next major release given that this is a breaking change to the interface)?

src/Microsoft.Extensions.Hosting.Abstractions/IHostLifetime.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public interface IHostLifetime
+    {
+        /// <summary>
+        /// Called at the start of <see cref="IHost.StartAsync(CancellationToken)"/> which will wait until the callback is invoked before
+        /// continuing. This can be used to delay startup until the signaled by an external event.
```
JunTaoLuo: until the signaled => until signaled.

src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public static class HostingHostBuilderExtensions
```
JunTaoLuo: Why is this separate from HostBuilderExtensions?

src/Microsoft.Extensions.Hosting/Internal/ConsoleLifetime.cs
```diff
@@ -0,0 +1,52 @@
...
+            ApplicationLifetime = applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime));
+            ApplicationLifetime.ApplicationStarted.Register(() =>
+            {
+                Console.WriteLine("Application started. Press Ctrl+C to shut down.");
+                Console.WriteLine($"Hosting environment: {Environment.EnvironmentName}");
+                Console.WriteLine($"Content root path: {Environment.ContentRootPath}");
+            });
+        }
+
+        public IHostingEnvironment Environment { get; }
```
JunTaoLuo: Is there a need to expose this? It seems to be only used by the constructor here.

src/Microsoft.Extensions.Hosting.Abstractions/IHostBuilder.cs
davidfowl: Nah.

src/Microsoft.Extensions.Hosting/HostBuilder.cs
```diff
@@ -0,0 +1,211 @@
...
+            services.AddLogging();
+            
+            foreach (var configureServicesAction in _configureServicesActions)
+            {
+                configureServicesAction(_hostBuilderContext, services);
+            }
+
+            // We can't invoke the factory directly because we don't know the genric type at compile time.
+            // TContainerBuilder IServiceProviderFactory<TContainerBuilder>.CreateBuilder(IServiceCollection)
+            var createBuilderMethod = _serviceProviderFactory.GetType().GetMethod("CreateBuilder");
```
JunTaoLuo: Need to handle exceptions? What if there are overloads of CreateBuilder in the type? AmbiguousMatchException? In fact, this approach looks odd overall.

src/Microsoft.Extensions.Hosting.Abstractions/IHostLifetime.cs
glennc: There are weirdnesses here that I want to understand. Some of which might be a bit existential.

Why does IHostLifetime need to notify IApplicationLifetime if it is called by the IHost and the IHost already has IApplicationLifetime?

I don't understand why OnStarted isn't just `public Task OnStarted` where the `IHost` will await it, allowing the specific HostLifetime to delay if it wants to or just return. Doing this would what I think is the contract clearer.

The delegate appears to only be useful in a Host that doesn't use `IApplicationLifetime`. In our host implementations it is never going to be different to what is currently passed to the `HostLifetime`, correct? If this is true is not using `IApplicationLifetime` something we would expect an `IHost` to do, or do we view `IApplicationLifetime` as an integral part of the contract that a Host should adhere to?

Tratcher: Agreed, I don't see why the host needs to register with StopAsync if all it's going to register is a call to IApplicationLifetime.StopApplication. The IHostLifetime can call IApplicationLifetime directly.

And yes, OnStarted could go with a task model, it would just have to be cancellable.

@davidfowl ?

davidfowl: > I don't understand why OnStarted isn't just public Task OnStarted where the IHost will await it, allowing the specific HostLifetime to delay if it wants to or just return. Doing this would what I think is the contract clearer.

That works. IMO it's a calling pattern. We're talking about calling a method when something starts vs waiting until something starts to complete the Task returned. Semantically, calling StartAsync on IHostLifetime is bizarre as you're really asking the object to call you back when it has started. It's more like `IHostLifetime.WaitForStartAsync()`. See our HttpResponse. OnStarting event for a similar pattern.

> The delegate appears to only be useful in a Host that doesn't use IApplicationLifetime. In our host implementations it is never going to be different to what is currently passed to the HostLifetime, correct? If this is true is not using IApplicationLifetime something we would expect an IHost to do, or do we view IApplicationLifetime as an integral part of the contract that a Host should adhere to?
  
> Agreed, I don't see why the host needs to register with StopAsync if all it's going to register is a call to IApplicationLifetime.StopApplication. The IHostLifetime can call IApplicationLifetime directly.

I feel like I'm missing something here. The entire point of this interface is the abstract the lifetime from the IHost implementation so that things like Run and RunAsync can be implemented without hardcoding ties into various lifetime mechanisms. The IApplicationLifetime is about firing host events as they happen so the application can react.

src/Microsoft.Extensions.Hosting/HostBuilder.cs
Tratcher: I don't think we need to get too fancy here. Providing a factory at all is less common, and it's restricted to a specific interface. We can search CreateBuilder and CreateServiceProvider overloads if people ask for it. It's pretty ugly code that I'd rather avoid if we can.

As for handling exceptions, I'm not too concerned. The inputs were restricted via strong types so we know they at least match the interface signature. Reflection has good exceptions for duplicate methods so we don't need to wrap that. The only interesting error will be if someone's generics don't match between UseServiceProviderFactory and ConfigureContainer, and the reflection exception in that case is pretty clear too.

samples/GenericWebHost/WebHostExtensions.cs
anurse: I think that's a second set of changes anyway.

samples/GenericWebHost/WebHostExtensions.cs
anurse: Another option would be to use `IStartup` (a modified version with only a `Configure` method) and have this register a delegate-based Startup class.

src/Microsoft.Extensions.Hosting.Abstractions/IHostLifetime.cs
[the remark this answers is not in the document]
davidfowl: You can be it doesn't matter right now.

src/Microsoft.Extensions.Hosting/HostBuilder.cs
```diff
@@ -0,0 +1,211 @@
...
+        private IConfiguration _hostConfiguration;
+        private IConfiguration _appConfiguration;
+        private HostBuilderContext _hostBuilderContext;
+        private IHostingEnvironment _hostingEnvironment;
+        private IServiceProvider _appServices;
+
+        /// <summary>
+        /// A central location for sharing state between components during the host building process.
+        /// </summary>
+        public IDictionary<object, object> Properties => new Dictionary<object, object>();
```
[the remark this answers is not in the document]
davidfowl: Not sure what you mean. This is a property bag of object-> object, there's no config here. It's used for things like idempotency and sharing state across calls.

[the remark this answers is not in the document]
Tratcher: It is the same instance as on the HostBuilderContext.

src/Microsoft.Extensions.Hosting/Internal/LoggerEventIds.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Hosting.Internal
+{
+    internal static class LoggerEventIds
+    {
+        public const int Starting = 1;
```
anurse: Can we use `EventId` and give them names as well?

src/Microsoft.Extensions.Hosting/Internal/ConsoleLifetime.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// </summary>
+    public class ConsoleLifetime : IHostLifetime
+    {
+        public ConsoleLifetime(IHostingEnvironment environment, IApplicationLifetime applicationLifetime)
+        {
+            Environment = environment ?? throw new ArgumentNullException(nameof(environment));
+            ApplicationLifetime = applicationLifetime ?? throw new ArgumentNullException(nameof(applicationLifetime));
+            ApplicationLifetime.ApplicationStarted.Register(() =>
+            {
+                Console.WriteLine("Application started. Press Ctrl+C to shut down.");
```
anurse: This shouldn't be in our built-in ConsoleLifetime.

src/Microsoft.Extensions.Hosting/HostBuilder.cs
anurse: We should be able to do this without reflection via a few internal adapter types that have non-generic interfaces. I think that would be preferable to reflection...

src/Microsoft.Extensions.Hosting/HostBuilderExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public static class HostBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a delegate for configuring the provided <see cref="LoggerFactory"/>. This may be called multiple times.
```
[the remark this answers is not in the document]
anurse: Well, it's an `ILoggingBuilder` though. I get that `ILoggerFactory` is the final result, but this seems like a doc comment that just wasn't updated when we introduced `ILoggingBuilder`.

samples/GenericHostSample/MyServiceA.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace GenericHostSample
+{
+    public class MyServiceA : IHostedService
+    {
+        private bool _stopping;
```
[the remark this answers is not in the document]
Tratcher: That's more these samples warrant. There aren't any significant multi-threading concerns here.

samples/GenericHostSample/ProgramFullControl.cs
```diff
@@ -0,0 +1,51 @@
...
+                    config.AddCommandLine(args);
+                })
+                .ConfigureServices((hostContext, services) =>
+                {
+                    services.AddScoped<IHostedService, MyServiceA>();
+                    services.AddScoped<IHostedService, MyServiceB>();
+                })
+                .Build();
+
+            var s = host.Services;
```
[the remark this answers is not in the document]
Tratcher: Yes, just for visibility.

samples/GenericWebHost/WebHostExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+                });
+                services.AddSingleton<IHostedService, WebHostService>();
+                
+                var listener = new DiagnosticListener("Microsoft.AspNetCore");
+                services.AddSingleton<DiagnosticListener>(listener);
```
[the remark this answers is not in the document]
Tratcher: When the same instance is registered twice it's better to be explicit as to why.

src/Microsoft.Extensions.Hosting/Internal/ConsoleLifetime.cs
Tratcher: Discussed earlier with fowler. This is existing behavior that has been consolidated to an appropriate place. If you're not using console it goes away. We can also give ConsoleOptions to control it. Note this used to be built into RunAsync, without it you don't get any startup output now.

anurse: Fair, but definitely give me an option. I want to be able to use this for command line tools (I have some design thoughts for that, but they're all out of scope for our built in components) and I definitely don't want this printed out in those cases.

src/Microsoft.Extensions.Hosting/Internal/LoggerEventIds.cs
anurse: https://github.com/aspnet/Hosting/issues/1205 filed for never.

samples/GenericHostSample/MyServiceA.cs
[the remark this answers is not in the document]
anurse: Samples are guidance and best-practices for our users. Being correct is important.

samples/GenericHostSample/ProgramFullControl.cs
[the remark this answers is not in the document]
anurse: Not sure it's needed. IntelliSense would show it, so it doesn't exactly **need** to be "documented" via a sample.

samples/GenericWebHost/WebHostExtensions.cs
Tratcher: We'll iterate on the generic web host design between now and 3.0. We're starting with the proof of concept.

samples/GenericHostSample/MyServiceA.cs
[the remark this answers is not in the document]
Tratcher: Any examples? We don't run many background threads.

src/Microsoft.Extensions.Hosting/Internal/ServiceFactoryAdapter.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Hosting.Internal
+{
+    public abstract class ServiceFactoryAdapter
```
JunTaoLuo: Why an abstract class instead of an interface? There are no members/properties, default implementation and no non-public access modifiers.

Tratcher: We've used abstracts elsewhere like security to have a base version of the same class without the generic. In this very simple scenario an interface could also be used. There's no strict reason either way.

JunTaoLuo: I think it's better to use an interface until there's proven need for an abstract base class. The base classes in Security seem to have properties or default method implementation which aren't possible on interfaces. The only reason to keep this as a class may be if we anticipate to add additional methods in the near future and I don't think there is reason to. Also even if it's trivial, let's keep the separate classes in separate files.

test/Microsoft.Extensions.Hosting.Tests/Fakes/FakeServiceCollection.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
```
JunTaoLuo: copyright

samples/GenericWebHost/WebHostExtensions.cs
anurse: Agreed

src/Microsoft.Extensions.Hosting/Internal/ServiceFactoryAdapter.cs
anurse: Should this be true-internal? This isn't something anyone would actually be able to affect. We directly construct instances of it in the HostBuilder. The "abstraction" is not a public-facing abstraction, it's purely to handle a generic value.

davidfowl: Agreed, make it internal.

samples/GenericHostSample/MyServiceB.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace GenericHostSample
+{
+    public class MyServiceB : IHostedService
+    {
+        private bool _stopping;
```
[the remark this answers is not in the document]
jkotalik: For my own knowledge, what does CTS stand for?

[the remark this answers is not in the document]
davidfowl: https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokensource?view=netcore-2.0

src/Microsoft.Extensions.Hosting.Abstractions/IHostLifetime.cs
davidfowl: I was thinking of a more succinct way to put this:

- The IHostLifetime is used by the host. After a bit of thinking, I think they should potentially stack (support multiple implementations).
- The IApplicationLifetime is consumed by the application and implemented by the host.

src/Microsoft.Extensions.Hosting/Internal/ApplicationLifetime.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Hosting.Internal
+{
+    /// <summary>
+    /// Allows consumers to perform cleanup during a graceful shutdown.
+    /// </summary>
+    public class ApplicationLifetime : IApplicationLifetime
```
davidfowl: @Tratcher I think this should be implemented on the IHost itself.

src/Microsoft.Extensions.Hosting.Abstractions/IHostLifetime.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.Extensions.Hosting
+{
+    public interface IHostLifetime
+    {
+        /// <summary>
+        /// Called at the start of <see cref="IHost.StartAsync(CancellationToken)"/> to see if the host is ready to start.
+        /// This can be used to delay startup until signaled by an external event.
+        /// </summary>
+        /// <param name="cancellationToken">Used to indicate that StartAsync has been aborted.</param>
+        Task WaitForStartAsync(CancellationToken cancellationToken);
```
davidfowl: After discussion with @Tratcher , we should revert this to the original interface definition.

src/Microsoft.Extensions.Hosting/Internal/ApplicationLifetime.cs
Tratcher: Let's punt that discussion

src/Microsoft.Extensions.Hosting/ConsoleLifetimeOptions.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Hosting
+{
+    public class ConsoleLifetimeOptions
```
davidfowl: Nice

src/Microsoft.Extensions.Hosting.Abstractions/HostingAbstractionsHostBuilderExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading;
+
+namespace Microsoft.Extensions.Hosting
+{
+    public static class HostingAbstractionsHostBuilderExtensions
+    {
+        /// <summary>
+        /// Start the web host and listen on the specified urls.
```
JunTaoLuo: nit: no web, there are a few more instances of this to clean up.

src/Microsoft.Extensions.Hosting/Internal/Host.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            Services = services ?? throw new ArgumentNullException(nameof(services));
+            _applicationLifetime = Services.GetRequiredService<IApplicationLifetime>() as ApplicationLifetime;
+            _logger = Services.GetRequiredService<ILogger<Host>>();
+            _hostLifetime = Services.GetRequiredService<IHostLifetime>();
+        }
+
+        public IServiceProvider Services { get; }
+
+        public async Task StartAsync(CancellationToken cancellationToken = default(CancellationToken))
```
JunTaoLuo: nit: default instead of default()

src/Microsoft.Extensions.Hosting/Internal/ConsoleLifetime.cs
```diff
@@ -0,0 +1,61 @@
...
+                });
+            }
+
+            // Console applications start immediately.
+            callback(state);
+        }
+
+        public void RegisterStopCallback(Action<object> callback, object state)
+        {
+            AppDomain.CurrentDomain.ProcessExit += (sender, eventArgs) => callback(state);
```
JunTaoLuo: Wondering if we should design the lifetimes to be inheritable. For example we could have ConsoleLifetime inherit from HostProcessLifetime and then add the callback registration for the Console.CancelKeyPress event. This may be useful if someone wants to extend ConsoleLifetime but with an additional event.

samples/GenericHostSample/ServiceBaseLifetime.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private Action<object> _startCallback;
+        private Action<object> _stopCallback;
+        private object _startState;
+        private object _stopState;
+
+        public void RegisterDelayStartCallback(Action<object> callback, object state)
+        {
+            _startCallback = callback ?? throw new ArgumentNullException(nameof(callback));
+            _startState = state ?? throw new ArgumentNullException(nameof(state));
```
JunTaoLuo: state could be null

src/Microsoft.Extensions.Hosting/Internal/HostProcessLifetime.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Hosting.Internal
+{
+    public class HostProcessLifetime : IHostLifetime
```
JunTaoLuo: If the other lifetime is called ConsoleLifetime, let's call this HostLifetime.

anurse: ProcessLifetime would be more accurate.

JunTaoLuo: Oops typo, I meant that. ~~Host~~ProcessLifetime

PR #1209, opened 2017-09-14
[Design] Add configure options startup sample

HaoK:
For https://github.com/aspnet/Options/issues/224

New startup sugar to make it easier to plug in a class that configure options using the new Options API

@davidfowl @dougbu @ajcvickers @divega

### DIFF
--- a/samples/SampleStartups/StartupConfigureOptions.cs
+++ b/samples/SampleStartups/StartupConfigureOptions.cs
@@ -1,0 +1,109 @@
+using System;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+
+// Note that this sample will not run. It is only here to illustrate usage patterns.
+
+namespace SampleStartups
+{
+    public class StartupConfigureOptions : StartupBase
+    {
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public override void Configure(IApplicationBuilder app)
+        {
+            app.Run(async (context) =>
+            {
+                await context.Response.WriteAsync("Hello World!");
+            });
+        }
+
+        // Entry point for the application.
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                //.UseKestrel()
+                .UseFakeServer()
+                .UseStartup<StartupConfigureOptions>()
+                // Register transient IConfigureOptions
+                .ConfigureOptions<ConfigureOptions>()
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -141,8 +131,34 @@ public static IWebHostBuilder ConfigureLogging(this IWebHostBuilder hostBuilder,
...
-        }
+            => hostBuilder.ConfigureServices((context, collection) => collection.AddLogging(builder => configureLogging(context, builder)));
+
+        /// <summary>
+        /// Registers an object that will have all of its I[Post]ConfigureOptions registered.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder" /> to configure.</param>
+        /// <param name="configureOptionsInstance">Object that will have all of its I[Post]ConfigureOptions registered.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder ConfigureOptions(this IWebHostBuilder hostBuilder, object configureOptionsInstance)
```
davidfowl: Why do we need these?

HaoK: Guess we don't, so are we okay with just having the new ConfigureOptions methods, and no new sugar needs to be added to Hosting for https://github.com/aspnet/Options/issues/224?

I can still add a sample/tests to hosting if we want, but if there's no new APIs, its mostly just as an example

samples/SampleStartups/StartupConfigureOptions.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        // Entry point for the application.
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                //.UseKestrel()
+                .UseFakeServer()
+                .UseStartup<StartupConfigureOptions>()
+                // Register transient IConfigureOptions
+                .ConfigureOptions<ConfigureOptions>()
```
HaoK: these would just become `.ConfigureServices(services => services.ConfigureOptions<ConfigureOptions>())`

Tratcher: Not much point in adding tests in Hosting if there's no Hosting API to test. You have tests in Options right?

HaoK: Yep, no point to touch hosting if we don't add any new APIs, which is why I sent this PR before actually adding the tests to hosting :)

PR #1212, opened 2017-09-15
Preserve state in HostBuilder.Properties

jdom:
`HostBuilder.Properties` was returning a new Dictionary on every access


davidfowl:
Good catch!

PR #1215, opened 2017-09-15
Add BackgroundService, a base class for long running HostedServices

davidfowl:
- Today the IHostedService pattern is a StartAsync/StopAsync pattern. Neither of these
methods are supposed to return a long running task that represents an execution. If
you wanted to have some logic run on a timer every 5 minutes, it's unnatural to do so
with simple async idioms. This base class implements IHostedService and exposes
a pattern where a long running async Task can be returned.
- Added tests

/cc @glennc

### DIFF
--- a/samples/GenericHostSample/MyServiceA.cs
+++ b/samples/GenericHostSample/MyServiceA.cs
@@ -5,38 +5,22 @@ using Microsoft.Extensions.Hosting;
 
 namespace GenericHostSample
 {
-    public class MyServiceA : IHostedService
+    public class MyServiceA : BackgroundService
     {
-        private bool _stopping;
-        private Task _backgroundTask;
-
-        public Task StartAsync(CancellationToken cancellationToken)
+        protected override async Task ExecuteAsync(CancellationToken cancellationToken)
         {
             Console.WriteLine("MyServiceA is starting.");
-            _backgroundTask = BackgroundTask();
-            return Task.CompletedTask;
-        }
 
-        private async Task BackgroundTask()
-        {
-            while (!_stopping)
+            cancellationToken.Register(() => Console.WriteLine("MyServiceA is stopping."));
+
+            while (!cancellationToken.IsCancellationRequested)
             {
-                await Task.Delay(TimeSpan.FromSeconds(5));
                 Console.WriteLine("MyServiceA is doing background work.");
+
+                await Task.Delay(TimeSpan.FromSeconds(5));
             }
 
             Console.WriteLine("MyServiceA background task is stopping.");
-        }
-
-        public async Task StopAsync(CancellationToken cancellationToken)
-        {
-            Console.WriteLine("MyServiceA is stopping.");
-            _stopping = true;
-            if (_backgroundTask != null)
-            {
-                // TODO: cancellation
-                await _backgroundTask;
-            }
         }
     }
 }
--- a/samples/GenericHostSample/MyServiceA.cs
+++ b/samples/GenericHostSample/MyServiceA.cs
@@ -17,7 +17,7 @@ namespace GenericHostSample
             {
                 Console.WriteLine("MyServiceA is doing background work.");
 
-                await Task.Delay(TimeSpan.FromSeconds(5));
+                await Task.Delay(TimeSpan.FromSeconds(5), stoppingToken);
             }
 
             Console.WriteLine("MyServiceA background task is stopping.");
--- a/samples/GenericHostSample/MyServiceA.cs
+++ b/samples/GenericHostSample/MyServiceA.cs
@@ -7,13 +7,13 @@ namespace GenericHostSample
 {
     public class MyServiceA : BackgroundService
     {
-        protected override async Task ExecuteAsync(CancellationToken cancellationToken)
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         {
             Console.WriteLine("MyServiceA is starting.");
 
-            cancellationToken.Register(() => Console.WriteLine("MyServiceA is stopping."));
+            stoppingToken.Register(() => Console.WriteLine("MyServiceA is stopping."));
 
-            while (!cancellationToken.IsCancellationRequested)
+            while (!stoppingToken.IsCancellationRequested)
             {
                 Console.WriteLine("MyServiceA is doing background work.");
 
--- a/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
@@ -12,16 +12,8 @@ namespace Microsoft.Extensions.Hosting
     /// </summary>
     public abstract class BackgroundService : IHostedService, IDisposable
     {
-        private static Action<object> _cancelToken = CancelToken;
-
         private Task _executingTask;
         private readonly CancellationTokenSource _stoppingCts = new CancellationTokenSource();
-        private readonly CancellationTokenSource _cancelledCts = new CancellationTokenSource();
-
-        public BackgroundService()
-        {
-            CancelledToken = _cancelledCts.Token;
-        }
 
         /// <summary>
         /// This method is called when the <see cref="IHostedService"/> starts. The implementation should return a task that represents
@@ -32,30 +24,22 @@ namespace Microsoft.Extensions.Hosting
         protected abstract Task ExecuteAsync(CancellationToken stoppingToken);
 
         /// <summary>
-        /// Triggered if cancellation is triggered during <see cref="IHostedService.StartAsync(CancellationToken)"/> or <see cref="IHostedService.StopAsync(CancellationToken)"/>.
-        /// </summary>
-        public CancellationToken CancelledToken { get; }
-
-        /// <summary>
         /// Triggered when the application host is ready to start the service.
         /// </summary>
         /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
         public virtual Task StartAsync(CancellationToken cancellationToken)
         {
-            using (cancellationToken.Register(_cancelToken, _cancelledCts))
+            // Store the task we're executing
+            _executingTask = ExecuteAsync(_stoppingCts.Token);
+
+            // If the task is completed then return it, this will bubble cancellation and failure to the caller
+            if (_executingTask.IsCompleted)
             {
-                // Store the task we're executing
-                _executingTask = ExecuteAsync(_stoppingCts.Token);
-
-                // If the task is completed then return it, this will bubble cancellation and failure to the caller
-                if (_executingTask.IsCompleted)
-                {
-                    return _executingTask;
-                }
-
-                // Otherwise it's running
-                return Task.CompletedTask;
+                return _executingTask;
             }
+
+            // Otherwise it's running
+            return Task.CompletedTask;
         }
 
         /// <summary>
@@ -70,30 +54,24 @@ namespace Microsoft.Extensions.Hosting
                 return;
             }
 
-            using (cancellationToken.Register(_cancelToken, _cancelledCts))
+            try
             {
-                try
-                {
-                    // Signal cancellation to the executing method
-                    _stoppingCts.Cancel();
-                }
-                finally
-                {
-                    // Wait until the task completes or the stop token triggers
-                    await Task.WhenAny(_executingTask, Task.Delay(-1, cancellationToken));
-                }
+                // Signal cancellation to the executing method
+                _stoppingCts.Cancel();
             }
+            finally
+            {
+                // Wait until the task completes or the stop token triggers
+                await Task.WhenAny(_executingTask, Task.Delay(-1, cancellationToken));
+            }
+
         }
 
         public virtual void Dispose()
         {
             _stoppingCts.Cancel();
-            _cancelledCts.Cancel();
 
             _stoppingCts.Dispose();
-            _cancelledCts.Dispose();
         }
-
-        private static void CancelToken(object state) => ((CancellationTokenSource)state).Cancel();
     }
 }
--- a/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
@@ -9,41 +9,53 @@ namespace Microsoft.Extensions.Hosting
     /// <summary>
     /// Base class for implementing a long running <see cref="IHostedService"/>.
     /// </summary>
-    public abstract class BackgroundService : IHostedService
+    public abstract class BackgroundService : IHostedService, IDisposable
     {
         private Task _executingTask;
-        private CancellationTokenSource _cts;
+        private CancellationTokenSource _executionCts;
+        private readonly CancellationTokenSource _startCts = new CancellationTokenSource();
+        private readonly CancellationTokenSource _stopCts = new CancellationTokenSource();
 
         /// <summary>
         /// This method is called when the <see cref="IHostedService"/> starts. The implementation should return a task that represents
         /// the lifetime of the long running operation(s) being performed.
         /// </summary>
-        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that fires when the <see cref="IHostedService"/> is stopped or cancelled.</param>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that fires when the <see cref="IHostedService"/> is stopped.</param>
         /// <returns>A <see cref="Task"/> that represents the long running operations.</returns>
         protected abstract Task ExecuteAsync(CancellationToken cancellationToken);
 
-        public Task StartAsync(CancellationToken cancellationToken)
+        /// <summary>
+        /// A <see cref="CancellationToken"/> that represents the <see cref="IHostedService.StartAsync(CancellationToken)"/> call.
+        /// </summary>
+        public CancellationToken StartToken => _startCts.Token;
+
+        /// <summary>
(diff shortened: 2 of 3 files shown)


Tratcher:
Update the samples that do exactly this?
https://github.com/aspnet/Hosting/blob/1c3fa82908fe2cb773626b6613843213286a482b/samples/GenericHostSample/MyServiceA.cs

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that fires when the <see cref="IHostedService"/> is stopped or cancelled.</param>
+        /// <returns>A <see cref="Task"/> that represents the long running operations.</returns>
+        protected abstract Task ExecuteAsync(CancellationToken cancellationToken);
+
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            // If the token is cancellable, create a linked token so we can trigger cancellation 
+            // on shutdown and based on the incoming token
+            _cts = cancellationToken.CanBeCanceled ?
+                CancellationTokenSource.CreateLinkedTokenSource(cancellationToken) :
```
Tratcher: Linking these is odd, their lifetimes are completely different. I understand why you have both, but maybe you should pass them in separately to ExecuteAsync.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,66 @@
...
+                new CancellationTokenSource();
+
+            // Store the task we're executing
+            _executingTask = ExecuteAsync(_cts.Token);
```
Tratcher: This doesn't honor the spirit of StartAsync, it should block until the service has successfully started. This pattern readily doesn't differentiate between starting and running. The cancellation tokens share share the same confusion.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            // If the task is completed then return it, this will bubble cancellation and failure to the caller
+            if (_executingTask.IsFaulted)
```
Tratcher: or IsCancelled, they're different.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            try
+            {
+                // Signal cancellation to the executing method
+                _cts.Cancel();
+            }
+            finally
+            {
+                // Wait until the task completes or the stop token triggers
+                await Task.WhenAny(_executingTask, Task.Delay(-1, cancellationToken));
```
Tratcher: This mechanic does not distinguish between a graceful and abortive shutdown of the service, it just gives up waiting for it to shutdown. This is another case where it seems like you need to expose both tokens directly to the consumer.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: I originally had that but I didn't want to bubble cancellation to the caller. Meh I guess it's harmless.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: @Tratcher passing two tokens to Execute async would be insane. What would the code do besides link them? Either we're saying that you cant cancel Startup of the service or we can.

Either way passing 2 tokens is a bad design.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: There's no easy way to indicate that other than introducing some other control flow mechanic (Like a callback). Is it really worth it? It's similar to the 2 token thing.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: I removed that because there's no way to tell if the executing task was cancelled because of stopping or because the token fired. I can treat timeouts as abortive.

davidfowl:
🆙 📅

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,88 @@
...
+        private readonly CancellationTokenSource _startCts = new CancellationTokenSource();
+        private readonly CancellationTokenSource _stopCts = new CancellationTokenSource();
+
+        /// <summary>
+        /// This method is called when the <see cref="IHostedService"/> starts. The implementation should return a task that represents
+        /// the lifetime of the long running operation(s) being performed.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that fires when the <see cref="IHostedService"/> is stopped.</param>
+        /// <returns>A <see cref="Task"/> that represents the long running operations.</returns>
+        protected abstract Task ExecuteAsync(CancellationToken cancellationToken);
```
Tratcher: This CT is non-traditional because it indicates a request for graceful shutdown rather than abortive. Also move to an appropriately named property?

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        /// <summary>
+        /// A <see cref="CancellationToken"/> that represents the <see cref="IHostedService.StartAsync(CancellationToken)"/> call.
+        /// </summary>
+        public CancellationToken StartToken => _startCts.Token;
```
Tratcher: Single AbortToken

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// the lifetime of the long running operation(s) being performed.
+        /// </summary>
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that fires when the <see cref="IHostedService"/> is stopped.</param>
+        /// <returns>A <see cref="Task"/> that represents the long running operations.</returns>
+        protected abstract Task ExecuteAsync(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// A <see cref="CancellationToken"/> that repesents the ungraceful shutdown.
+        /// </summary>
+        public CancellationToken ShutdownToken => _shutdownCts.Token;
```
Tratcher: AbortToken?

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    /// <summary>
+    /// Base class for implementing a long running <see cref="IHostedService"/>.
+    /// </summary>
+    public abstract class BackgroundService : IHostedService, IDisposable
+    {
+        private static Action<object> _cancelToken = CancelToken;
+
+        private Task _executingTask;
+        private CancellationTokenSource _executionCts;
```
Tratcher: why initialize one CTS and not the other?

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,85 @@
...
+                {
+                    // Wait until the task completes or the stop token triggers
+                    await Task.WhenAny(_executingTask, Task.Delay(-1, cancellationToken));
+                }
+            }
+        }
+
+        public virtual void Dispose()
+        {
+            _shutdownCts.Dispose();
```
Tratcher: Abort on Dispose? Otherwise nothing signals the background task.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: I'd prefer if it was part of the method because it's the main use case, "tell me when I'm being stopped". Everything else is edge casey. As for naming, we use tokens to represent all sorts of things now even things that aren't abortive.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: I could do both I guess.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
glennc: I'm not completely clear how this can be used. When would this fire? If it is an ungraceful shutdown how can anyone have time to react to it? Otherwise it would be graceful. Would it not?

Tratcher: Consider the WebHost 5 second shutdown timeout we have today. StopAsync is graceful until that timeout expires, then it becomes ungracefull and kestrel starts killing sockets.

Tratcher: The process still waits for Kestrel to finish before shutting down, it just asks Kestrel to hurry it up.

glennc: So it is more like a Stopping/Stop scenario where ShutdownToken is stop and the other graceful one is, we are stopping start cleaning up?

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,85 @@
+﻿using System;
```
muratg: Copyright notice.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <param name="cancellationToken">A <see cref="CancellationToken"/> that fires when the <see cref="IHostedService"/> is stopped.</param>
+        /// <returns>A <see cref="Task"/> that represents the long running operations.</returns>
+        protected abstract Task ExecuteAsync(CancellationToken cancellationToken);
+
+        /// <summary>
+        /// A <see cref="CancellationToken"/> that repesents the ungraceful shutdown.
+        /// </summary>
+        public CancellationToken ShutdownToken => _shutdownCts.Token;
+
+        public virtual Task StartAsync(CancellationToken cancellationToken)
```
muratg: Do we null check `CancellationToken`s?

Tratcher: No, they're structs.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
Tratcher: Yes.

davidfowl: It's really not going to be used much. This for the 0.000001% case that you want to handle the shutdown timeout after being signalled for stopping in the first place. I'm leaving it as it is.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: What do you mean? What combination of things would cause that to matter?

davidfowl:
@Tratcher ok, updated, we can argue about names for the next year. I'm going to merge it now. I added a test for StartAsync -> Dispose.

davidfowl:
After playing with this I think it's better for implementors to override StartAsync and StopAsync to handle ungraceful shutdown. I can't think of a scenario where the usage would be straightforward usage that doesn't include combining both tokens.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// the lifetime of the long running operation(s) being performed.
+        /// </summary>
+        /// <param name="stoppingToken">Triggered when <see cref="IHostedService.StopAsync(CancellationToken)"/> is called.</param>
+        /// <returns>A <see cref="Task"/> that represents the long running operations.</returns>
+        protected abstract Task ExecuteAsync(CancellationToken stoppingToken);
+
+        /// <summary>
+        /// Triggered if cancellation is triggered during <see cref="IHostedService.StartAsync(CancellationToken)"/> or <see cref="IHostedService.StopAsync(CancellationToken)"/>.
+        /// </summary>
+        public CancellationToken CancelledToken { get; }
```
Tratcher: OnCancelled

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                try
+                {
+                    // Signal cancellation to the executing method
+                    _stoppingCts.Cancel();
+                }
+                finally
+                {
+                    // Wait until the task completes or the stop token triggers
+                    await Task.WhenAny(_executingTask, Task.Delay(-1, cancellationToken));
```
Tratcher: Timeout.Infinate

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            using (cancellationToken.Register(_cancelToken, _cancelledCts))
+            {
+                try
+                {
+                    // Signal cancellation to the executing method
+                    _stoppingCts.Cancel();
+                }
+                finally
```
Tratcher: Why is this a finally block? If this throws why wait for _executingTask? If _executingTask throws then you loose the original exception.

samples/GenericHostSample/MyServiceA.cs
```diff
@@ -5,38 +5,22 @@
...
-        {
-            while (!_stopping)
+            stoppingToken.Register(() => Console.WriteLine("MyServiceA is stopping."));
+
+            while (!stoppingToken.IsCancellationRequested)
             {
-                await Task.Delay(TimeSpan.FromSeconds(5));
                 Console.WriteLine("MyServiceA is doing background work.");
+
+                await Task.Delay(TimeSpan.FromSeconds(5));
```
Tratcher: Include the OnCancelled token.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,99 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Hosting
+{
+    /// <summary>
+    /// Base class for implementing a long running <see cref="IHostedService"/>.
+    /// </summary>
+    public abstract class BackgroundService : IHostedService, IDisposable
```
Tratcher: Should there be a lambda implementation of this? E.g. `webHostBuilder/services.AddBackgroundService(stop => { ... })`

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
Tratcher: Timeout.Infinate.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,77 @@
...
+                await Task.WhenAny(_executingTask, Task.Delay(-1, cancellationToken));
+            }
+
+        }
+
+        public virtual void Dispose()
+        {
+            _stoppingCts.Cancel();
+
+            _stoppingCts.Dispose();
```
Tratcher: Don't dispose cancelled tokens, it's redundant and creates a race condition for people registering with the token.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
davidfowl: Eventually yes.

src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -0,0 +1,75 @@
...
+        protected abstract Task ExecuteAsync(CancellationToken stoppingToken);
+
+        /// <summary>
+        /// Triggered when the application host is ready to start the service.
+        /// </summary>
+        /// <param name="cancellationToken">Indicates that the start process has been aborted.</param>
+        public virtual Task StartAsync(CancellationToken cancellationToken)
+        {
+            // Store the task we're executing
+            _executingTask = ExecuteAsync(_stoppingCts.Token);
```
khellang: @davidfowl Does it make sense to start these tasks using `TaskCreationOptions.LongRunning`?

davidfowl: I don’t know if that makes sense but it might align with their expectations. They don’t get a dedicated thread to block, this is async. If we have them a long running task I’d rather have the method be synchronous. That way it’s clear that if you go async you’ll lose the current thread.

khellang: Oh, and there's also the problem of synchronization context and deadlocks 😢

Issue #1219, opened 2017-09-16
Add extension methods to HostBuilder to Configure things without the HostContext


davidfowl:
Right now all of the lambdas have 2 args, the host context and the other arg. There should be an overload that omits the host context.

PR #1220, opened 2017-09-17
Added overloads without the HostingContext

davidfowl:
Fixes https://github.com/aspnet/Hosting/issues/1219

### DIFF
--- a/src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
+++ b/src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
@@ -62,6 +62,53 @@ namespace Microsoft.Extensions.Hosting
         }
 
         /// <summary>
+        /// Adds a delegate for configuring the provided <see cref="ILoggingBuilder"/>. This may be called multiple times.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IHostBuilder" /> to configure.</param>
+        /// <param name="configureLogging">The delegate that configures the <see cref="ILoggingBuilder"/>.</param>
+        /// <returns>The same instance of the <see cref="IHostBuilder"/> for chaining.</returns>
+        public static IHostBuilder ConfigureLogging(this IHostBuilder hostBuilder, Action<ILoggingBuilder> configureLogging)
+        {
+            return hostBuilder.ConfigureServices((context, collection) => collection.AddLogging(builder => configureLogging(context, builder)));
+        }
+        /// <summary>
+        /// Sets up the configuration for the remainder of the build process and application. This can be called multiple times and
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
```diff
@@ -62,6 +62,53 @@ public static IHostBuilder ConfigureLogging(this IHostBuilder hostBuilder, Actio
...
+        }
+        /// <summary>
+        /// Sets up the configuration for the remainder of the build process and application. This can be called multiple times and
+        /// the results will be additive. The results will be available at <see cref="HostBuilderContext.Configuration"/> for
+        /// subsequent operations, as well as in <see cref="IHost.Services"/>.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IHostBuilder" /> to configure.</param>
+        /// <param name="configureDelegate"></param>
+        /// <returns>The same instance of the <see cref="IHostBuilder"/> for chaining.</returns>
+        public static IHostBuilder ConfigureAppConfiguration(this IHostBuilder hostBuilder, Action<IConfigurationBuilder> configureDelegate)
```
Tratcher: These are generic enough they can move to Ext.Hosting.Abstractions.

src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
```diff
@@ -62,6 +62,53 @@ public static IHostBuilder ConfigureLogging(this IHostBuilder hostBuilder, Actio
...
+            return hostBuilder.ConfigureAppConfiguration((context, builder) => configureDelegate(builder));
+        }
+
+        /// <summary>
+        /// Adds services to the container. This can be called multiple times and the results will be additive.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IHostBuilder" /> to configure.</param>
+        /// <param name="configureDelegate"></param>
+        /// <returns>The same instance of the <see cref="IHostBuilder"/> for chaining.</returns>
+        public static IHostBuilder ConfigureServices(this IHostBuilder hostBuilder, Action<IServiceCollection> configureDelegate)
```
Tratcher: Might as well use these in a few of the tests.

src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
davidfowl: I didn't move them because all can't be moved without adding more deps.

Tratcher: I didn't say all, just most of your new ones.

PR #1222, opened 2017-09-18
Misc generic host pr feedback

Tratcher:
@JunTaoLuo following up on your comments from the prior PR.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,7 +10,6 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,6 +10,7 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
 </Project>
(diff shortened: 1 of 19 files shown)


src/Microsoft.AspNetCore.Hosting.Abstractions/IWebHost.cs
```diff
@@ -31,13 +31,13 @@ public interface IWebHost : IDisposable
         /// <summary>
         /// Starts listening on the configured addresses.
         /// </summary>
-        Task StartAsync(CancellationToken cancellationToken = default(CancellationToken));
+        Task StartAsync(CancellationToken cancellationToken = default);
```
davidfowl: What's going on here?

src/Microsoft.AspNetCore.Hosting/Microsoft.AspNetCore.Hosting.csproj
```diff
@@ -6,6 +6,7 @@
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;hosting</PackageTags>
+    <LangVersion>latest</LangVersion>
```
davidfowl: This should be done solution wide @natemcmaster any ideas?

src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.Extensions.Hosting
     public static class HostingHostBuilderExtensions
     {
         /// <summary>
-        /// Specify the environment to be used by the web host.
+        /// Specify the environment to be used by the host.
```
JunTaoLuo: There are a few instances at HostingAbstractionsHostExtensions.cs IHostingEnvironment.cs

src/Microsoft.AspNetCore.Hosting/Microsoft.AspNetCore.Hosting.csproj
natemcmaster: > This should be done solution wide @natemcmaster any ideas?

Add it to Directory.Build.props

Directory.Build.props
```diff
@@ -10,6 +10,7 @@
     <PublicSign Condition="'$(OS)' != 'Windows_NT'">true</PublicSign>
     <VersionSuffix Condition="'$(VersionSuffix)'!='' AND '$(BuildNumber)' != ''">$(VersionSuffix)-$(BuildNumber)</VersionSuffix>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <LangVersion>latest</LangVersion>
```
davidfowl: /cc @Eilon how do you feel about using the latest C# everywhere? Maybe this should be latest minor since we don't want to use C# 8 until it's done

Tratcher: @davidfowl The settings go the other way. Latest major (`default`) is like LTS and latest minor (`latest`) is like FTS.  However, we could explicitly say `7.1`.

Eilon: Does `latest` mean "Latest RTM version" or "actual latest experimental crazy town insane features"?

PR #1223, opened 2017-09-19
Schedule ExecuteAsync instead of running it inline

davidfowl:
- This prevents bad actors from blocking start up but also comes
with a bunch of caveats. StartAsync merely schedules the ExecuteAsync call
so there's no guarantee that ExecuteAsync runs before StartAsync or Dispose
(in theory).

I'm not sure I like scheduling the work. I like the guarantee that when StartAsync returns, we know the ExecuteAsync delegate has started.

### DIFF
--- a/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
@@ -12,8 +12,13 @@ namespace Microsoft.Extensions.Hosting
     /// </summary>
     public abstract class BackgroundService : IHostedService, IDisposable
     {
+        private static readonly Func<object, Task> _executeBackgroundTask = ExecuteBackgroundTaskAsync;
+
         private Task _executingTask;
         private readonly CancellationTokenSource _stoppingCts = new CancellationTokenSource();
+
+        // For testing purposes only
+        internal Task ExecutingTask => _executingTask;
 
         /// <summary>
         /// This method is called when the <see cref="IHostedService"/> starts. The implementation should return a task that represents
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
```diff
@@ -62,14 +64,34 @@ public virtual Task StartAsync(CancellationToken cancellationToken)
             finally
             {
                 // Wait until the task completes or the stop token triggers
-                await Task.WhenAny(_executingTask, Task.Delay(Timeout.Infinite, cancellationToken));
-            }
+                var task = await Task.WhenAny(_executingTask, Task.Delay(Timeout.Infinite, cancellationToken));
```
Tratcher: You still haven't given a good reason for this to be in a finally block.

davidfowl: It's because throwing on Cancel doesn't mean the task itself has unwound. Cancellation happens on a different thread and the executing task is running on.

Tratcher: You already abandon the task if the other CT trips. How is that much different from the shutdown CT throwing? May may also end up masking an exception.

davidfowl: Its a best effort. The callback might throw because it was handed off to another component that throw not because the background task itself throws. The goal was to avoid just leaving a dangling task. It's a best effort.

> You may also end up masking an exception.

Masking it how?

Tratcher: By masking I mean that Cancel can throw and then _executingTask can throw and you lose the exception from Cancel.

davidfowl: Yea, I get that, if we were being pedantic we could throw an aggregate exception but I dunno if it matters

glennc:
Should there be a token I can wait on after calling start to know that everything has started? I am concerned about not knowing if anything has actually started. I am thinking of something that sends a signal after starting to say it can now handle processing, how do I know when to send that signal?

davidfowl:
Closing this, it doesn't feel good

Issue #1214, opened 2017-09-15
 Microsoft.AspNetCore.Hosting.FunctionalTests.ShutdownTests.ShutdownTestWaitForShutdown fails on linux


davidfowl:
```
[xUnit.net 00:00:44.3372819]         | TestLifetime Information: Finished test ShutdownTestWaitForShutdown in 43.2488036s
Failed   Microsoft.AspNetCore.Hosting.FunctionalTests.ShutdownTests.ShutdownTestWaitForShutdown
Error Message:
 System.Exception : dotnet publish exited with exit code : 1
Stack Trace:
   at Microsoft.AspNetCore.Server.IntegrationTesting.ApplicationDeployer.DotnetPublish(String publishRoot) in /home/travis/build/aspnet/Hosting/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs:line 49
   at Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer.<DeployAsync>d__7.MoveNext() in /home/travis/build/aspnet/Hosting/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs:line 41
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Hosting.FunctionalTests.ShutdownTests.<ShutdownTestWaitForShutdown>d__2.MoveNext() in /home/travis/build/aspnet/Hosting/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs:line 99
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

PR #1224, opened 2017-09-19
Flaky hosting functional tests

### LINKED ISSUE #1213: InjectedStartup_DefaultApplicationNameIsEntryAssembly test broken on linux
Travis keeps failing with here.

```
InjectedStartup_DefaultApplicationNameIsEntryAssembly in 52.0073982s
Failed   Microsoft.AspNetCore.Hosting.FunctionalTests.WebHostBuilderTests.InjectedStartup_DefaultApplicationNameIsEntryAssembly_CoreClr
Error Message:
 Assert.Equal() Failure
          ↓ (pos 0)
Expected: IStartupInjectionAssemblyName
Actual:   Now listening on: http://localhost:5000
          ↑ (pos 0)
```

@JunTaoLuo my theory is that this commit https://github.com/aspnet/Hosting/commit/b187c1adc4db343ad985e3d45cfb4f77180966e6 broke things. I haven't looked into it much though.

JunTaoLuo:
Fixes #1213. The fix is to filter out application startup messsages so tests only examine the application output.

### DIFF
--- a/test/Microsoft.AspNetCore.Hosting.FunctionalTests/Properties/AssemblyInfo.cs
+++ b/test/Microsoft.AspNetCore.Hosting.FunctionalTests/Properties/AssemblyInfo.cs
@@ -1,0 +1,6 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly)]
--- a/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
+++ b/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
@@ -1,9 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.IntegrationTesting;
 using Microsoft.AspNetCore.Testing;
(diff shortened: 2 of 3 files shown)


test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
```diff
@@ -16,6 +17,9 @@ namespace Microsoft.AspNetCore.Hosting.FunctionalTests
 {
     public class ShutdownTests : LoggedTest
     {
+        private static readonly Regex NowListeningRegex = new Regex(@"^\s*Now listening on: (?<url>.*)$");
```
JunTaoLuo: We could either make this public in SelfHostDeployer (which is an odd thing to expose) or make this internal visible to this test.

JunTaoLuo:
@davidfowl Will file a followup issue to optionally disable startup messages in Hosting itself.

JunTaoLuo:
Also fixes #1214

JunTaoLuo:
Tested 4 times in a row without any failures. I think the fixes are ready to be merged.

test/Microsoft.AspNetCore.Hosting.FunctionalTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
Tratcher: https://github.com/aspnet/IISIntegration/blob/1297798546b6aed8f591d1e1548a4374bc5b6613/test/IISIntegration.FunctionalTests/Properties/AssemblyInfo.cs#L3

anurse:
Would it be more stable to simply get a collection of lines written and check if the desired lines are present rather than checking that a specific exact string was written? Basically do a series `.Contains` checks on a list of lines at the end of the test. That way you'd still be able to verify the messages you need but disregard the ones that aren't important.

JunTaoLuo:
@anurse Yea I tried that out in an previous attempt, see https://github.com/aspnet/Hosting/compare/johluo/old-ci-flaky-travis?expand=1. I found that a little awkward too since I needed to determine when the application is done. I ended up just counting the messages until the expected number were received. Either approach would work but I felt this one is a little more explicit in terms of what messages we were filtering out.

JunTaoLuo:
@davidfowl filed followup in https://github.com/aspnet/Hosting/issues/1225

anurse:
Makes me wonder if we need some kind of "API mode" for all our console output (including logs) that returns consistent well-defined data structures via stdout... Anyway, that's out of scope here.

Issue #1225, opened 2017-09-20
Add option to disable hosting startup messages


JunTaoLuo:
Currently there is no way to disable the startup message:
```
Now listening on: http://localhost:5000
Application started. Press Ctrl+C to shut down.
```
We can add an option to disable writing these messages to console output. We will also need to add an option to SelfHostDeployer to stop scraping startup messages to determine the application has started and parse the url the app is listening on.

As a side effect, we can simplify the filtering in a few functional tests. https://github.com/aspnet/Hosting/blob/e53abdb2b8c95d71281f719b5967067191a7cc65/test/Microsoft.AspNetCore.Hosting.FunctionalTests/WebHostBuilderTests.cs#L62-L64 
Also, if it's possible to disable startup messages, can we print a warning for duplicate hosting startup assemblies?

Tratcher:
https://github.com/aspnet/Hosting/blob/e53abdb2b8c95d71281f719b5967067191a7cc65/src/Microsoft.Extensions.Hosting/ConsoleLifetimeOptions.cs#L12

Issue #1042, opened 2017-04-25
Expose WebHostOptions as a first class way to configure the IWebHostBuilder


davidfowl:
Today we have special extension methods that configure settings on the `IWebHostBuilder` with various options, `PreferHostingUrls`, `CaptureStartupErrors` etc. Instead, we should expose `WebHostOptions` and a method `ConfigureOptions(Action<WebHostOptions>)` that lets you programmatically set configuration. This win over an configuration based logic.

JosephWoodward:
This would be awesome 👍 Can think of a lot of great use cases for such an extension.

muratg:
@davidfowl If we want this in 2.0.0, it needs to be designed this week.

muratg:
@davidfowl will do the investigation

muratg:
cc @glennc

davidfowl:
This is low value so I'm going to close this.

PR #1232, opened 2017-09-28
Add option to enable/disable writing startup messages

JunTaoLuo:
#1225


Tratcher:
Or we put this on WebHostOptions like everything else and add that to the service container. We already add other POCOs like HostBuilderContext.

JunTaoLuo:
I started with that approach, remind me again why we concluded not to do that? Something about purity and it's an option for running the host? Also, that would expose other web host options to be configured via ConfigureServices but none of those changes would really affect how webhostbuilder builds.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -151,26 +152,36 @@ public IWebHost Build()
...
+            var hostingServiceProvider = hostingServices.BuildServiceProvider();
+            
+            if (_options.WriteStatusMessages)
             {
-                Console.WriteLine("The environment variable 'Hosting:Environment' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
-            }
+                // Warn about deprecated environment variables
+                if (Environment.GetEnvironmentVariable("Hosting:Environment") != null)
+                {
+                    Console.WriteLine("The environment variable 'Hosting:Environment' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
```
JunTaoLuo: Do we intend to ever remove these fallbacks? Also, should we log this instead of writing to the console?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -151,26 +152,36 @@ public IWebHost Build()
...
+            var hostingServiceProvider = hostingServices.BuildServiceProvider();
 
-            if (Environment.GetEnvironmentVariable("ASPNET_ENV") != null)
+            if (_options.WriteStatusMessages)
             {
-                Console.WriteLine("The environment variable 'ASPNET_ENV' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
+                // Warn about deprecated environment variables
+                if (Environment.GetEnvironmentVariable("Hosting:Environment") != null)
+                {
+                    Console.WriteLine("The environment variable 'Hosting:Environment' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
```
JunTaoLuo: Do we intend to ever remove these fallbacks? Also, should we log this instead of writing to the console?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -151,26 +152,36 @@ public IWebHost Build()
...
+                }
             }
 
-            if (Environment.GetEnvironmentVariable("ASPNETCORE_SERVER.URLS") != null)
+            var logger = hostingServiceProvider.GetRequiredService<ILogger<WebHost>>();
+            // Warn about duplicate HostingStartupAssemblies
+            foreach (var assemblyName in _options.HostingStartupAssemblies.GroupBy(a => a, StringComparer.OrdinalIgnoreCase).Where(g => g.Count() > 1))
             {
-                Console.WriteLine("The environment variable 'ASPNETCORE_SERVER.URLS' is obsolete and has been replaced with 'ASPNETCORE_URLS'");
+                logger.LogWarning($"The assembly {assemblyName} was specified multiple times. Hosting startup assemblies should only be specified once.");
```
jkotalik: See that we are now explicitly logging multiple startup assemblies with the same name. I know we decided to silently ignore before.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -151,26 +152,36 @@ public IWebHost Build()
...
+                }
+
+                if (Environment.GetEnvironmentVariable("ASPNETCORE_SERVER.URLS") != null)
+                {
+                    Console.WriteLine("The environment variable 'ASPNETCORE_SERVER.URLS' is obsolete and has been replaced with 'ASPNETCORE_URLS'");
+                }
             }
 
-            if (Environment.GetEnvironmentVariable("ASPNETCORE_SERVER.URLS") != null)
+            var logger = hostingServiceProvider.GetRequiredService<ILogger<WebHost>>();
```
jkotalik: For my own knowledge, should we be getting the required logger here? I thought the whole point of Console.Writeline was to avoid pulling in the logger.

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
```diff
@@ -30,6 +30,12 @@ public WebHostOptions(IConfiguration configuration, string applicationNameFallba
             WebRoot = configuration[WebHostDefaults.WebRootKey];
             ContentRootPath = configuration[WebHostDefaults.ContentRootKey];
             PreventHostingStartup = WebHostUtilities.ParseBool(configuration, WebHostDefaults.PreventHostingStartupKey);
+
+            if (!string.IsNullOrEmpty(configuration[WebHostDefaults.WriteStatusMessages]))
```
jkotalik: Just to be sure, is PreventHostingStartup a required part of configuration? Just trying to understand why we don't do the same check for it.

test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs
```diff
@@ -16,12 +15,7 @@ namespace Microsoft.AspNetCore.Hosting.FunctionalTests
 {
     public class ShutdownTests : LoggedTest
     {
-        private static readonly Regex NowListeningRegex = new Regex(@"^\s*Now listening on: (?<url>.*)$");
-        private const string ApplicationStartedMessage = "Application started. Press Ctrl+C to shut down.";
-
-        public ShutdownTests(ITestOutputHelper output) : base(output)
-        {
-        }
+        public ShutdownTests(ITestOutputHelper output) : base(output) { }
```
jkotalik: Writing a unit test would be painful because we would need to attach to stdout, right?

src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
```diff
@@ -160,6 +160,17 @@ public static IWebHostBuilder PreferHostingUrls(this IWebHostBuilder hostBuilder
         }
 
         /// <summary>
+        /// TODO:
```
jkotalik: Make sure you get that.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -151,26 +152,36 @@ public IWebHost Build()
             }
             _webHostBuilt = true;
 
-            // Warn about deprecated environment variables
-            if (Environment.GetEnvironmentVariable("Hosting:Environment") != null)
-            {
-                Console.WriteLine("The environment variable 'Hosting:Environment' is obsolete and has been replaced with 'ASPNETCORE_ENVIRONMENT'");
-            }
+            var hostingServices = BuildCommonServices(out var hostingStartupErrors);
```
jkotalik: May should note that this ordering changed to guarantee _options is set.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
jkotalik: Nvrm, read feedback above.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
jkotalik: We should probably go all or nothing here, either using Console.Writeline or the logger.

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
Tratcher: Is it because ParseBool returns false if the value is missing and would override the defaults? Some claim all bool defaults should be false anyways, and you name the property accordingly. SuppressConsoleMessages?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
Tratcher: Let's just remove these, I haven't seen a lot of trouble with then recently.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -76,7 +77,7 @@ public static void Run(this IWebHost host)
             var done = new ManualResetEventSlim(false);
             using (var cts = new CancellationTokenSource())
             {
-                AttachCtrlcSigtermShutdown(cts, done, shutdownMessage: "Application is shutting down...");
+                AttachCtrlcSigtermShutdown(cts, done, shutdownMessage: host.Services.GetService<WebHostOptions>().WriteStatusMessages ? "Application is shutting down..." : string.Empty);
```
Tratcher: GetRequiredService
break this line up / move the message definition to a local var.

src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
```diff
@@ -91,22 +92,27 @@ public static void Run(this IWebHost host)
 
                 var hostingEnvironment = host.Services.GetService<IHostingEnvironment>();
                 var applicationLifetime = host.Services.GetService<IApplicationLifetime>();
+                var options = host.Services.GetService<WebHostOptions>();
```
Tratcher: GetRequiredService

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: Remove the fallbacks or just remove the messages?

JunTaoLuo: Filed https://github.com/aspnet/Hosting/issues/1233 for removing the fallbacks in the next major release.

src/Microsoft.AspNetCore.Hosting/Internal/WebHostOptions.cs
JunTaoLuo: Yea I thought about that but I supposed at some point, we might consolidate the web host and generic host. It would be rather odd that one has the option to suppress console messages while the other enables it. We could make both suppress? Or we can just live with the discrepancy.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: Nah, I don't think we need to doc each minor change like an update for the order of operations. That's what commit histories are for.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
JunTaoLuo: Well the reason we didn't do this before is that we didn't want to write to console. I think we should have a warning somewhere since odering of startup assemblies can matter and having duplicates is often just wrong. However it is not a major issue and I haven't seen anybody complaining about it so I guess I'm content if we decide to not even log a warning that can be disabled.

jkotalik: I have no preference. Just want to make sure everyone is on the same page.

JunTaoLuo:
:up::date:

src/Microsoft.AspNetCore.Hosting.Abstractions/WebHostDefaults.cs
```diff
@@ -17,6 +17,7 @@ public static class WebHostDefaults
         public static readonly string ContentRootKey = "contentRoot";
         public static readonly string PreferHostingUrlsKey = "preferHostingUrls";
         public static readonly string PreventHostingStartupKey = "preventHostingStartup";
+        public static readonly string WriteStatusMessages = "writeStatusMessages";
```
Tratcher: The others are all suffixed with 'Key'.

src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs
```diff
@@ -160,6 +160,17 @@ public static IWebHostBuilder PreferHostingUrls(this IWebHostBuilder hostBuilder
         }
 
         /// <summary>
+        /// Specify if startup status messages should be written to console.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="writeStatusMessages"><c>true</c> to enable writing of hosting startup status messages; otherwise <c>false</c>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder WriteStatusMessages(this IWebHostBuilder hostBuilder, bool writeStatusMessages)
```
davidfowl: Ugh, does this really need to be first class?

Tratcher: This method is optional, but several users have asked for it, it's not just a test feature.

davidfowl: I know, I read the issues just like you, I *really* think we should avoid adding *more* of these methods and instead do https://github.com/aspnet/Hosting/issues/1042. It's much clearer and more discoverable.

JunTaoLuo: Spoke offline and this seems out of scope. We can explore alternatives separately.

PR #1236, opened 2017-10-11
Remove the hosting dependency on WebHostBuilderFactory shared sources

javiercn:
We need to do this because https://github.com/aspnet/EntityFrameworkCore/tree/f4c9f4c4be0e6f159506e583eda0ba68536356b2/src/EFCore.Design does not have a dependency in hosting and I don't want to introduce it.

This change does not affect any public API.


Tratcher:
So how do they run it without the types?

javiercn:
@Tratcher They do everything through reflection. As the tool runs in the context of your app, the types are there at runtime. See https://github.com/aspnet/EntityFrameworkCore/blob/f4c9f4c4be0e6f159506e583eda0ba68536356b2/src/EFCore.Design/Design/Internal/AppServiceProviderFactory.cs#L76 for more details

Issue #1208, opened 2017-09-14
Should the generic host have a default timeout for StopAsync?


Tratcher:
IWebHost.StopAsync has a default 5 second timeout for graceful shutdown. This was not included in the initial IHost.StopAsync design. Should it be added?

Tratcher:
https://github.com/aspnet/Hosting/pull/1204

PR #1237, opened 2017-10-13
Default timeout for IHost.StopAsync. Create Host with DI.

Tratcher:
#1208 This ports the default shutdown timeout from WebHost to Host. It also introduces HostOptions for configuring the value, and changes Host to be instantiated via DI.

I recommend the https://github.com/aspnet/Hosting/pull/1237/files?w=1 view when reviewing Host.cs

### DIFF
--- a/src/Microsoft.Extensions.Hosting/HostBuilder.cs
+++ b/src/Microsoft.Extensions.Hosting/HostBuilder.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Extensions.Hosting
             BuildAppConfiguration();
             CreateServiceProvider();
 
-            return new Host(_appServices);
+            return _appServices.GetRequiredService<IHost>();
         }
 
         private void BuildHostConfiguration()
@@ -179,6 +179,7 @@ namespace Microsoft.Extensions.Hosting
             services.AddSingleton(_appConfiguration);
             services.AddSingleton<IApplicationLifetime, ApplicationLifetime>();
             services.AddSingleton<IHostLifetime, ProcessLifetime>();
+            services.AddSingleton<IHost, Host>();
             services.AddOptions();
             services.AddLogging();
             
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Hosting/HostBuilder.cs
```diff
@@ -179,6 +179,7 @@ private void CreateServiceProvider()
             services.AddSingleton(_appConfiguration);
             services.AddSingleton<IApplicationLifetime, ApplicationLifetime>();
             services.AddSingleton<IHostLifetime, ProcessLifetime>();
+            services.AddSingleton<IHost, Host>();
```
muratg: Don't these need to be after AddOptions?

Tratcher: It's not order sensitive.

src/Microsoft.Extensions.Hosting/Internal/Host.cs
```diff
@@ -8,22 +8,31 @@
...
+        private readonly IHostLifetime _hostLifetime;
+        private readonly ApplicationLifetime _applicationLifetime;
+        private readonly HostOptions _options;
         private IEnumerable<IHostedService> _hostedServices;
 
-        internal Host(IServiceProvider services)
+        public Host(IServiceProvider services, IApplicationLifetime applicationLifetime, ILogger<Host> logger,
+            IHostLifetime hostLifetime, IOptions<HostOptions> options)
         {
+            if (options == null)
```
halter73: Nit: Might be worth guarding agains and NRE in `StopAsync` by ensuring `options.Value` isn't null. ConsoleLifetime does `options?.Value ?? throw new ArgumentNullException(nameof(options));` which I think is reasonable enough.

test/Microsoft.Extensions.Hosting.Tests/HostTests.cs
```diff
@@ -505,6 +505,67 @@ public void HostCanBeStarted()
         }
 
         [Fact]
+        public async Task HostStopAsyncUsesDefaultTimeoutIfGivenTokenDoesNotFire()
```
halter73: Nit: Shorten the `Task.Delay(TimeSpan.FromSeconds(8))` in `HostStopAsyncCanBeCancelledEarly` to be less than 5 seconds.

PR #1239, opened 2017-10-16
Add RepositoryRoot

ryanbrandenburg:


jkotalik:
I don't think this needs to be a PR. We can just make one PR in a main repository and then push the rest to dev.

ryanbrandenburg:
I prefer to do it via PR because it automatically runs travis and appveyor tests, which have found lots of problems in this kind of change in the past. I don't expect people to review them (though you're welcome to if you want), I usually CC someone with context on one of them, then merge the rest once their tests pass.

PR #1243, opened 2017-10-26
Add a way to prevent specific hosting startup from runnning

pakrym:
Prevent hosting from even trying to load assembly


Tratcher:
:-1: my comments from https://github.com/aspnet/AzureIntegration/issues/110#issuecomment-335585275 still apply.

davidfowl:
👍 I like it.

PR #1248, opened 2017-10-28
Add TestServer support for directly constructing HttpContexts

Tratcher:
#816, #1135 Customers frequently ask how to test aspects of their application not exposed from HttpClient/HttpRequestMessage. These include client certificates, local and remote IPs and ports, windows auth user, adding/removing or manipulating features, etc.. This capability used to exist in Katana but was dropped when refactoring for Core.

This adds a new HttpContextBuilder that constructs the test context, lets you modify it as you see fit, and then send it. This gives you deep control over the request state for testing.

I've also moved ClientHandler and WebSocketClient to use this new layer to reduce duplication.

/cc: @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
+++ b/src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
@@ -33,12 +33,7 @@ namespace Microsoft.AspNetCore.TestHost
         /// <param name="application">The <see cref="IHttpApplication{TContext}"/>.</param>
         public ClientHandler(PathString pathBase, IHttpApplication<Context> application)
         {
-            if (application == null)
-            {
-                throw new ArgumentNullException(nameof(application));
-            }
-
-            _application = application;
+            _application = application ?? throw new ArgumentNullException(nameof(application));
 
             // PathString.StartsWithSegments that we use below requires the base path to not end in a slash.
             if (pathBase.HasValue && pathBase.Value.EndsWith("/"))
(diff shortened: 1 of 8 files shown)


davidfowl:
At first glance this looks over engineered. What does this do that an Action\<HttpContext\> can’t do?

Tratcher:
What part?

Most of the code already existed in ClientHandler and WebSocketClient. This deduplicates that code and gives it an intermediate API that mirrors the RequestBuilder.

Check out the BasicMiddleware PR https://github.com/aspnet/BasicMiddleware/pull/267, the new API is a lot cleaner than even HttpClient for some common scenarios.

davidfowl:
The HttpContextBuilder with the extension methods, I’m not seeing what is much easier.

Tratcher:
The fluent APIs for HttpContextBuilder are only sugar to reduce the amount of code you need to type, that's separate from the benefits of providing more direct access to server data structures.

To achieve the same abstraction benefits without the fluent API you could do something like this:
```
// TestServer
public Task<HttpContext> SendAsync(Action<HttpContext> configureRequest, CanceallationToken cancel = default);

var context = await server.SendAsync(c => { c.Request.Path = "/foo"; });
```

In this version HttpContextBuilder would be internal.

davidfowl:
Agreed. Why do we need the builder? Can you remove it from this PR?

src/Microsoft.AspNetCore.TestHost/ClientHandler.cs
```diff
@@ -64,187 +59,74 @@ public ClientHandler(PathString pathBase, IHttpApplication<Context> application)
...
-
-            internal RequestState(HttpRequestMessage request, PathString pathBase, IHttpApplication<Context> application)
-            {
-                _request = request;
-                _application = application;
-                _responseTcs = new TaskCompletionSource<HttpResponseMessage>();
-                _requestAbortedSource = new CancellationTokenSource();
-                _pipelineFinished = false;
+                req.Protocol = "HTTP/" + request.Version.ToString(fieldCount: 2);
+                req.Method = request.Method.ToString();
```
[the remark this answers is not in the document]
Tratcher: SetMethod was removed.

Tratcher:
Rebased and removed the builder syntax.

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
```diff
@@ -0,0 +1,125 @@
...
+            var request = _httpContext.Request;
+            request.Protocol = "HTTP/1.1";
+            request.Method = HttpMethods.Get;
+
+            _httpContext.Features.Set<IHttpResponseFeature>(_responseFeature);
+            var requestLifetimeFeature = new HttpRequestLifetimeFeature();
+            requestLifetimeFeature.RequestAborted = _requestAbortedSource.Token;
+            _httpContext.Features.Set<IHttpRequestLifetimeFeature>(requestLifetimeFeature);
+            
+            _responseStream = new ResponseStream(ReturnResponseMessageAsync, AbortRequest);
```
JunTaoLuo: This can be created at the declaration without being in the constructor no?

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -53,73 +53,20 @@ public Action<HttpRequest> ConfigureRequest
...
-                var request = httpContext.Request;
-                request.Protocol = "HTTP/1.1";
+                var request = context.Request;
                 var scheme = uri.Scheme;
                 scheme = (scheme == "ws") ? "http" : scheme;
                 scheme = (scheme == "wss") ? "https" : scheme;
                 request.Scheme = scheme;
-                request.Method = "GET";
+                request.Method = HttpMethods.Get;
                 var fullPath = PathString.FromUriComponent(uri);
```
JunTaoLuo: nit: let's use the same logic as `ClientHandler` https://github.com/aspnet/Hosting/pull/1248/files#diff-22356e5b7de1c9d8b3046bd1c6418043R81

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        internal void Configure(Action<HttpContext> configureContext)
+        {
+            if (configureContext == null)
+            {
+                throw new ArgumentNullException(nameof(configureContext));
+            }
+
+            configureContext(_httpContext);
```
JunTaoLuo: `=> configureContext?.Invoke(_httpContext) ?? throw new ArgumentNullException(nameof(configureContext));`?

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
```diff
@@ -0,0 +1,125 @@
...
+        internal async Task ReturnResponseMessageAsync()
+        {
+            // Check if the response has already started because the TrySetResult below could happen a bit late
+            // (as it happens on a different thread) by which point the CompleteResponseAsync could run and calls this
+            // method again.
+            if (!_responseFeature.HasStarted)
+            {
+                // Sets HasStarted
+                await _responseFeature.FireOnSendingHeadersAsync();
+                var setResult = _responseTcs.TrySetResult(_httpContext);
```
JunTaoLuo: what is setResult used for?

src/Microsoft.AspNetCore.TestHost/WebSocketClient.cs
```diff
@@ -53,73 +53,20 @@ public Action<HttpRequest> ConfigureRequest
...
-                // Request
-                var request = httpContext.Request;
-                request.Protocol = "HTTP/1.1";
+                var request = context.Request;
                 var scheme = uri.Scheme;
                 scheme = (scheme == "ws") ? "http" : scheme;
                 scheme = (scheme == "wss") ? "https" : scheme;
                 request.Scheme = scheme;
-                request.Method = "GET";
+                request.Method = HttpMethods.Get;
```
JunTaoLuo: not needed, it's set to Get by default.

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
Tratcher: I tried but it wouldn't let me capture the delegates for the methods before the constructor.

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
Tratcher: Yuck

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
Tratcher: To prevent the compiler warning about ignoring return values

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
```diff
@@ -0,0 +1,250 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting.Server;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using static Microsoft.AspNetCore.Hosting.Internal.HostingApplication;
```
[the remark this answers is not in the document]
muratg: Agreed, But did we start doing this more widely now?  I know Razor was doing this but I haven't seen this pattern before much.

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
davidfowl: I second that yuck

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
davidfowl: There's a first class feature called "discards" for that now.

```C#
_ = _responseTcs.TrySetResult(_httpContext);
```

src/Microsoft.AspNetCore.TestHost/HttpContextBuilder.cs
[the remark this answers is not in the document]
Tratcher: Only in a few places where there's a strangely nested type.

Issue #1207, opened 2017-09-14
WebHostService should call IWebHost.StopAsync in OnStop


Tratcher:
https://github.com/aspnet/Hosting/blob/8b30efbe750663fa79146108f2b5f9d6bf0728de/src/Microsoft.AspNetCore.Hosting.WindowsServices/WebHostService.cs#L47-L53
This will enable graceful shutdown. This is a regression in 2.0 when graceful shutdown was moved from Dispose to the new StopAsync API.

PR #1257, opened 2017-11-07
Trigger graceful shutdown on stopping webhost service

JunTaoLuo:
https://github.com/aspnet/Hosting/issues/1207


src/Microsoft.AspNetCore.Hosting.WindowsServices/WebHostService.cs
```diff
@@ -48,7 +48,7 @@ protected sealed override void OnStop()
         {
             _stopRequestedByWindows = true;
             OnStopping();
-            _host?.Dispose();
+            _host.StopAsync().GetAwaiter().GetResult();
```
Tratcher: try/finally Dispose?

JunTaoLuo: Anything we want to do for OnStopping?

Issue #1218, opened 2017-09-16
Reconcile conflicts between types in Microsoft.Extensions.Hosting and Microsoft.AspNetCore.Hosting


davidfowl:
We have a few types that overlap between the 2 hosting models and we need to decide what the long term and short term plans are:

- IHostingEnvironment
- IApplicationLifetime

A few options:
- We could add the Hosting.Extensions versions of these to the container of the WebHost and gradually change all of the libraries to use the new interfaces. This allows both systems to work side by side so we can do the gradual change over.

/c @glennc @Tratcher

muratg:
@davidfowl Anything you'd like to do in 2.x timeline?  Otherwise this mostly looks like breaking changes that we can only do in 3.0.

davidfowl:
@muratg no I'd like to do something in the 2.x timeline. There doesn't need to be any breaking changes here.

muratg:
@davidfowl so do you want to copy some of the interfaces to MEH from MAH and start using them? If these are exposed on any public API, that is a breaking change, right?  Do you have a picture of what you want in your mind?

davidfowl:
@muratg I sent a PR 😄

alexk8:
@davidfowl , 
Sorry guys, I'm a simple dev, and I dont understand why this issue was closed, the conflict (kind of) leaked into 2.1 release.... We now have two not-interchangable IHostingEnvironment. Why do we have a duplicate, not an extension? I propose a fix:
```
namespace Microsoft.AspNetCore.Hosting
{
    public interface IHostingEnvironment : Microsoft.Extensions.Hosting.IHostingEnvironment
    {
        string WebRootPath { get; set; }
        IFileProvider WebRootFileProvider { get; set; }
    }
}
```
Do you find it appropriate?

davidfowl:
What does this fix?

alexk8:
This fixes a confusing code duplication, and particularly needed for this:
```
            var builder = new WebHostBuilder()
            .ConfigureAppConfiguration((ctx, cfgBld) =>
            {
                var env = ctx.HostingEnvironment;
                // here, I needed Microsoft.Extensions.Hosting.IHostingEnvironment, but get Microsoft.AspNetCore.Hosting.IHostingEnvironment

                // this works, but isn't a safe cast
                var env2 = (Microsoft.Extensions.Hosting.IHostingEnvironment)env;

                //now I can use env2
            });
```

Tratcher:
[...]

PR #1259, opened 2017-11-09
Support the Extensions.Abstractions from Hosting.Abstractions

davidfowl:
- The goal here is to enable components that use hosting abstractions to use
the web host. It lets us start to decouple components from the web host abstractions
where possible while not breaking any existing components. This will allow things
to work in both the generic host and the web host. The one snafu is the WebHostBuilderContext
which has an IHostingEnvironment typed as the AspNetCore.Abstractions type.
- Updated tests.

#1218

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
     /// <summary>
     /// Allows consumers to perform cleanup during a graceful shutdown.
     /// </summary>
-    public class ApplicationLifetime : IApplicationLifetime
+    public class ApplicationLifetime : IApplicationLifetime, Extensions.Hosting.IApplicationLifetime
     {
         private readonly CancellationTokenSource _startedSource = new CancellationTokenSource();
         private readonly CancellationTokenSource _stoppingSource = new CancellationTokenSource();
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.Hosting
     /// </summary>
     public class WebHostBuilder : IWebHostBuilder
     {
-        private readonly IHostingEnvironment _hostingEnvironment;
+        private readonly HostingEnvironment _hostingEnvironment;
         private readonly List<Action<WebHostBuilderContext, IServiceCollection>> _configureServicesDelegates;
 
         private IConfiguration _config;
@@ -239,7 +239,8 @@ namespace Microsoft.AspNetCore.Hosting
 
             var services = new ServiceCollection();
             services.AddSingleton(_options);
-            services.AddSingleton(_hostingEnvironment);
+            services.AddSingleton<IHostingEnvironment>(_hostingEnvironment);
+            services.AddSingleton<Extensions.Hosting.IHostingEnvironment>(_hostingEnvironment);
             services.AddSingleton(_context);
 
             var builder = new ConfigurationBuilder()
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
     /// <summary>
     /// Allows consumers to perform cleanup during a graceful shutdown.
     /// </summary>
-    public class ApplicationLifetime : IApplicationLifetime
+    public class ApplicationLifetime : IApplicationLifetime, Extensions.Hosting.IApplicationLifetime
```
muratg: Isn't this breaking?

davidfowl: Why would it be?

JunTaoLuo:
For decoupling, do we also need to update the extension methods that currently only take M.A.H's version of IHostingEnvironment?

JunTaoLuo:
FYI, even when we the ApplicationLifetime and HostingEnvironment as both types, only use the M.A.H's version in the WebHost mostly, https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L128

davidfowl:
> For decoupling, do we also need to update the extension methods that currently only take M.A.H's version of IHostingEnvironment?

Which ones? We should add ones to the Hosting.Extensions.Abstractions.

> FYI, even when we the ApplicationLifetime and HostingEnvironment as both types, only use the M.A.H's version in the WebHost mostly, https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L128

That's fine it's the same instance internally.

JunTaoLuo:
Do we need to update these? https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingEnvironmentExtensions.cs I need to check if there are any other extensions.

davidfowl:
@JunTaoLuo nope. Those could be replicated in the Extensions.Hosting.Abstractions project.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.Hosting
     /// </summary>
     public class WebHostBuilder : IWebHostBuilder
     {
-        private readonly IHostingEnvironment _hostingEnvironment;
+        private readonly HostingEnvironment _hostingEnvironment;
```
JunTaoLuo: Actually I think this is only used in BuildCommonServices so there's no need to keep this class member.

davidfowl:
> The changes aren't breaking so they are fine to take but I find this change kind of odd, why not just switch to the extension types in 3.0 when we are going to make the breaking change anyway. Is not switching over blocking any other work?

The idea is that we start building components that can plug into both hosts. This needs to be done so that we can start doing that. One of the first consumers of this will be the bedrock effort.

src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
Tratcher: It's interesting that you choose to have the concrete class implement both interfaces rather than having the interfaces chain. Consumers must explicitly choose one interface or the other. Wouldn't chaining them make the transition even smoother? An app could convert one method at a time and work backwards.

E.g. MyMethod(Ex.H.IApplicationLifetime lifetime) could accept either Extensions.Hosting.IApplicationLifetime or AspNetCore.Hosting.IApplicationLifetime.

Or do you assume everything will come directly from DI and they only need to update one constructor?

davidfowl: The goal is to have new code take in extensions hosting where possible via ctor injection. This change also doesn’t preclude the derivation if we see benefits there.

Issue #1260, opened 2017-11-09
Allow IISExpress deployer to deploy either inprocess or outofprocess based on a deployment parameter


jkotalik:
Current in IISIntegration, we have two sample applications that can be consolidated into one if we can change the hostingModel value in the .config file to inprocess or outofprocess. We can add an inprocess or out of process parameter here: https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/RemoteWindowsDeployer/RemoteWindowsDeploymentParameters.cs#L8 and then change the .config file using a special string "[HostingModel]" and rewrite it to inprocess or outofprocess.

jkotalik:
#1261

PR #1261, opened 2017-11-09
Adds hosting model as an extra deployment parameter

jkotalik:
For #1260

### DIFF
--- a/korebuild-lock.txt
+++ b/korebuild-lock.txt
@@ -1,2 +1,2 @@
-version:2.1.0-preview1-15549
-commithash:f570e08585fec510dd60cd4bfe8795388b757a95
+version:2.1.0-preview1-15564
+commithash:1f3f14382764e06b7e691e5ee89d12a280249284
--- a/korebuild-lock.txt
+++ b/korebuild-lock.txt
@@ -1,2 +1,2 @@
-version:2.1.0-preview1-15564
-commithash:1f3f14382764e06b7e691e5ee89d12a280249284
+version:2.1.0-preview1-15549
+commithash:f570e08585fec510dd60cd4bfe8795388b757a95
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
@@ -106,6 +106,8 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
 
         public string PublishedApplicationRootPath { get; set; }
 
+        public HostingModel HostingModel { get; set; }
+
         /// <summary>
         /// Environment variables to be set before starting the host.
         /// Not applicable for IIS Scenarios.
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
@@ -148,8 +148,9 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                         {
                             Logger.LogTrace($"Config File Content:{Environment.NewLine}===START CONFIG==={Environment.NewLine}{{configContent}}{Environment.NewLine}===END CONFIG===", serverConfig);
                         }
-
-                        serverConfig.Replace("[HostingModel]", DeploymentParameters.HostingModel.ToString());
+                        var hostingModel = DeploymentParameters.HostingModel.ToString();
+                        serverConfig.Replace("[HostingModel]", hostingModel);
+                        Logger.LogDebug("Writing HostingModel '{hostingModel} to ", hostingModel);
 
                         File.WriteAllText(DeploymentParameters.ServerConfigLocation, serverConfig);
                     }
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
@@ -148,9 +148,13 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                         {
                             Logger.LogTrace($"Config File Content:{Environment.NewLine}===START CONFIG==={Environment.NewLine}{{configContent}}{Environment.NewLine}===END CONFIG===", serverConfig);
                         }
-                        var hostingModel = DeploymentParameters.HostingModel.ToString();
-                        serverConfig.Replace("[HostingModel]", hostingModel);
-                        Logger.LogDebug("Writing HostingModel '{hostingModel} to ", hostingModel);
+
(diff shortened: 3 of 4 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -106,6 +106,8 @@ public class DeploymentParameters
 
         public string PublishedApplicationRootPath { get; set; }
 
+        public HostingModel HostingModel { get; set; }
```
JunTaoLuo: Set a default value?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -148,6 +148,9 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
                         {
                             Logger.LogTrace($"Config File Content:{Environment.NewLine}===START CONFIG==={Environment.NewLine}{{configContent}}{Environment.NewLine}===END CONFIG===", serverConfig);
                         }
+                        var hostingModel = DeploymentParameters.HostingModel.ToString();
+                        serverConfig.Replace("[HostingModel]", hostingModel);
+                        Logger.LogDebug("Writing HostingModel '{hostingModel} to ", hostingModel);
```
JunTaoLuo: mismatched single quote and missing "config" after "to"

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -148,6 +148,9 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
                         {
                             Logger.LogTrace($"Config File Content:{Environment.NewLine}===START CONFIG==={Environment.NewLine}{{configContent}}{Environment.NewLine}===END CONFIG===", serverConfig);
                         }
+                        var hostingModel = DeploymentParameters.HostingModel.ToString();
+                        serverConfig.Replace("[HostingModel]", hostingModel);
```
JunTaoLuo: Why not just chain it here: https://github.com/aspnet/Hosting/blob/05fd382b93cb51668d1d8858e79292601ca71b18/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs#L138-L141

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
jkotalik: Other values don't have default values (including enums). See: https://github.com/aspnet/Hosting/pull/1261/files#diff-8478a3aebbec4604e8f2cf79bb152195R105

JunTaoLuo: By default, it would take the first value as the default. Do we want to default all tests to InProcess?

korebuild-lock.txt
```diff
@@ -1,2 +1,2 @@
-version:2.1.0-preview1-15549
-commithash:f570e08585fec510dd60cd4bfe8795388b757a95
+version:2.1.0-preview1-15564
```
JunTaoLuo: @natemcmaster are these changes expected every now and then?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
jkotalik: The contentRoot and port are inputs to StartIISExpressAsync. HostingModel isn't guaranteed to be set. I will add a check to see if HostingModel exists before setting it though.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
jkotalik: Probably should be out of process.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -149,6 +149,13 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
                             Logger.LogTrace($"Config File Content:{Environment.NewLine}===START CONFIG==={Environment.NewLine}{{configContent}}{Environment.NewLine}===END CONFIG===", serverConfig);
                         }
 
+                        if (serverConfig.Contains("[HostingModel]"))
```
JunTaoLuo: Not needed as string.Replace will no-op if the old value is not found. Also we should not change the config after the changes were logged in https://github.com/aspnet/Hosting/pull/1261/files#diff-727b9f7850879229f0c886f7448a0b04R149. I still think it's best to include these updates at https://github.com/aspnet/Hosting/pull/1261/files#diff-727b9f7850879229f0c886f7448a0b04R138

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
JunTaoLuo: I'm okay with changing the order of the enums but personally I prefer the more explicit approach where we just set the default value. Takes out all ambiguity.

korebuild-lock.txt
natemcmaster: sure, if you need a new version of build tools. The only reason to hold off for now is that the jump from 2.1.0-preview1-15549 to 2.1.0-preview1-15564 puts you ahead of aspnet/Universe

natemcmaster: cref https://github.com/aspnet/Universe/blob/dev/korebuild-lock.txt

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
jkotalik: This check is for logging. It doesn't make sense to say we are replacing the hostingModel if the hosting model isn't replaced. I do agree with moving it above the config file content though.

JunTaoLuo: These are debug logs so there's no harm in saying we tried replacing the hosting model but didn't end up finding anything to replace. Then again, I don't feel too strongly about it either way.

jkotalik:
The tests in [IISIntegration](https://github.com/aspnet/IISIntegration/tree/dev/test/IISIntegration.IISServerFunctionalTests) are effectively the same as what are in ServerTests. However those tests are currently disabled until ANCM is in universe. If IIS integration relies on a new change to ANCM, there is no way to pick up that change. Eventually, ServerTests will have tests for InProcess, however it is too painful.

PR #1262, opened 2017-11-10
Fix path check for ANCM

jkotalik:
Mistake on #1258


pakrym:
We need tests for `IISExpressDeployer`.

jkotalik:
It would require tests for all deployers. I'll make an issue for that.

PR #1267, opened 2017-11-11
Delay IServer and Startup.Configure until StartAsync

Tratcher:
#1263 Tools want to call Startup.ConfigureServices without resolving Kestrel options like loading server certs. Delay initializing the server until StartAsync. The catch is that we also have to delay Startup.Configure as the server features are available there.

The main consequence is that this delays many exceptions from Build to StartAsync. There is some amount of risk here for scenarios where those two aren't called together, but we don't have any examples of those scenarios.

TODO: Test this with Kestrel to make sure Kestrel's options are properly delayed.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.ExceptionServices;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -83,7 +84,6 @@ public IServiceProvider Services
         {
             get
             {
-                EnsureApplicationServices();
```
davidfowl: Does this means Services is null until Start is called?

Tratcher: No. The WebHostBuilder always calls Initialize which sets the Services. I removed this Ensure line because it was redundant. 
https://github.com/aspnet/Hosting/blob/311a92ddee5a009b2e96beb18838e816b66d533b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L193

davidfowl: Ah good

Tratcher:
Confirmed, this delays all Kestrel configuration errors to StartAsync.

Tratcher:
@muratg we should discuss the risks

divega:
Nice!

PR #1286, opened 2017-12-05
Changes OS bitness check to Environment rather than System.Runtime.InteropServices.RuntimeInformation

jkotalik:
Root issue: https://github.com/dotnet/corefx/issues/25706
For now, we can by using Environment.Is64BitOperatingSystem. This will fix many AppVeyor failures across the stack and Win10 build issues on the Fall Creator's update. This hopefully will be fixed via the net471 fix.

@pranavkm can I use the Environment directly here rather than using reflection?
cc/ @pakrym


benaadams:
Is the bit size of the OS important or the bit size of the process?

Tratcher:
This is used to get the correct version of ANCM to match IIS Express. We pull IIS Express from c:\program files\ so the bitness varies based on OS. We'll have to re-work this in the future to load the right bitness of IIS Express for in-proc ANCM and skip the 64bit tests on 32bit OS's.

PR #1291, opened 2017-12-09
Add default app data service to provide an app data path and filesystemprovider

javiercn:
Adds a default IHostingApplicationData service and implementation to allow apps and frameworks to search for data/files they need for work.
By default we follow the same conventions as data protection to define storage locations and if none of them is found we resort to AppContext.BaseDirectory/APP_DATA


src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationData.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    public class HostingApplicationData : IHostingApplicationData
+    {
+        public string ApplicationDataPath { get; set; }
+
+        public IFileProvider ApplicationDataFileProvider { get; set; }
+
+        internal void Initialize(WebHostOptions options)
+        {
+            CreateHostingApplicationData(options.ApplicationDataPath);
```
Tratcher: Combine these methods

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationData.cs
```diff
@@ -0,0 +1,85 @@
...
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    public class HostingApplicationData : IHostingApplicationData
+    {
+        public string ApplicationDataPath { get; set; }
+
+        public IFileProvider ApplicationDataFileProvider { get; set; }
+
+        internal void Initialize(WebHostOptions options)
```
Tratcher: Don't pass in WebHostOptions if you're only going to consume one field, pass that value in directly.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationData.cs
```diff
@@ -0,0 +1,85 @@
...
+            directoryInfo = directoryInfo ??
+                GetOrCreateDirectory(GetFolderPath(localAppDataFromEnvPath, "ASP.NET")) ??
+                GetOrCreateDirectory(GetFolderPath(userProfilePath, "AppData", "Local", "ASP.NET")) ??
+                GetOrCreateDirectory(GetFolderPath(homePath, ".aspnet")) ??
+                GetOrCreateDirectory(GetFolderPath(localAppDataFromSystemPath, "ASP.NET")) ??
+                GetOrCreateDirectory(GetFolderPath(AppContext.BaseDirectory, "APP_DATA"));
+
+            return directoryInfo?.FullName;
+
+            DirectoryInfo GetOrCreateDirectory(string path)
```
Tratcher: It's a little strange that a method named GetOrCreateDirectory would return null.

src/Microsoft.Extensions.Hosting/HostBuilder.cs
```diff
@@ -150,6 +154,69 @@ private string ResolveContentRootPath(string contentRootPath, string basePath)
...
+            var homePath = Environment.GetEnvironmentVariable("HOME");
+
+            // To preserve backwards-compatibility with 1.x, Environment.SpecialFolder.LocalApplicationData
+            // cannot take precedence over $LOCALAPPDATA and $HOME/.aspnet on non-Windows platforms.
+            // We do this in here too to keep consistency with the locations where data protection stores keys.
+            directoryInfo = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+                GetOrCreateDirectory(GetFolderPath(localAppDataFromSystemPath, "ASP.NET")) :
+                null;
+
+            directoryInfo = directoryInfo ??
```
Tratcher: Show this search list to @DamianEdwards and explain to him what kind of files you'd expect to store here.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationData.cs
```diff
@@ -0,0 +1,85 @@
...
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Microsoft.Extensions.FileProviders;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    public class HostingApplicationData : IHostingApplicationData
```
Tratcher: Why did you structure this type differently for AspNetCore vs Extensions? E.g. for AspNetCore the initialization code is in the type, but for Extensions it's in the calling type.

src/Microsoft.Extensions.Hosting/HostBuilder.cs
```diff
@@ -150,6 +154,69 @@ private string ResolveContentRootPath(string contentRootPath, string basePath)
...
+                GetOrCreateDirectory(GetFolderPath(localAppDataFromEnvPath, "ASP.NET")) ??
+                GetOrCreateDirectory(GetFolderPath(userProfilePath, "AppData", "Local", "ASP.NET")) ??
+                GetOrCreateDirectory(GetFolderPath(homePath, ".aspnet")) ??
+                GetOrCreateDirectory(GetFolderPath(localAppDataFromSystemPath, "ASP.NET")) ??
```
Tratcher: You're still using ASP.NET for Extensions? Should this use a more generic name?

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationData.cs
javiercn: I'm following the exact same pattern as IHostingEnvironment/HostingEnvironment. They are structured differently between AspNetCore.Hosting and Extensions.Hosting.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationData.cs
javiercn: I can call it GetOrTryCreateDirectory, do you have a suggestion for a better name?

src/Microsoft.Extensions.Hosting/HostBuilder.cs
javiercn: This mimics the locations DataProtection uses, we don't want to define/introduce a new set of locations. If in the future we change the locations we will change them all together to something new.

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationData.cs
```diff
@@ -0,0 +1,85 @@
...
+            // To preserve backwards-compatibility with 1.x, Environment.SpecialFolder.LocalApplicationData
+            // cannot take precedence over $LOCALAPPDATA and $HOME/.aspnet on non-Windows platforms.
+            // We do this in here too to keep consistency with the locations where data protection stores keys.
+            directoryInfo = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+                GetOrCreateDirectory(GetFolderPath(localAppDataFromSystemPath, "ASP.NET")) :
+                null;
+
+            directoryInfo = directoryInfo ??
+                GetOrCreateDirectory(GetFolderPath(localAppDataFromEnvPath, "ASP.NET")) ??
+                GetOrCreateDirectory(GetFolderPath(userProfilePath, "AppData", "Local", "ASP.NET")) ??
```
mkArtakMSFT: Some of these literals are used multiple times ("ASP.NET" folder). Aren't there constants for these well-known paths/folders ?

src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingApplicationData.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.FileProviders;
+
+namespace Microsoft.AspNetCore.Hosting
+{
+    public interface IHostingApplicationData
```
davidfowl: I'm not sure I like this name. We have ContentRoot and WebRoot. We should stick to that paradigm.

davidfowl:
Where is the call to Use* call on the `IWebHostBuilder`? UseApplicationDataRoot?

src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingApplicationData.cs
javiercn: I had application data root but I hated it. @glennc had some more high level feedback, so we are pausing this for preview 1 until you're back and we can all discuss it, but it seems that you're ok with the approach I took? See https://github.com/aspnet/Hosting/issues/1294 for details on an alternative proposal

PR #1298, opened 2017-12-20
Attempts to fix urls from appsettings.json, CLI

b-stime:
WebHost checks _config for the deprecated "server.urls" key but it also checks _config for the new key . I believe it should use _startup.Configure for the new key. The ultimate problem I'm trying to solve is that:

WebHost.CreateDefaultBuilder(args)
                .UseStartup<Startup>()
                .Build();

...doesn't read the urls from appsettings.json or the command line. The builder pushes the values into the Startup constructor so they should be available via Startup.Configure.

I haven't tested this change--I'm editing via the GitHub web interface.


Tratcher:
This code doesn't compile. _startup.Configure is a method, not an IConfiguration instance. I'm closing this and we can discuss it issue over at #1299.

PR #1302, opened 2017-12-22
Add missing baselines

ryanbrandenburg:
These baselines were missing, so I added them.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.WindowsServices/baseline.netcore.json
+++ b/src/Microsoft.AspNetCore.Hosting.WindowsServices/baseline.netcore.json
@@ -1,0 +1,122 @@
+{
+  "AssemblyIdentity": "Microsoft.AspNetCore.Hosting.WindowsServices, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
+  "Types": [
+    {
+      "Name": "Microsoft.AspNetCore.Hosting.WindowsServices.WebHostService",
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Hosting/baseline.netcore.json
```diff
@@ -0,0 +1,397 @@
+{
+  "AssemblyIdentity": "Microsoft.Extensions.Hosting, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
```
Tratcher: Isn't the baseline supposed to be the prior public release? This component is new, it hasn't shipped before.

ryanbrandenburg: Is that true of the other component too?

Tratcher: No, WindowsServices 2.0 had a net461 version, it should have a baseline. Only the netcoreapp2.0 TFM is new.

muratg: Do we have a way to differentiate the baselines based on the TFM?

ryanbrandenburg: We differentiate Core vs Framework based on the name of the baseline file generated.

ryanbrandenburg:
:up::date:

Issue #1060, opened 2017-05-02
Allow custom IServiceCollection in build method


ti24horas:
Is that possible to pass a custom IServiceCollection instead of using the internal IServiceCollection created in Build method?

My case is that the WebHostBuilder is just part of a bigger application (which contains part of code running as a service, for example)

Because the services are registered by an external IServiceCollection, it would be impossible to utilize the existing IServiceProvider without duplicating registrations (or delegating them to existing IServiceProvider).

davidfowl:
This isn't possible today. The WebHostBuilder is the root of the DI universe and you can't replace the built in `IServiceCollection`.

ti24horas:
@davidfowl I wonder if the IServiceCollection could be passed as a parameter in "Build" method of WebHostBuilder (as an optional parameter, btw), I think it would have a slight impact on what's running today.

I just don't want to mimic the WebHost class just to be able to pass an existing service configuration.

PR #1304, opened 2017-12-27
Enable to pass an existing IServiceCollection (#1060).

koraykazgan:
Allow to pass an existing IServiceCollection instance to WebHostBuilder which should be used instead of creating a new one inside WebHostBuilder. Could be a solution to issue #1060.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -32,6 +32,7 @@ namespace Microsoft.AspNetCore.Hosting
         private WebHostBuilderContext _context;
         private bool _webHostBuilt;
         private List<Action<WebHostBuilderContext, IConfigurationBuilder>> _configureAppConfigurationBuilderDelegates;
+        private Func<IServiceCollection> _serviceCollectionDelegate;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


davidfowl:
I'm not 100% certain I like this. The caveats here are that the service collection cannot be modified after the it is give to the web host. That's not clear from the method that this is the case. 

Also, why is it a Func\<IServiceCollection\> instead of just an IServiceCollection

koraykazgan:
> The caveats here are that the service collection cannot be modified after the it is give to the web host

I see your point. But the main issue here is that WebHostBuilder is passing the serviceProvider to WebHost to do service location. Of course this is not easy to change but therefore its the serviceCollection's last stop. 

> That's not clear from the method that this is the case.

Maybe it is enough to just change the methods name to for example BuildServiceCollection?

> Also, why is it a Func<IServiceCollection> instead of just an IServiceCollection

I also thought about that after my pull request. My first intention was to be able to create the ServiceCollection as late as possible. But you are right, maybe it is just simpler to pass an IServiceCollection.

davidfowl:
Take a look at  https://github.com/aspnet/Hosting/commit/6b095cf533950e7fc93ebc475093bafac70058cb and see if you still need this feature. I don't think you do but I don't understand your scenario enough to make that call.

davidfowl:
Closing this out. If you can't make the change above work I'd like to understand why

PR #1306, opened 2017-12-29
Set the default deploy configuration based on compilation configuration

natemcmaster:
Currently tests always run in "Debug". This changes the default deployment config to match the test project's compilation config.


natemcmaster:
Nvm, this wouldn't work because the conditional is determined at compile time by the caller assembly. We'd have to update all test assemblies to call the method on DeploymentParameters.

We could use AssemblyConfigurationAttribute instead, but this also depends on knowing which assembly to look for.

Issue #1313, opened 2018-01-18
DI: Container does not dispose existing services when encountering exception in ConfigureServices()


Tratcher:
_From @helgemahrt on January 18, 2018 16:32_

I first encountered this issue in the context of Service Fabric and reported it here: [https://github.com/Azure/service-fabric-aspnetcore/issues/42](https://github.com/Azure/service-fabric-aspnetcore/issues/42) Please view this items for details. In a nutshell: When there's an exception in `ConfigureServices()` and we've added an IDisposable to the services collection, which has already been instantiated, the container doesn't dispose it correctly. 
The issue is amplified by Service Fabric, because it retries starting the web app/API without starting a new process, so the leak becomes more apparent. In our case, this lead to a high CPU hang which disabled some production level machines. However, the issue also reproduces in a standalone ASP.NET Core 2.0 web API. 
It doesn't matter whether I add an instance of my service manually to the container or whether I let the container create it, the Dispose() function is never called and the timer continues to fire after exiting ConfigureServices(), when the web app already failed to start. E.g. this also fails:
```
            services.AddSingleton<MyServiceImpl>();

            services.AddMvc();

            services.BuildServiceProvider().GetService<MyServiceImpl>();

            throw new Exception("Asdf");
```

_Copied from original issue: aspnet/Home#2800_

davidfowl:
@helgemahrt Nobody is disposing the container you built, is that what you're talking about? Why are you calling BuildServiceProvider?

davidfowl:
Ok so the problem is that if building the host throws and the host process doesn't terminate, there's no way to dispose the host (you never got a valid host because IWebHostBuilder.Build throws).

davidfowl:
In 2.1 failures in Configure no longer throw as part of building the service provider so that should be fine.  

If ConfigureServices throws though, it's a similar issue (but only for services added from the host or via the WebHostBuilder).

davidfowl:
We should consider patching this @muratg @Eilon @DamianEdwards

Eilon:
@davidfowl can you conceive of a workaround to this, for anyone affected in the meantime?

muratg:
@davidfowl Are you working on this in the preview1 branch or dev only?

davidfowl:
@Eilon @muratg There's no good workaround for 2.0.x or 1.x. It requires a bunch of . This fix is in dev only. Do we want it in preview1?

muratg:
[...]

PR #1318, opened 2018-01-27
Dispose the host if Initialize fails

davidfowl:
#1313

### DIFF
--- a/test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
+++ b/test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
@@ -370,6 +370,23 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         [Fact]
+        public void ServiceProviderDisposedOnBuildException()
+        {
+            var service = new DisposableService();
+            var hostBuilder = new WebHostBuilder()
+                .UseServer(new TestServer())
+                .ConfigureServices(services =>
+                {
+                    // Added as a factory since instances are never disposed by the container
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -1068,6 +1095,24 @@ public void Configure(IWebHostBuilder builder)
...
         }
 
+        public class StartupWithResolvedDisposableThatThrows
+        {
+            public StartupWithResolvedDisposableThatThrows(DisposableService service)
+            {
+
+            }
+
+            public void ConfigureServices(IServiceCollection services)
```
JunTaoLuo: nit: `public void ConfigureServices(IServiceCollection services) => throw new InvalidOperationException()`

test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs
```diff
@@ -1068,6 +1095,24 @@ public void Configure(IWebHostBuilder builder)
...
+            {
+                throw new InvalidOperationException();
+            }
+
+            public void Configure(IApplicationBuilder app)
```
JunTaoLuo: nit: continue the trend where empty methods are shortened as `public void Foo(...) {}`?

davidfowl: I don't think that's a consistent trend.

PR #500, opened 2015-11-23
Remove single moq usage

pakrym:
Issue: https://github.com/aspnet/Hosting/issues/499
@Tratcher @pranavkm

PR #1321, opened 2018-01-30
Allow external container to resolve IStartup

### LINKED ISSUE #550: Allow acceptance of pre-created IServiceProvider instance
Currently Asp.Net 5 is solely in charge of instantiating an `IServiceProvider` instance to be used.  While this works perfectly fine in Asp.Net websites, this is not ideal in circumstances where you are embedding Asp.Net 5 in another application (service or console application for example).   For example, a console application that is running in the background might use Asp.Net 5 to expose http endpoints for easy communication with external systems (health checks, event notifications, etc...)

In this circumstance, Asp.Net is a relatively minor piece of infrastructure for the application as a whole but the current implementation of Asp.Net's dependency injection strategy requires that Asp.Net is the sole creator of the ioc container.  This means that if you want a single ioc container for the whole application you must gather your service descriptors, pass them to Asp.Net, start the Asp.Net host, then call `IWebApplication.ApplicationServices` in order to retrieve the ioc container so it can be used in the non-Asp.net portions of the application.  Without doing this, any service that's registered as an instance or singleton risks not working as expected (since each ioc container can now creates it's own instance).

This seems to me to cause my application to be too tightly coupled to Asp.Net for dependency injection, and means that if any other minor infrastructure makes the same assumption (requires it to be the creator of ioc containers) it becomes impossible to have consistent and sane dependency injection strategy.

It seems like we should at least have the option to allow calling classes to request services that Asp.Net needs to register, let the original class create the `IServiceProvider` instance, then pass it to the `WebApplicationBuilder` for usage.

An example of how this could look:

```
    public class WebHost
    {
        private IHostingEngine _hostingEngine;
        private IApplication _application;

        public IServiceCollection GetServices()
        {
            // WebHostBuilder already has BuildHostedServices().  If this was decoupled from 
            // the call to execute the configureServices delegate inside then we should be able to 
            // somehow get a list of all services that Asp.Net wants to register by default
            var services = WebHostBuilder.GetDefaultServices();  
            services.AddMvc();
            return services;
        }

        public void Start(IServiceProvider serviceProvider)
        {
            _hostingEngine = new WebHostBuilder(GetConfig())
                .UseServer("Microsoft.AspNet.Server.Kestrel")
                .UseServiceProvider(serviceProvider)
                .UseStartup<WebHost>()
                .Build();

            _application = _hostingEngine.Start();
        }
    }

........

        {
            var services = GetApplicationServices();

            using (var host = new WebHost())
            {
                var webServices = host.GetServices();
                foreach (var webService in webServices)
                {
                    services.Add(webService);
                }

                var container = CreateIocContainer();
                host.Start(container);
            }
        }
```

Then (theoretically) the WebHostBuilder.Build() checks if a IServiceProvider was given, if so then it ignores the building of services, gets the resolution it needs and passes it into WebApplication's constructor. I do see that WebApplication adds to the collection but that can also be abstracted away by an IWebApplication..GetRequiredServices() method.

### LINKED ISSUE #1309: Allow resolving Startup from external service provider
When Unity container is configured to replace built in DI engine we discovered that it is not being used to create and resolve Startup class.

Currently WebHost builder is hard coded to create new built-in/internal service provider and resolve Startup class from that provider.  Configured provider is ignored and created later. As result no types registered with Unity are available to the Startup constructor.

For more information please see [this project](https://github.com/unitycontainer/microsoft-dependency-injection)

ENikS:
Fixed #1309

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
@@ -176,7 +176,10 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             if (_applicationServices == null)
             {
+                _applicationServices = _hostingServiceProvider.GetServices<IServiceProvider>()
+                                                              .LastOrDefault();
                 EnsureStartup();
+
                 _applicationServices = _startup.ConfigureServices(_applicationServiceCollection);
             }
         }
@@ -188,7 +191,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 return;
             }
 
-            _startup = _hostingServiceProvider.GetService<IStartup>();
+            _startup = (_applicationServices ?? _hostingServiceProvider).GetService<IStartup>();
 
             if (_startup == null)
             {
--- a/test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
+++ b/test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
@@ -70,7 +70,7 @@ namespace Microsoft.AspNetCore.TestHost
                 app.Use((ctx, next) => ctx.Response.WriteAsync(
                     $"{ctx.RequestServices.GetRequiredService<SimpleService>().Message}, {ctx.RequestServices.GetRequiredService<TestService>().Message}"));
         }
-
+        
         public class ThirdPartyContainer
         {
             public IServiceCollection Services { get; set; }
@@ -81,6 +81,84 @@ namespace Microsoft.AspNetCore.TestHost
             public ThirdPartyContainer CreateBuilder(IServiceCollection services) => new ThirdPartyContainer { Services = services };
 
             public IServiceProvider CreateServiceProvider(ThirdPartyContainer containerBuilder) => containerBuilder.Services.BuildServiceProvider();
+        }
+
+        [Fact]
+        public async Task ExternalContainerInstanceCanBeUsedToGetIStartup()
+        {
+            var external = new UnityContainer();
+            external.Register(typeof(TestService)); // Register outside builder service
+
+            var builder = new WebHostBuilder()
+                .ConfigureTestServices(services =>  // Register inside builder service
+                    services.AddSingleton(new SimpleService { Message = "OverridesConfigureServices" }))
+                .UseStartup<UnityStartup>()
+                .ConfigureServices(s => {           // Register outside provider instance
+                    s.AddSingleton<IServiceProvider>(external.BuildServiceProvider(s));
+                });
+
+            var host = new TestServer(builder);
+
+            var response = await host.CreateClient().GetStringAsync("/");
+
+            Assert.Equal("ConfigureServices, UnityStartupConstructor", response);
+        }
+
+        public class UnityContainer 
+        {
+            internal TestServiceCollection Registrations { get; set; } = new TestServiceCollection();
+
+
+            public void Register(Type registeredType)
+            {
+                Registrations.AddTransient(registeredType);
+            }
+
+            public IServiceProvider BuildServiceProvider(IServiceCollection services)
+            {
+                var collection = new TestServiceCollection();
+
+                foreach (var service in services) collection.Add(service);
+                foreach (var service in Registrations) collection.Add(service);
+                collection.AddSingleton(this);
+
+                return collection.BuildServiceProvider();
+            }
+
+            internal class TestServiceCollection : System.Collections.Generic.List<ServiceDescriptor>, 
+                                                   IServiceCollection
+            {
+            }
+        }
+
+        public class UnityStartup : IStartup
+        {
+            UnityContainer _container;
+            TestService _outside;
+
+            public UnityStartup(UnityContainer container, TestService outside)
+            {
+                _container = container ?? throw new ArgumentNullException(nameof(container));
+                _outside = outside   ?? throw new ArgumentNullException(nameof(outside));
+
+                _container.Registrations.AddSingleton(new TestService { Message = "UnityStartupConstructor" });
+            }
+
+            public void Configure(IApplicationBuilder app) =>
+                app.Use((ctx, next) => ctx.Response.WriteAsync(
+                    $"{ctx.RequestServices.GetRequiredService<SimpleService>().Message}, " +
+                    $"{ctx.RequestServices.GetRequiredService<TestService>().Message}"));
+
+            public IServiceProvider ConfigureServices(IServiceCollection services)
+            {
+                services.AddSingleton(new SimpleService { Message = "ConfigureServices" });
+
+                // Normally container would register more types from colection
+                // but for this test will just create updated provider
+                foreach (var service in _container.Registrations) services.Add(service);
+
+                return services.BuildServiceProvider();
+            }
         }
 
         [Fact]


src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -188,7 +195,7 @@ private void EnsureStartup()
                 return;
             }
 
-            _startup = _hostingServiceProvider.GetService<IStartup>();
+            _startup = (_applicationServices ?? _hostingServiceProvider).GetService<IStartup>();
```
ENikS: I am not sure if resolved IStartup instance should be registered with _hostingServiceProvider as well. If it does, let me know. I'll add it.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -176,8 +176,15 @@ private void EnsureApplicationServices()
         {
             if (_applicationServices == null)
             {
+                _applicationServices = (IServiceProvider)_applicationServiceCollection
```
davidfowl: - Doesn't this code just not call _startup.ConfigureServices(...) if an IServiceProvider is specified? 
- Any code that looks at service registrations is usually assuming too much.

ENikS: It just skips calling provider factory if instance is provided. The rest is the same as usual.

davidfowl:
Generally I don't like the approach and I think #550 cannot be fixed without changing the overall design of the WebHostBuilder. It *has* to be the owner of the DI container as things are designed today. This PR makes an attempt to make that not the case but it actually doesn't work well as it basically skips calling into ConfigureServices in the startup class.

ENikS:
If you run the test you will see that with exception of ServiceProviderFactory everything else is called properly. Calling Factory methods it this scenario are pointless when service provider already instantiated.

ENikS:
@davidfowl 
> This PR makes an attempt to make that not the case but it actually doesn't work well as it basically skips calling into ConfigureServices in the startup class.

Just checked it again and every single configure method is being called.

davidfowl:
I don’t see how any of this can work. If you pass a concrete service provider to the WebHostBuilder, what’s the contract? What service provider is used for each of the scenarios?  

Part of the confusion is that you claim this fixes #500, but it really doesn’t.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
davidfowl: Doesn't this code grab the IServiceProvider out of the `IServiceCollection` (which itself is an anti-pattern) and then skips calling _startup.ConfigureServices if it's an instance?

ENikS:
It fixes #550 ([Allow acceptance of pre-created IServiceProvider instance](https://github.com/aspnet/Hosting/issues/550)). I'll refactor the test so it is more obvious.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
```diff
@@ -176,8 +176,15 @@ private void EnsureApplicationServices()
...
+                    .LastOrDefault(d => d.ServiceType == typeof(IServiceProvider) &&
+                                        d.ImplementationInstance != null)?
+                    .ImplementationInstance;
```
ENikS: We can not do this: ``_hostingServiceProvider.GetService<IServiceProvider>();`` because it will return _hostingServiceProvider. We need to look it up in the list.

davidfowl: I think the entire feature is undesirable really and this is a hack.

davidfowl: The problem is that there are 3 service providers and the service provider you get when you build is different to the one added to the IServiceCollection. It's using the IServiceCollection like a property bag to shuttle the information to the WebHost

ENikS:
I replaced solution with less 'naive' approach. It should be more robust now.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
davidfowl: Why are you resolving the IStartup from the application services? That's just broken. The hosting service provider is responsible for building the IStartup. There are 2 separate phases.

davidfowl:
I fundamentally have an aversion to this change because IMO it doesn't make much sense to add an IServiceProvider to the IServiceCollection. There are 2 service providers:

- The hosting service provider - This activates the IStartup and the IServiceProviderFactory (it also has the LoggerFactory which causes a bunch of problems).
- The application service provider - This is a combination of the hosting services + the services added in IStartup.ConfigureServices. 

There's no physical way to replace the application service provider because hosting fundamentally needs to call into user code to then build the "final" service provider. Passing in an already constructed one will never work. This is what #550 is asking for.

Replacing the hosting service provider is achievable and it's basically what you have to do to achieve the "use a third party container to resolve IStartup". I believe the cleanest way to achieve this is to use the IServiceProviderFactory outside of any dependency injection (so removing it from the hosting container) and using it on both the hosting services as well as the application services. Today it's only used on the application services.

This PR attempts to shuttle an IServiceProvider implementation through the hosting services (which is a clever "hack") and then resolve the "last" IServiceProvider instance to resolve the IStartup. This feels *extremely* fragile and can break if the default dependency injection implementation changes to disallow overriding the internal IServiceProvider implementation. On top of that, to make things work well, you really need to preserve the existing hosting services when resolving the IStartup or things may fall apart. It means something like this would just fail:

```
public static IWebHost BuildWebHost(string[] args)
{
    var hostingServices = new ServiceCollection();
    return WebHost.CreateDefaultBuilder(args)
        .UseStartup<Startup>()
        .ConfigureServices(services =>
        {
            services.AddSingleton<IServiceProvider>(hostingServices.BuildServiceProvider());
        })
        .Build();
}
```

If you want to go this route, then we need to define a separate interface that hosting can use to replace the hosting services independently from application services. The WebHostBuilder does something like this:

```
public interface IHostingServiceProviderFactory
{
    IServiceProvider BuildServiceProvider(IServiceCollection services);
}
```

Internally, hosting would resolve this from the container (we might have to build *another* container) and call BuildServiceProvider passing in the initial collection it creates internally to get the final hosting IServiceProvider.

src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs
ENikS: At this point in time ``_applicationServices`` holds reference to externally supplied ``IServiceProvider`` or ``null`` in which case normal flow kicks in and ``IStartup`` is resolved from ``_hostingServiceProvider``.

ENikS:
Your reservations are based on the notion that DI container has to be built before it could be used. Unity and some other containers do not require Build step. Registrations could be adjusted on the fly. Added, removed, replaced. 
So it is possible to use Unity before, during and after WebHost is created.

The behavior of the DI container this solution is based on is **Required** by the [Microsoft.Extensions.DependencyInjection](https://github.com/aspnet/DependencyInjection/blob/dev/src/DI.Specification.Tests/DependencyInjectionSpecificationTests.cs) specification and will not change unless specification itself changes. At which point this whole implementation will stop working anyway.

In normal flow events happen in this sequence:

- Internal DI resolves IStartup
- ConfigureContainer on resolved type is called
- Container factory is called with collection configured in previous step
- Rest of the execution uses resolved container

In proposed PR flow is as follows:

- Builder checks if IServiceProvider is registered
- If provider registered IStartup is resolved from it
- If provider not registered  Internal DI resolves IStartup
- IStartup.ConfigureServices(IServiceCollection services)  (by default when type derived from IStartup )
- Rest of the execution uses provider returned by ConfigureServices (Default behavior when type derived from IStartup )

ENikS:
> On top of that, to make things work well, you really need to preserve the existing hosting services when resolving the IStartup or things may fall apart. It means something like this would just fail:

As long as services are in service collection they will be registered with the container and available during resolution.

davidfowl:
> As long as services are in service collection they will be registered with the container and available during resolution.

Only if you use the service collection provided by ConfigureServices which is required.

ENikS:
Yes, but this solution does not require collection replacement. It uses provided by the framework collection. It other words: It always uses collection provided by ConfigureServices

davidfowl:
@ENikS you say "it" but thats not expressed in any contract. The example I show is completely broken. So you can't just provide any service provider and you have to understand this hidden contract to make any of this work.

ENikS:
What are you talking about, what contract? What am I missing?
The test runs, everything initializes, what exactly are you talking about?

ENikS:
I sent you another [PR](https://github.com/aspnet/Hosting/pull/1322) which solves both of these problems.

ENikS:
I am withdrawing this PR because #1322 provides more complete solution

PR #1322, opened 2018-01-30
Allow external service providers (#550 and #1309)

ENikS:
This proposed solution allows replacement of application as well as framework service providers.
It should satisfy requirements in both #550 and #1309 and might eliminate need for #1060

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -203,7 +203,7 @@ namespace Microsoft.AspNetCore.Hosting
                 throw;
             }
 
-            IServiceProvider GetProviderFromFactory(IServiceCollection colection)
+            IServiceProvider GetProviderFromFactory(IServiceCollection collection)
             {
                 var provider = colection.BuildServiceProvider();
                 var factory = provider.GetService<IServiceProviderFactory<IServiceCollection>>();
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -205,10 +205,10 @@ namespace Microsoft.AspNetCore.Hosting
 
             IServiceProvider GetProviderFromFactory(IServiceCollection collection)
             {
-                var provider = colection.BuildServiceProvider();
+                var provider = collection.BuildServiceProvider();
                 var factory = provider.GetService<IServiceProviderFactory<IServiceCollection>>();
 
-                return factory?.CreateServiceProvider(factory.CreateBuilder(colection)) ?? provider;
+                return factory?.CreateServiceProvider(factory.CreateBuilder(collection)) ?? provider;
             }
         }
 
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -153,7 +153,7 @@ namespace Microsoft.AspNetCore.Hosting
 
             var hostingServices = BuildCommonServices(out var hostingStartupErrors);
             var applicationServices = hostingServices.Clone();
-            var hostingServiceProvider = hostingServices.BuildServiceProvider();
+            var hostingServiceProvider = GetProviderFromFactory(hostingServices) ?? hostingServices.BuildServiceProvider();
 
             if (!_options.SuppressStatusMessages)
             {
@@ -201,6 +201,16 @@ namespace Microsoft.AspNetCore.Hosting
                 // will dispose services that were constructed until the exception was thrown
                 host.Dispose();
                 throw;
+            }
+
+            IServiceProvider GetProviderFromFactory(IServiceCollection colection)
+            {
+                var factory = (IServiceProviderFactory<IServiceCollection>)
+                    colection.LastOrDefault(d => d.ServiceType == typeof(IServiceProviderFactory<IServiceCollection>) &&
+                                                 d.ImplementationInstance != null)?
+                             .ImplementationInstance;
+
+                return factory?.CreateServiceProvider(factory.CreateBuilder(colection));
             }
         }
 
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -212,7 +212,7 @@ namespace Microsoft.AspNetCore.Hosting
                 {
                     using (provider)
                     {
-                        return factory.CreateServiceProvider(factory.CreateBuilder(collection))
+                        return factory.CreateServiceProvider(factory.CreateBuilder(collection));
                     }
                 }
                 
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -153,7 +153,7 @@ namespace Microsoft.AspNetCore.Hosting
 
             var hostingServices = BuildCommonServices(out var hostingStartupErrors);
             var applicationServices = hostingServices.Clone();
-            var hostingServiceProvider = GetProviderFromFactory(hostingServices) ?? hostingServices.BuildServiceProvider();
+            var hostingServiceProvider = GetProviderFromFactory(hostingServices);
 
             if (!_options.SuppressStatusMessages)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -153,7 +153,7 @@ public IWebHost Build()
 
             var hostingServices = BuildCommonServices(out var hostingStartupErrors);
             var applicationServices = hostingServices.Clone();
-            var hostingServiceProvider = hostingServices.BuildServiceProvider();
+            var hostingServiceProvider = GetProviderFromFactory(hostingServices) ?? hostingServices.BuildServiceProvider();
```
ENikS: At present configured factories are ignored and builder uses hardcoded ``hostingServices.BuildServiceProvider()``. Instead it should check if proper factory is registered and if yes, use it to create provider.
Fall back on hardcoded call if nothing is registered.

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
```diff
@@ -71,6 +71,74 @@ public class ThirdPartyContainerStartup
...
+
+        [Fact]
+        public async Task ContainerInstanceCanBeUsedToGetIStartup()
+        {
+            var builder = new WebHostBuilder()
+                .UseStartup<ThirdPartyContainerStartup>()
+                .ConfigureTestServices(services => services.AddSingleton(new SimpleService { Message = "OverridesConfigureServices" }))
+                .ConfigureServices(s => {
+                    s.AddSingleton<IServiceProviderFactory<ThirdPartyContainer>, ThirdPartyContainerServiceProviderFactory>();
+                    s.AddSingleton<IServiceProvider>((a) => s.BuildServiceProvider()); // Provide outside provider instance
```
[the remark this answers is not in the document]
ENikS: No, it is not. But I thought I'd include a demonstration of using these either separately of combined.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: I almost like this change. Lets instead build the service provider another time:

```C#
var hostingServiceProvider = hostingServices.BuildServiceProvider();
hostingServiceProvider = GetProviderFromFactory(hostingServiceProvider) ?? hostingServiceProvider;

IServiceProvider GetProviderFromFactory(IServiceProvider provider)
{
    var factory = provider.GetService<IServiceProviderFactory<IServiceCollection>>();
    return factory?.CreateServiceProvider(factory.CreateBuilder(colection));
}
```

It's a bit more robust.

ENikS: Done

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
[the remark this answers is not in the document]
ENikS: done

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -202,6 +202,14 @@ public IWebHost Build()
                 host.Dispose();
                 throw;
             }
+
+            IServiceProvider GetProviderFromFactory(IServiceCollection colection)
```
davidfowl: nit: spelling colection-> collection

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
```diff
@@ -71,6 +71,56 @@ public class ThirdPartyContainerStartup
...
+                return services;
+            }
+
+            public IServiceProvider CreateServiceProvider(IServiceCollection containerBuilder)
+            {
+                return containerBuilder.BuildServiceProvider();
+            }
+        }
+
+
```
davidfowl: nit: Remove the extra space

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
```diff
@@ -71,6 +71,56 @@ public class ThirdPartyContainerStartup
...
+                    s.AddSingleton<IServiceProviderFactory<IServiceCollection>>(new ExternalContainerFactory()));
+
+            var host = new TestServer(builder);
+
+            var response = await host.CreateClient().GetStringAsync("/");
+
+            Assert.Equal("ApplicationServicesEqual:True", response);
+        }
+
+
```
davidfowl: nit: Remove extra space new line.

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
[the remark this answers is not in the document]
davidfowl: It hasn't been removed yet it seems.

ENikS:
What time zone are you in? Do you work in the middle of the night?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -202,6 +202,14 @@ public IWebHost Build()
                 host.Dispose();
                 throw;
             }
+
+            IServiceProvider GetProviderFromFactory(IServiceCollection collection)
+            {
+                var provider = collection.BuildServiceProvider();
```
davidfowl: Hmm after thinking bit, we might need to dispose the temporary container we created If there's an `IServiceProviderFactory<IServiceCollection>` so this code gets a bit more complicated:

```C#
if (factory != null)
{
     using (provider)
     {
          return factory.CreateServiceProvider(factory.CreateBuilder(collection))
     }
}
return provider;
```

ENikS: You don't want to do that. In case factory is resolved from a type and had any disposable dependencies this code will discard these but it might still be used in the provider.
The provider should die quietly as any other Gen0 variable during next partial collection.

davidfowl: Any service provider we create should be disposed so that services get disposed. The provider may not be collected if there’s a singleton relying on dispose to unroot itself (we had an example recently where a timer was causing problems). The reasons it’s ok here is because we’re creating a whole new container anyways so disposing services after creation should be fine.

ENikS: I understand

test/Microsoft.AspNetCore.TestHost.Tests/TestServerTests.cs
```diff
@@ -71,6 +71,34 @@ public class ThirdPartyContainerStartup
                     $"{ctx.RequestServices.GetRequiredService<SimpleService>().Message}, {ctx.RequestServices.GetRequiredService<TestService>().Message}"));
         }
 
+        [Fact]
+        public async Task ExternalContainerInstanceCanBeUsedForEverything()
```
davidfowl: How does this verify the custom container is being used?

ENikS: How would you like it to be verified? Can you think of all your comments and do them all at once?

ENikS: I am not that familiar with your testing framework, I used one of the tests as example and run it in  debugger to verify. Feel free to change it any way you like.

davidfowl: Heh, sorry, I have a bad habit of re-reviewing things where they are big changes. A good test would be to add a service in the ExternalContainer implementation and make sure it's resolved in the Startup constructor.

I would also not add a test to TestServerTests.cs I would add a test to https://github.com/aspnet/Hosting/blob/dev/test/Microsoft.AspNetCore.Hosting.Tests/WebHostBuilderTests.cs

ENikS: Is this it? Think about it, I am catching flight in a few hours. I have time just for one more change.

davidfowl: I can take over if you want. It's pretty much done.

ENikS: I would appreciate if you do. I am getting low on time.

davidfowl:
Replaced by https://github.com/aspnet/Hosting/pull/1325

PR #1325, opened 2018-01-31
Allow overriding the hosting service provider 

davidfowl:
Replaces https://github.com/aspnet/Hosting/pull/1322

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
@@ -143,9 +143,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                     {
                         // Get the default factory
                         var serviceProviderFactory = HostingServiceProvider.GetRequiredService<IServiceProviderFactory<IServiceCollection>>();
-
-                        // Don't bother calling CreateBuilder since it just returns the default service collection
-                        applicationServiceProvider = serviceProviderFactory.CreateServiceProvider(services);
+                        var builder = serviceProviderFactory.CreateBuilder(services);
+                        applicationServiceProvider = serviceProviderFactory.CreateServiceProvider(builder);
                     }
 
                     return applicationServiceProvider ?? services.BuildServiceProvider();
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -202,6 +202,22 @@ public IWebHost Build()
                 host.Dispose();
                 throw;
             }
+
+            IServiceProvider GetProviderFromFactory(IServiceCollection collection)
```
Tratcher: Only used once and not complex, inline.

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
```diff
@@ -143,9 +143,8 @@ IServiceProvider ConfigureServicesWithContainerConfiguration(IServiceCollection
                     {
                         // Get the default factory
                         var serviceProviderFactory = HostingServiceProvider.GetRequiredService<IServiceProviderFactory<IServiceCollection>>();
-
-                        // Don't bother calling CreateBuilder since it just returns the default service collection
-                        applicationServiceProvider = serviceProviderFactory.CreateServiceProvider(services);
+                        var builder = serviceProviderFactory.CreateBuilder(services);
```
ENikS: Is it guaranteed that factory exists?

davidfowl: This has been the case since 2.0.  There’s always one registered. Unless somebody removes it.

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
davidfowl: I think it should remain a method. It's not like it's 2 lines of code. It's a few.

src/Microsoft.AspNetCore.Hosting/Internal/StartupLoader.cs
Tratcher: *shrug* it's a weird use of embedded methods.

PR #1327, opened 2018-02-02
Don't throw in Nginx deployer dispose

BrennanConroy:


ryanbrandenburg:
This effects 2.1 too. We might want to port this back to 2.1 to ensure to ensure that whatever the over-arching fix for this issue is works there too.

BrennanConroy:
This is in 2.1...?

ryanbrandenburg:
http://aspnetci/viewLog.html?buildId=399313&tab=buildResultsDiv&buildTypeId=Releases_21Public_UbuntuUniverse

PR #1337, opened 2018-02-28
Use dispose pattern in BackgroundService

twsouthwick:
BackgroundService implements IDiposable, but doesn't implement the dispose pattern. Since this type is new for 2.1, this should be fixed before release.

### DIFF
--- a/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
+++ b/src/Microsoft.Extensions.Hosting.Abstractions/BackgroundService.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Extensions.Hosting
     /// </summary>
     public abstract class BackgroundService : IHostedService, IDisposable
     {
+        private bool disposed = false;
         private Task _executingTask;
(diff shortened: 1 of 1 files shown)


Tratcher:
👎 This pattern is overused. It's primarily intended for native resources with finalizers.

twsouthwick:
That may be true, but it's still the recommended practice when implementing `IDisposable` on a base class: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implementing-the-dispose-pattern-for-a-base-class.

poke:
I don’t believe it’s worth adding this. This pattern is only needed when you have unmanaged resources *and* when you also have a finalizer that calls `Dispose(false)`. Since the majority of the background threads will likely have *no* unmanaged resources whatsoever, I don’t think we need to follow the full pattern here. If there are background thread implementations later that do have unmanaged resources, they can always implement the pattern themselves.

Eilon:
@Tratcher / @muratg - it looks like the conclusion is that we can close this PR?

muratg:
@twsouthwick Thanks for the PR!  We don't believe this adds enough value to accept the change.

PR #1342, opened 2018-03-05
Convert TestHost ResponseStream to use Pipes.

Tratcher:
TestHost has always used an in-memory write-through buffer stream so that a test app could write to the response body and the test code could consume that data asynchronously. Basically this was pipes five years before pipes was a thing.  Now I've converted it over to pipes as a way of testing the pipes API and functionality and reducing the internal code complexity.

Open issues:
https://github.com/dotnet/corefx/issues/27732 - FlushAsync completes ReadAsync with no data
https://github.com/aspnet/BuildTools/issues/592 - ApiCheck can't handle the new Memory<T> types.
Asymmetry - PipeWriter has a sync Write extension, but PipeReader does not have a sync Read extension.

### DIFF
--- a/src/Microsoft.AspNetCore.TestHost/ResponseStream.cs
+++ b/src/Microsoft.AspNetCore.TestHost/ResponseStream.cs
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Concurrent;
+using System.Buffers;
+using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.IO;
+using System.IO.Pipelines;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.TestHost/ResponseStream.cs
```diff
@@ -93,126 +77,68 @@ public override void SetLength(long value)
...
-                            {
-                                CheckAborted();
-                                // Graceful close
-                                return totalRead;
-                            }
-                            WaitForDataAsync().Wait();
-                        }
-                        _topBuffer = new ArraySegment<byte>(topBuffer);
+                        var actual = Math.Min(readableBuffer.Length, count);
+                        readableBuffer = readableBuffer.Slice(0, actual);
```
jkotalik: This is a bug in both of our loops but you don't use the offset parameter at all.

src/Microsoft.AspNetCore.TestHost/ResponseStream.cs
```diff
@@ -93,126 +77,68 @@ public override void SetLength(long value)
...
-                        _topBuffer.Count - actualCount);
-                    totalRead += actualCount;
-                    offset += actualCount;
-                    count -= actualCount;
+
+                    cancellationToken.ThrowIfCancellationRequested();
+                    Debug.Assert(!result.IsCanceled); // It should only be canceled by cancellationToken.
+
+                    // Try again. TODO: I shouldn't need to do this, there wasn't any data.
+                    _pipe.Reader.AdvanceTo(readableBuffer.End, readableBuffer.End);
```
jkotalik: Should be fixed by the next pipelines change, correct @davidfowl  @pakrym

davidfowl: No, the reader still wakes up for 0 byte reads. The thing that is fixed right now is completing the pipe without advancing, that works

davidfowl: https://github.com/dotnet/corefx/issues/27732

src/Microsoft.AspNetCore.TestHost/ResponseStream.cs
```diff
@@ -348,13 +243,9 @@ internal void Complete()
...
-            // Prevent race with WaitForDataAsync
-            lock (_signalReadLock)
-            {
-                // Throw for further writes, but not reads.  Allow reads to drain the buffered data and then return 0 for further reads.
-                _complete = true;
-                _readWaitingForData.TrySetResult(null);
-            }
+            // Throw for further writes, but not reads.  Allow reads to drain the buffered data and then return 0 for further reads.
+            _complete = true;
+            _pipe.Writer.Complete();
```
davidfowl: When do you complete the reader?

Tratcher:
Related question. Should the PipeReader and PipeWriter have a property that tells me if they're complete? Is the only test to call them and catch the exception?

davidfowl:
> Related question. Should the PipeReader and PipeWriter have a property that tells me if they're complete? Is the only test to call them and catch the exception?

That's not really the usage pattern and that in itself could be a bit racy. The ideal usage pattern is an async loop where there is a single reader and single writer. At the each of reading and writing you complete the appropriate side of the pipe.

In this case of wrapping a stream, you're acting on external method calls so it's your job to serialize those into the pipe itself. The _complete boolean would be used here to determine when the request is complete. If you don't want to throw, you need your own state (which you conveniently have 😄 )

Tratcher:
I intentionally do not complete the reader when the request 'completes'. The test still needs to read from the body after the app finishes the request. I've added a complete in ReadAsync if it reports complete.

src/Microsoft.AspNetCore.TestHost/ResponseStream.cs
Tratcher: Added in ReadAsync

src/Microsoft.AspNetCore.TestHost/ResponseStream.cs
Tratcher: Fixed, see the next line.

jkotalik:
Microsoft.AspNetCore.TestHost.ClientHandlerTests.ClientCancellationAborts seems to be failing still.

PR #1351, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


Eilon:
Why was a baseline removed?

Tratcher:
It wasn't removed, but it was trimmed of types from the Internal namespace. Those weren't supposed to be listed anyways right?

Eilon:
@Tratcher oh ok removing the .Internal is certainly fine. But yeah I wouldn't have thought they'd be in there to begin with. @javiercn any idea on that?

javiercn:
We need to keep .Internal for patch releases. 2.0.0 to 2.0.1 we should not do any public API breaking change even in the .Internal namespace. For that, we need to capture .Internal in the baselines. The tool has a mode to exclude things on the .Internal namespace that can be use for 2.0->2.1 type of changes.

Does that make sense?

Eilon:
@javiercn ah ok should we use that switch here then? Or, @ryanbrandenburg , is it being used already?

ryanbrandenburg:
I did not set any switches when doing this, I just did the "normal" generation. @javiercn what is the switch we need to turn on for this 2.0->2.1 move?

javiercn:
@ryanbrandenburg The normal generation should be fine. It's when checking that we should not pass --exclude-public-internal

ryanbrandenburg:
Best guess is @BrennanConroy is the one to review this. Assign whoever is a better choice if it's not you.

ryanbrandenburg:
The failure in appveyor is an unrelated unit test.

src/Microsoft.AspNetCore.Server.IntegrationTesting/baseline.netcore.json
```diff
@@ -1,5 +1,5 @@
 {
-  "AssemblyIdentity": "Microsoft.AspNetCore.Server.IntegrationTesting, Version=0.4.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
+  "AssemblyIdentity": "Microsoft.AspNetCore.Server.IntegrationTesting, Version=0.4.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
```
Tratcher: Should this be excluded as a prototype?

PR #1360, opened 2018-03-20
Make the RequestServicesContainerMiddleware thinner

davidfowl:
- Today the request services middleware is responsible for making sure there are request scoped sevices.
This PR tries introduces some breaking changes that are hopefully acceptable in order to gain some performance.
- Here are the assumptions this PR makes:
 - Since this middleware is first in the pipeline, the only thing that can
  set a default service provider would be the server itself. Since we have no servers that do that
  I removed that code that tries to noop if there's an existing service provider.
 - This PR no longer restores the previous service provider feature since it gets replaced every request
 anyways. Kestrel also clears out the feature on each request so it shouldn't be a problem (in theory).
 Once again, since this middleware is first, it is the last thing that runs before the server re-gains
 control on the way out so there's no need to restore anything.
 - We use the RegisterForDispose method to dispose of the IServiceProvider instead of doing it inline.

@Tratcher come at me bro 😄.

```
02:50:02.981] Using worker Wrk
[02:50:03.633] Running session '86d93bd805864be48327ffc1c344fb58' with description ''
[02:50:03.633] Starting scenario Plaintext on benchmark server...
[02:50:12.412] Warmup
[02:50:12.428] Starting scenario Plaintext on benchmark client...
[02:50:28.414] Scenario Plaintext completed on benchmark client
[02:50:28.417] Stopping scenario Plaintext on benchmark client...
[02:50:28.447] Starting scenario Plaintext on benchmark client...
[02:50:44.251] Scenario Plaintext completed on benchmark client
[02:50:44.251] Stopping scenario Plaintext on benchmark client...
[02:50:44.348] RequestsPerSecond:           1921644
[02:50:44.348] Latency on load (ms):        2.2
[02:50:44.348] Max CPU (%):                 92
[02:50:44.348] WorkingSet (MB):             374
[02:50:44.348] Startup Main (ms):           373
[02:50:44.348] First Request (ms):          149.3
[02:50:44.349] Latency (ms):                0.7
[02:50:44.349] Socket Errors:               0
[02:50:44.349] Bad Responses:               0
[02:50:44.349] SDK:                         2.2.0-preview1-007522
[02:50:44.349] Runtime:                     2.1.0-preview2-26314-02
[02:50:44.349] ASP.NET Core:                2.1.0-preview3-30384
```

/cc @DamianEdwards @benaadams @sebastienros 

Putting on my pedantic hat here's what can go wrong:
- If somebody replaces the IApplicationBuilderFactory, they can inject middleware before the RequestServicesContainerMiddleware middleware. They would still see the service providers feature added by the middleware.
- If somebody removes the IStartupFilter that adds this middleware and does their own thing, the middleware won't clean itself up so middleware that run after it will observe the service providers feature.
- If there's a bug in RegisterForDispose, then service providers won't get disposed. Well, and many other things that need to get disposed 😄.

PS: We have no tests for the things I removed. Ooops we have no tests at all https://github.com/aspnet/Hosting/search?utf8=%E2%9C%93&q=AutoRequestServicesStartupFilter&type=

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
@@ -30,34 +30,30 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             _scopeFactory = scopeFactory;
         }
 
-        public async Task Invoke(HttpContext httpContext)
+        public Task Invoke(HttpContext httpContext)
         {
             Debug.Assert(httpContext != null);
 
             // local cache for virtual disptach result
             var features = httpContext.Features;
-            var existingFeature = features.Get<IServiceProvidersFeature>();
 
-            // All done if RequestServices is set
-            if (existingFeature?.RequestServices != null)
+            var replacementFeature = new RequestServicesFeature(httpContext, _scopeFactory);
+
+            features.Set<IServiceProvidersFeature>(replacementFeature);
+            var task = _next.Invoke(httpContext);
+
+#if NETCOREAPP2_1
+            if (!task.IsCompletedSuccessfully)
+#else
+            if (task.Status != TaskStatus.RanToCompletion)
+#endif
             {
-                await _next.Invoke(httpContext);
-                return;
+                return InvokeAsyncAwaited(task);
             }
 
-            var replacementFeature = new RequestServicesFeature(_scopeFactory);
+            return Task.CompletedTask;
 
-            try
-            {
-                features.Set<IServiceProvidersFeature>(replacementFeature);
-                await _next.Invoke(httpContext);
-            }
-            finally
-            {
-                replacementFeature.Dispose();
-                // Restore previous feature state
-                features.Set(existingFeature);
-            }
+            async Task InvokeAsyncAwaited(Task resultTask) => await resultTask;
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
@@ -40,20 +40,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             var replacementFeature = new RequestServicesFeature(httpContext, _scopeFactory);
 
             features.Set<IServiceProvidersFeature>(replacementFeature);
-            var task = _next.Invoke(httpContext);
-
-#if NETCOREAPP2_1
-            if (!task.IsCompletedSuccessfully)
-#else
-            if (task.Status != TaskStatus.RanToCompletion)
-#endif
-            {
-                return InvokeAsyncAwaited(task);
-            }
-
-            return Task.CompletedTask;
-
-            async Task InvokeAsyncAwaited(Task resultTask) => await resultTask;
+            return _next.Invoke(httpContext);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


benaadams:
This would make this change no longer necessary? https://github.com/aspnet/KestrelHttpServer/pull/2355

benaadams:
The pattern currently in Hosting has been a perf bugbear of mine for ages and is hard to optimize (get, bunch of checking, swap, restore). With this new pattern, can already see a few potential improvements on the improvements... 😄

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -30,34 +30,30 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFac
...
-            if (existingFeature?.RequestServices != null)
-            {
-                await _next.Invoke(httpContext);
-                return;
-            }
+            var replacementFeature = new RequestServicesFeature(httpContext, _scopeFactory);
 
-            var replacementFeature = new RequestServicesFeature(_scopeFactory);
+            features.Set<IServiceProvidersFeature>(replacementFeature);
+            var task = _next.Invoke(httpContext);
```
Tratcher: If you're not going to do any cleanup afterwards then why not return immediately here?

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesFeature.cs
```diff
@@ -30,6 +33,7 @@ public IServiceProvider RequestServices
                     _scope = _scopeFactory.CreateScope();
                     _requestServices = _scope.ServiceProvider;
                     _requestServicesSet = true;
+                    _context.Response.RegisterForDispose(this);
```
Tratcher: This does create one new possibility for a leak. Consider if someone called RequestServices for the first time inside of their RegisterForDispose/OnCompleted callback. It would be too late to register an additional disposal. 

Not a huge risk, the contract of OnCompleted tells you not to touch HttpContext anyways.

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
davidfowl: Yea this is true. I was shuffling the code so much before I don't even need the other part.

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesFeature.cs
davidfowl: Yep agreed. Not sure that matters.

davidfowl: I could move RegisterForDispose first, that way if it blows up the scope would never be created in the first place.

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
```diff
@@ -30,34 +30,17 @@ public RequestServicesContainerMiddleware(RequestDelegate next, IServiceScopeFac
...
             var features = httpContext.Features;
-            var existingFeature = features.Get<IServiceProvidersFeature>();
 
-            // All done if RequestServices is set
-            if (existingFeature?.RequestServices != null)
-            {
-                await _next.Invoke(httpContext);
-                return;
-            }
+            var replacementFeature = new RequestServicesFeature(httpContext, _scopeFactory);
```
benaadams: nit: don't call this `replacementFeature` as its not replacing anything now

src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesFeature.cs
khellang: Is this what caused `RequestServices` to be disposed and reset to `null` before the `OnStarting` callbacks are invoked?

khellang: Umm. Never mind. I see this was done for 2.1. After updating from 2.0.8 to 2.1-rc1-final, I no longer see `RequestServices` being `null` inside `OnStarting`. Weird 🤔

Issue #1347, opened 2018-03-09
HostLifetime in its current form doesn't work for WindowsServices


glennc:
Creating a HostLifetime to work with ServiceBase doesn't currently work.

We need to re-evaluate what we want to do here before preview2. This was one of the major reasons we did HostLifetime. If it isn't going to work we need to change it.

@davidfowl @Tratcher @muratg

muratg:
Do we know why already (and just need to design a solution), or do we need to investigate further?

glennc:
The Run method of ServiceBase blocks and our current surface area doesn't really let you handle that in a 100% correct way. You can kind of get it to work but it's not really the way you want it to be.

glennc:
I can put together some code with comments that shows the problems.

Rinsen:
Do you have any sample on that "kind of get it to work" @glennc ? I tried to get the sample to work but failed...

Tratcher:
@Rinsen putting a Task.Run around the Run call here is the main workaround. https://github.com/aspnet/Hosting/blob/10a56f198169bc75bd07efbfb7a8d5a04630aa0a/samples/GenericHostSample/ServiceBaseLifetime.cs#L36

muratg:
Any design work left on this one?  Also, what's the cost?

Tratcher:
The PR is waiting for @davidfowl's review. We'll see if he accepts the modified design.

Rinsen:
Awesome! :) 

Thanks @Tratcher and I am looking forward to this feature in the 2.1 release! So very good to have the full stack with loggers and all in a Windows Service for background tasks 👍

PR #1361, opened 2018-03-20
Refactor Generic Host lifetimes to work better with ServiceBase 

Tratcher:
#1347
The RegisterDelayStartCallbackAsync callback is inlined to work around the ServiceBase.Run blocking call. 
RegisterStopCallback goes away, it only ever called StopApplication.
ProcessLifetime goes away, ConsoleLifetime becomes the default.


Tratcher:
@glennc one funny thing I noticed when debugging is that OnStop gets called 3 times, but it's fine because the CTS in StopApplication prevents any recursion.
1. OnStop called by the service.
2. OnStop => StopApplication => Stop => OnStop
3. Host.StopAsync => Stop => OnStop

davidfowl:
StartAsync cannot block. The usage pattern is too different from the other usage pattern. We need to pick one way and stick with it.

Tratcher:
Easy enough, I forced ServiceBase.Run onto a background thread.

samples/GenericHostSample/ServiceBaseLifetime.cs
```diff
@@ -23,23 +23,25 @@ public static Task RunAsServiceAsync(this IHostBuilder hostBuilder, Cancellation
...
 
-            Run(this);
-        }
+            ApplicationLifetime.ApplicationStopping.Register(Stop);
 
-        public void RegisterStopCallback(Action<object> callback, object state)
-        {
-            _stopCallback = callback ?? throw new ArgumentNullException(nameof(callback));
-            _stopState = state;
+            Task.Run(() => Run(this)); // Otherwise this would block and prevent Start from finishing.
```
davidfowl: Spawn a real thread so we don't block a thread pool thread.

Tratcher: I uncovered a design flaw when testing this. RegisterDelayStartCallbackAsync has no abort mechanic. If you try to run your service app outside of a service Run(this) pops up a dialog box to tell you what you did wrong, but Run itself doesn't throw, it exits early. OnStart is never called so the delay start callback is never called, and the app hangs in StartAsync.

Tratcher:
Redesigned and Updated. I got rid of all the callbacks.

test/Microsoft.Extensions.Hosting.Tests/HostTests.cs
```diff
@@ -414,7 +353,6 @@ public void HostCanBeStarted()
 
                 lifetime = (FakeHostLifetime)host.Services.GetRequiredService<IHostLifetime>();
                 Assert.Equal(1, lifetime.StartCount);
-                Assert.Equal(1, lifetime.StoppingCount);
```
davidfowl: Why did this change?

Tratcher: Stopping was one of the test callbacks we removed.

davidfowl: What do you mean? The test itself changed?

Tratcher: I merged RegisterStopCallback and RegisterDelayStartCallback into the new WaitForStartAsync so there's one less event to mock in the tests.

davidfowl: Oh, I thought this was IApplicationLifetime

PR #1362, opened 2018-03-21
Make the RequestServicesContainerMiddleware thinner (#1360)

davidfowl:
- Today the request services middleware is responsible for making sure there are request scoped services.
This PR tries introduces some breaking changes that are hopefully acceptable in order to gain some performance.
- Here are the assumptions this PR makes:
 - Since this middleware is first in the pipeline, the only thing that can
  set a default service provider would be the server itself. Since we have no servers that do that
  I removed that code that tries to noop if there's an existing service provider.
 - This PR no longer restores the previous service provider feature since it gets replaced every request
 anyways. Kestrel also clears out the feature on each request so it shouldn't be a problem (in theory).
 Once again, since this middleware is first, it is the last thing that runs before the server re-gains
 control on the way out so there's no need to restore anything.
 - We use the RegisterForDispose method to dispose of the IServiceProvider instead of doing it inline.


davidfowl:
@Tratcher putting this in release/2.1

PR #1368, opened 2018-03-28
Support inprocess hosting in IISExpressDeployer via switching the HostingModel property.

jkotalik:
I added the HostingModel property a while ago, but it was never consumed. I will make reaction PRs in ServerTests and MusicStore.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
@@ -4,10 +4,12 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 using Microsoft.AspNetCore.Server.IntegrationTesting.Common;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


jkotalik:
@Tratcher Microsoft.AspNetCore.TestHost.ClientHandlerTests.ClientDisposalCloses failed both on AppVeyor and Travis; Managed HttpClient?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -294,5 +302,20 @@ public override void Dispose()
...
+        // Transforms the web.config file to include the hostingModel="inprocess" element
+        // and adds the server type = Microsoft.AspNetServer.IIS such that Kestrel isn't added again in ServerTests
+        private void ModifyWebConfigToInProcess()
+        {
+            var webConfigFile = $"{DeploymentParameters.PublishedApplicationRootPath}/web.config";
+            var config = XDocument.Load(webConfigFile);
+            var element = config.Descendants("aspNetCore").FirstOrDefault();
+            element.SetAttributeValue("hostingModel", "inprocess");
+
+            var argsValue = element.Attribute("arguments").Value;
```
pakrym: Why add a server argument?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -294,5 +302,20 @@ public override void Dispose()
...
+            var newArgs = $"{argsValue} --server Microsoft.AspNetCore.Server.IIS";
```
Tratcher: 👀 This is the same as https://github.com/aspnet/IISIntegration/issues/729?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
jkotalik: You are right, this should be part of the test itself.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
jkotalik: Yeah, but this problem should be handled at the test level, rather than the server level.

PR #1374, opened 2018-04-10
Clean up pipes usage

Tratcher:
React to several pipes usability issues that have since been fixed.


Tratcher:
I figured out the issue with the flaky test #1379. It was only failing on Appveyor. Appveyor runs machines with very few cores and xunit limits the number of threads based on the number of cores. This test was blocking two different threads and waiting for a cancellation callback from a third thread. This has never failed locally or on our CI, but always failed on Appveyor.

Issue #1329, opened 2018-02-13
Console host does not shut down gracefully on sigterm (linux)


robbieknuth:
I've been testing out the new generic hosting model in preparation for 2.1, hoping to get rid of my custom shutdown code. However, the generic console host does not exit gracefully on sigterm in the same way the webhost does. Meaning none of the the `StopAsync` code of my `IHostedService`s are called. It's not clear to me why they behave differently... some interaction between the different lifetimes I suppose.

1) unpack zip [TestGenericShutdown.zip](https://github.com/aspnet/Hosting/files/1721499/TestGenericShutdown.zip)
2) from solution root, run dotnet restore
3) cd to project directory "cd TestGenericShutdown/"
4) Run the program:

-- for console host behavior:
> dotnet run -- console

-- for web host behavior

> dotnet run -- web

5) Send the program a SIGTERM by finding out the PID and doing `kill -SIGTERM PID`

I would expect the behavior between these two to be identical, but they are not.

Web host behavior:

>Hosting environment: Production
>Content root path: /MY/MACHINE
>Now listening on: http://localhost:5000
>Application started. Press Ctrl+C to shut down.
>info: TestGenericShutdown.DummyBackgroundTaskOne[0]
>      DummyBackgroundTaskOne is running.
>info: TestGenericShutdown.DummyBackgroundTaskTwo[0]
>      DummyBackgroundTaskTwo is running.
>
> `SIGTERM SENT`
>
>Application is shutting down...
>info: TestGenericShutdown.DummyBackgroundTaskOne[0]
>      DummyBackgroundTaskOne is stopping.
>info: TestGenericShutdown.DummyBackgroundTaskOne[0]
>      DummyBackgroundTaskOne is stopped.
>info: TestGenericShutdown.DummyBackgroundTaskTwo[0]
>      DummyBackgroundTaskTwo is stopping.
>info: TestGenericShutdown.DummyBackgroundTaskTwo[0]
>      DummyBackgroundTaskTwo is stopped.
>Web host exiting gracefully.

Console host behavior:
>Application started. Press Ctrl+C to shut down.
>Hosting environment: Production
>Content root path: /MY/MACHINE
>info: TestGenericShutdown.DummyBackgroundTaskTwo[0]
>      DummyBackgroundTaskTwo is running.
>info: TestGenericShutdown.DummyBackgroundTaskOne[0]
>      DummyBackgroundTaskOne is running.
>
> `SIGTERM SENT`

[...]

PR #1376, opened 2018-04-12
Delay process exit in ConsoleLifetime

Tratcher:
#1329 ConsoleLifetime listens for the ProcessExit event but does not block it to allow for graceful shutdown. CTL+C works because we can cancel that event.

@robbieknuth @xqrzd I was able to reproduce the same issue on Windows by closing the console window.

Solution: Block ProcessExit until ConsoleLifetime is disposed (when the rest of the Host is disposed). This is pretty much the same as what WebHost does.
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs#L43-L50

### DIFF
--- a/samples/GenericHostSample/MyServiceB.cs
+++ b/samples/GenericHostSample/MyServiceB.cs
@@ -5,7 +5,7 @@ using Microsoft.Extensions.Hosting;
 
 namespace GenericHostSample
 {
-    public class MyServiceB : IHostedService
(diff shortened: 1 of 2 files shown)


Tratcher:
A functional test would be pretty involved. WebHost has some similar tests but they were so flaky we disabled them.

PR #1378, opened 2018-04-13
Check for non-null RequestServices

### LINKED ISSUE #1377: Thinner RequestServicesMiddleware (#1360) removes ability to sub out request services
This change [to RequestServicesMiddleware](https://github.com/aspnet/Hosting/commit/b3625b8173b19935e22b4fe5ac0ea64edd934bd4#diff-682be05953da51fb7d5c7ed8130f8a5b) removes a check to see if the `IServiceProvidersFeature` is null and only put it in if it's required.

Unfortunately, the RequestServicesMiddleware _also_ takes the `IServiceScopeFactory` in during its constructor rather than during `Invoke`, which, for multitenant support, is too early - you can't identify a tenant that early in the pipeline since you don't have a context, which means you can't provide info to the service scope factory on which tenant needs a scope.

As a concrete example, [the Autofac ASP.NET Core multitenant integration](https://github.com/autofac/Autofac.AspNetCore.Multitenant) intentionally inserts itself in front of the RequestServicesMiddleware so it can put its own request services in place first, assuming the RequestServicesMiddleware will later effectively become a no-op.

This is a pretty painful breaking change to 2.1 - would you be open to a PR to put the null check back in RequestServicesMiddlware? Alternatively, is there a way to remove the default RequestServicesMiddleware altogether?

davidfowl:
Fixes #1377

/cc @tillig

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
@@ -34,14 +34,16 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             Debug.Assert(httpContext != null);
 
+            var features = httpContext.Features;
+            var servicesFeature = features.Get<IServiceProvidersFeature>();
+
             // All done if RequestServices is set
-            if (httpContext.RequestServices != null)
+            if (servicesFeature?.RequestServices != null)
             {
                 return _next.Invoke(httpContext);
             }
 
-            var servicesFeature = new RequestServicesFeature(httpContext, _scopeFactory);
-            httpContext.Features.Set<IServiceProvidersFeature>(servicesFeature);
+            features.Set<IServiceProvidersFeature>(new RequestServicesFeature(httpContext, _scopeFactory));
             return _next.Invoke(httpContext);
         }
     }
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestServicesContainerMiddleware.cs
@@ -34,12 +34,14 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         {
             Debug.Assert(httpContext != null);
 
-            // local cache for virtual disptach result
-            var features = httpContext.Features;
+            // All done if RequestServices is set
+            if (httpContext.RequestServices != null)
+            {
+                return _next.Invoke(httpContext);
+            }
 
             var servicesFeature = new RequestServicesFeature(httpContext, _scopeFactory);
-
-            features.Set<IServiceProvidersFeature>(servicesFeature);
+            httpContext.Features.Set<IServiceProvidersFeature>(servicesFeature);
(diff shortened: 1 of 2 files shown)


tillig:
I think this should work. It'll let me put in my own `RequestServicesFeature` and when the `httpContext.RequestServices` gets called in that null check is when the actual per-tenant scope can be created without getting replaced later. Thanks for the speedy turnaround and help on this!

sebastienros:
| Description |       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| ----------- | --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
|      Before | 1,859,607 |      90 |         356 |               4.2 |          433 |                  0 |            0 |
|       After | 1,755,683 |      92 |         373 |               3.6 |          421 |                  0 |            0 |

davidfowl:
Hmm let me see if we can do something about the performance hit.

tillig:
Stab in the dark, but it may be cheaper to `httpContext.Features.Get<IServiceProvidersFeature>()` and check that for null instead of resolving the whole `httpContext.RequestServices`. `RequestServices` will get lazy-instantiated anyway; the important part of the check (for my use case) is not replacing the existing feature if it's already there.

tillig:
You could possibly squeeze a tiny bit more out by checking if the feature is null and not calling feature.RequestServices at all.

davidfowl:
> You could possibly squeeze a tiny bit more out by checking if the feature is null and not calling feature.RequestServices at all.

C# does that for me. feature?.RequestServices turns into feature != null && feature.RequestServices != null and because of left to right lazy evaluation it should already have that effect.

Also in the benchmark case, we always hit that if check and it's always false so the extra null checking wouldn't help. In the 99% case, nobody has injected services before the request services middleware. In practice they just overwrite it afterwards but early enough that most things haven't been resolved yet so it doesn't end up mattering. This is what orchard does and all of the other multi tenancy systems out there do.

davidfowl:
Thinking about it more, why can't you just inject a Startup filter that runs after the middleware but replaces the feature? It would be early enough that nobody has changed it.

davidfowl:
@tillig actually the more I think about it, the more I think just running after is better all up. Why not just do that? Why do you need to run before this middleware?

tillig:
If you can:

- Tell me how to detect where in the list of startup filters I should be inserting my override
- Promise to never actually call `RequestServices` on the feature in your middleware (so no actual scope is created)
- Keep both of those things as part of the effective contract of how `RequestServicesMiddleware` works

...then I'm cool with that. Just seems like there's a bad assumption being made in here that other places don't make. Like how you can call `services.AddSomething()` and there's work in there to ensure whatever services being added aren't duplicated. Sort of a "do no harm" philosophy being ignored.

But, like I said, if I can hook in a different way, that's cool. If it's granular enough, I could also totally remove the default request services startup filter and bypass your middleware entirely, which is also absolutely acceptable for my scenario.

Tratcher:
Some of the other startup filters like IISIntegration with Windows Auth may resolve services.

davidfowl:
@tillig yeah, I think we'll keep this fix since it is a breaking change.

Issue #1296, opened 2017-12-12
SelfHostDeployer still produces AddressInUseException


ryanbrandenburg:
We run into [this](http://aspnetci/viewLog.html?buildId=366598&tab=buildResultsDiv&buildTypeId=Lite_MvcTest#testNameId7765957466554203500) kind of thing still. I believe previously @anurse added extra logging to this scenario.

muratg:
@anurse what's the work proposed here?

anurse:
Depends on if this is still a persistent problem on the CI. In order to figure out why it's still happening, we'd either need more logging or to make our tests way more defensive about addresses being in use.

ryanbrandenburg:
It is [still](http://aspnetci/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=397530&_focus=108053) happening on the CI. When you say more logging you mean in the SelfHostDeployer right?

anurse:
Yes

muratg:
@JunTaoLuo Whenever you have some free time, can you take a look?

ryanbrandenburg:
[Here's](http://aspnetci/httpAuth/app/rest/builds?locator=tag:PortFail,sinceDate:201710010T000000-0001&fields=build(startDate,buildTypeId,status,branchName,webUrl,tags)) a non-exhaustive (but reasonably through) list of the builds that were failed by this. As you can see it happens about once or more per-day.

ryanbrandenburg:
Any movement on this? It fails 2+ times a day.

JunTaoLuo:
I'll make some time to take a look at it this week.

mikeharder:
SelfHostDeployer currently works as follows:

1. Bind a socket to port `0`
1. Query the socket to determine the assigned port (say `56789`)
1. Dispose the socket
1. Start another process, and tell it to use the earlier assigned port `56789`
1. Other process binds to assigned port `56789`

This design was based on the assumption that the OS will not reuse the assigned port between step 3 and step 5.  This assumption is usually true, but sometimes the OS does reuse the port, which causes one of the external processes to fail to bind.  See repro at https://github.com/mikeharder/GetNextPort, which fails after a few seconds on my machine.

The most reliable workaround is to always start the external process on port `0`, then have the external process communicate back the assigned port.  SelfHostDeployer already has logic for this: 

[...]

PR #1383, opened 2018-04-16
 Change SelfHostDeployer to use dynamic ports by default

mikeharder:
* Should significantly reduce flaky test failures due to AddressInUse exceptions
* Addresses https://github.com/aspnet/Hosting/issues/1296

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
@@ -25,10 +25,15 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting.Common
             {
                 if (statusMessagesEnabled)
                 {
+                    // Most functional tests use this codepath and should directly
+                    // bind to dynamic port "0" and scrape the assigned port
+                    // from the status message, which should be 100% reliable.
                     return new UriBuilder("http", "127.0.0.1", 0).Uri;
                 }
                 else
                 {
+                    // If status messages are disabled, the less reliable GetNextPort()
+                    // must be used.
                     return new UriBuilder("http", "localhost", GetNextPort()).Uri;
                 }
             }
@@ -37,16 +42,31 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting.Common
                 var uriHint = new Uri(hint);
                 if (uriHint.Port == 0)
                 {
+                    // Only a few tests use this codepath, so it's fine to use the less reliable
+                    // GetNextPort() for simplicity.  These tests could be improved to bind 
+                    // to dynamic port "0" (when status messages are enabled) but the hostname
+                    // would need to be "127.0.0.1" or "[::1]".  Binding to dynamic port "0" on
+                    // "localhost" is not supported in Kestrel.
                     return new UriBuilder(uriHint) { Port = GetNextPort() }.Uri;
                 }
                 else
                 {
+                    // If the hint contains a specific port, return it unchanged.
                     return uriHint;
                 }
             }
         }
 
         // Copied from https://github.com/aspnet/KestrelHttpServer/blob/47f1db20e063c2da75d9d89653fad4eafe24446c/test/Microsoft.AspNetCore.Server.Kestrel.FunctionalTests/AddressRegistrationTests.cs#L508
+        //
+        // This method is an attempt to safely get a free port from the OS.  Most of the time,
+        // when binding to dynamic port "0" the OS increments the assigned port, so it's safe
+        // to re-use the assigned port in another process.  However, occasionally the OS will reuse
+        // a recently assigned port instead of incrementing, which causes flaky tests with AddressInUse
+        // exceptions.  This method should only be used when the application itself cannot use
+        // dynamic port "0" (e.g. IISExpress).  Most functional tests using raw Kestrel
+        // (with status messages enabled) should directly bind to dynamic port "0" and scrape 
+        // the assigned port from the status message, which should be 100% reliable.
         public static int GetNextPort()
         {
             using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -11,39 +11,68 @@ public static class TestUriHelper
...
             {
-                return BuildTestUri();
+                if (statusMessagesEnabled)
+                {
+                    // Most functional tests use this codepath and should directly bind to dynamic port "0" and scrape
+                    // the assigned port from the status message, which should be 100% reliable since the port is bound
+                    // once and never released.  Binding to dynamic port "0" on "localhost" (both IPv4 and IPv6) is not
+                    // supported, so the port is only bound on "127.0.0.1" (IPv4).  If a test explicitly requires IPv6,
+                    // it should provide a hint URL with "localhost" (IPv4 and IPv6) or "[::1]" (IPv6-only).
+                    return new UriBuilder("http", "127.0.0.1", 0).Uri;
```
Tratcher: Why default to v4 rather than v6?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -11,39 +11,68 @@ public static class TestUriHelper
...
+        {
+            return BuildTestUri(hint, statusMessagesEnabled: false);
+        }
+
+        public static Uri BuildTestUri(string hint, bool statusMessagesEnabled)
         {
             if (string.IsNullOrEmpty(hint))
             {
-                return BuildTestUri();
+                if (statusMessagesEnabled)
```
Tratcher: How is this going to work with HttpSys? It can't bind to port 0. Does it always have to set statusMessagesEnabled=false or provide a hint?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
mikeharder: If the test doesn't care, I assume IPv4 is safer since some environments don't support IPv6 (parts of Azure, maybe Travis/AppVeyor).

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -11,39 +11,68 @@ public static class TestUriHelper
...
+                    // supported, so the port is only bound on "127.0.0.1" (IPv4).  If a test explicitly requires IPv6,
+                    // it should provide a hint URL with "localhost" (IPv4 and IPv6) or "[::1]" (IPv6-only).
+                    return new UriBuilder("http", "127.0.0.1", 0).Uri;
+                }
+                else
+                {
+                    // If status messages are disabled, there is no status message from which to scrape the assigned port,
+                    // so the less reliable GetNextPort() must be used.  The port is bound on "localhost" (both IPv4 and IPv6),
+                    // since this is supported when using a specific (non-zero) port.
+                    return new UriBuilder("http", "localhost", GetNextPort()).Uri;
```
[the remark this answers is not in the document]
mikeharder: Adding retries is possible but difficult.  The AddressInUse exception doesn't come from `GetNextPort()`, it comes from the application started in an external process.  So the retry logic would need to be here:

https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs#L47

I think we'd need to try to start the app, and if it fails to start we'd need to scrape the output for AddressInUseException, and then retry on a different port.  We'd need to be careful to *not* retry if it failed for any other reason.  We'd also need to be careful to only retry if `BuildTestUri()` used `GetNextPort()`, since there's no point retrying if the test is using a hardcoded port.

Overall, this seems risky and error-prone, and it might cause more problems than it solves.  I think we should wait and see how many flaky tests remain after changing the default to use a dynamic port.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -11,39 +11,68 @@ public static class TestUriHelper
     {
         public static Uri BuildTestUri()
         {
-            return new UriBuilder("http", "localhost", GetNextPort()).Uri;
+            return BuildTestUri(null);
         }
 
         public static Uri BuildTestUri(string hint)
+        {
+            return BuildTestUri(hint, statusMessagesEnabled: false);
```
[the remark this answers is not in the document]
mikeharder: The `TestUriHelper` class and existing `BuildTestUri()` methods are public, so they can be called directly by test classes in other assemblies, not just by `SelfHostDeployer`.  If the methods are called directly, we have no way of knowing if status messages are enabled or not, so we can't rely on them.

I made the new overload of `BuildTestUri` internal to prevent other classes from calling it directly, since it should only be used by `SelfHostDeployer`.  I will also review all existing usage of `BuildTestUri()` and `BuildTestUri(string hint)` and see if we can remove them.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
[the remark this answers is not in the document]
JunTaoLuo: Yea it's just another enhancement we could add. Doesn't have to be in this PR which is why I approved it already.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/TestUriHelper.cs
```diff
@@ -11,39 +11,68 @@ public static class TestUriHelper
...
+
+        internal static Uri BuildTestUri(string hint, bool statusMessagesEnabled)
         {
             if (string.IsNullOrEmpty(hint))
             {
-                return BuildTestUri();
+                if (statusMessagesEnabled)
+                {
+                    // Most functional tests use this codepath and should directly bind to dynamic port "0" and scrape
+                    // the assigned port from the status message, which should be 100% reliable since the port is bound
```
halter73: Does SelfHostDeployer already scrape the assigned port from the status message? Does this mean that any test that uses SelfHostDeployer w/ Kestrel w/o specifying a port should just work?

mikeharder: SelfHostDeployer already scrapes the assigned port when status messages are enabled (which is the default for most tests).  So yes, once this change is merged and repos are updated to use this version, most tests should get more reliable with no other changes.

PR #1386, opened 2018-04-17
[Fixes #1242] Moves IStartupConfigureServicesFilter and IStartupConfigureContainerFilter interfaces to the internal namespace

### LINKED ISSUE #1242: IStartupConfigureServicesFilter is not executed when Configure is used
https://github.com/aspnet/Hosting/blob/872fc80d567892096f6cd6925600285570d866fe/test/Microsoft.AspNetCore.Hosting.Tests/StartupManagerTests.cs#L141-L166

There no reason why we can't do it except internal implementation details.

/cc @Tratcher @javiercn

javiercn:
#1242

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/Internal/IStartupConfigureContainerFilter.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/Internal/IStartupConfigureContainerFilter.cs
@@ -3,8 +3,12 @@
 
 using System;
 
-namespace Microsoft.AspNetCore.Hosting
+namespace Microsoft.AspNetCore.Hosting.Internal
 {
+    /// <summary>
+    /// This API supports the ASP.NET Core infrastructure and is not intended to be used
+    /// directly from your code. This API may change or be removed in future releases.
+    /// </summary>
     public interface IStartupConfigureContainerFilter<TContainerBuilder>
     {
         Action<TContainerBuilder> ConfigureContainer(Action<TContainerBuilder> container);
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Hosting.Abstractions/IStartupConfigureContainerFilter.cs
```diff
@@ -3,8 +3,12 @@
 
 using System;
 
-namespace Microsoft.AspNetCore.Hosting
+namespace Microsoft.AspNetCore.Hosting.Internal
```
Tratcher: Move to internal folder as well

javiercn: Done

Tratcher:
@javiercn you didn't actually implement the proposed changes in https://github.com/aspnet/Hosting/issues/1242#issuecomment-367146987.

pakrym:
Having cross repo `.Internal` type usage is bad but considering we don't have enough time for anything else...

javiercn:
@pakrym Just to clarify. This is only cross-assembly, not cross-repository, every other component just consumes this indirectly.

It's the cheapest thing we can do right now and it completely frees us to do anything we want in the next release.

Tratcher:
For comparison, what's the real change?

javiercn:
@Tratcher There are a couple of options we can do here:
* We can maintain the filters, but replacing the service provider factory to apply the filters there:
  * This is more work and is risky to do at this point as replacing the service provider factory might have unanticipated/unexpected consequences.
* We change the implementation to support more/other scenarios we discarded on the past.
 * This is more work and requires some design.

This change allows us to take any approach we want in 2.2.0 and is risk free.

javiercn:
@pakrym @Tratcher This got approved through ship-room. Any more feedback before I merge it?

Issue #1384, opened 2018-04-16
Parent Request Id is mislabeled as Context Id in log scope


ngbrown:
For the upcoming for ASP.Net Core 2.1; pull #1138 added to the log scope a `ContextId`, but the contents are only the parent request identifier and does not appear to be set if there was no parent request (no `Request-Id` header in the incoming HTTP request).

There is documentation/specifications (mostly by @lmolkova) around these terms at:

* [HttpCorrelationProtocol](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md)
* [FlatRequestId](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/FlatRequestId.md)
* [HierarchicalRequestId](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HierarchicalRequestId.md)

Specifically in the [Activity User Guide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/ActivityUserGuide.md), it says:

> When application calls external dependency to complete an operation, it may need to pass some of the ___context___ (e.g. correlation id) [emphasis added] along with dependency call to be able to correlate logs from multiple service.

I see that the expectation is that in the logs, there will be a "correlation id" of the context, or `ContextId` if you will, that can be used to link the logs entries across multiple services handling a single user request.

In the current preview code, the Log scope outputs a `ContextId` which is the value of an incoming header `Request-Id`, and only if that header is set.  This is not useful for seeing the entire request across multiple services in the logs.  What is needed is adding to the log scope the root request id to the log scope of all services involved in handling the request, including within the root service that generated the first request id.

lmolkova:
@ngbrown 
The spec implemented in .NET and ASP.NET is [HierarchicalRequestId](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HierarchicalRequestId.md)

i.e. 'correlationId' looks like `|root-id-that-never-changes.some_unique.suffix.1.2.3.`

You are right that 'parent request id' is the correct name for it. 

Instead of setting parent Request-Id header value to logger scopes, we should use `Activity.Current.Id` that is generated for each request even if there were no Request-Id header.

[...]

PR #1394, opened 2018-04-21
Use hierarchical Id from Activity for logging scope

ngbrown:
This addresses #1384 by using the `Activity` based hierarchical ids in the logger scope.

Using a structured logging framework is not synonymous with having a `DiagnosticListener` implemented.  So this pull request starts an `Activity` in the `HostingApplication.Context` if logging is enabled in addition to when the diagnostic listener is enabled.

Just searching the logs for `RequestId` starts with `|Guid.` wont work in mixed scenarios because a chain of services could start the `RequestId` hierarchical id over if the parent service only supports flat Ids.  The [spec suggests](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/FlatRequestId.md#mixed-hierarchical-and-non-hierarchical-scenario) also searching for `Correlation-Context: Id` like this:

```
CorrelationContext.Id == Guid || RequestId.startsWith('|Guid')
```

However, structured log providers don't necessarily destructure nested scope values (Serilog doesn't unless the name begins with "@") so to fully correlate a request in the logs, `CorrelationId` uses the `Correlation-Context: Id` header value if present, otherwise it uses the RootId that `Activity` pulls from a hierarchical ids.  (dotnet/corefx#29265 is related to updating the `RootId` for this reason).

Currently, if there is a customized version of `IHttpRequestIdentifierFeature` that is provided as a feature, then it's generated `TraceIdentifier` will no longer be part of the log scope.  I'm open to suggestions on how to make this customization.

This pull request would be less required to make hierarchical Id's work for logging if some of the above could be swapped out and resolved at the application level.  However `HostingApplication` doesn't allow enough customization and `new`s stuff up, including `HostingApplicationDiagnostics` and `HostingApplication.Context`  (`HostingApplication` is in turn is `new`ed by `WebHost`).

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
@@ -77,11 +77,12 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 // If the baggage includes the Id, then than is more correct as the root Id than the
                 // Activity.RootId in the case of a mixed hierarchical and non-hierarchical scenario.
                 var activityRootId = context.Activity.GetBaggageItem("Id") ?? context.Activity.RootId;
+                var requestId = context.Activity.Id;
 
                 // Scope may be relevant for a different level of logging, so we always create it
                 // see: https://github.com/aspnet/Hosting/pull/944
                 // Scope can be null if logging is not on.
-                context.Scope = _logger.RequestScope(httpContext, parentRequestId, activityRootId);
+                context.Scope = _logger.RequestScope(httpContext, requestId, parentRequestId, activityRootId);
 
                 if (_logger.IsEnabled(LogLevel.Information))
                 {
--- a/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
@@ -55,14 +55,15 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             if (diagnosticListenerEnabled || loggingEnabled)
             {
                 httpContext.Request.Headers.TryGetValue(RequestIdHeaderName, out parentRequestId);
+
+                if (_diagnosticListener.IsEnabled(ActivityName, httpContext) || loggingEnabled)
+                {
+                    context.Activity = StartActivity(httpContext, parentRequestId);
+                }
             }
 
             if (diagnosticListenerEnabled)
             {
-                if (_diagnosticListener.IsEnabled(ActivityName, httpContext))
-                {
-                    context.Activity = StartActivity(httpContext, parentRequestId);
-                }
                 if (_diagnosticListener.IsEnabled(DeprecatedDiagnosticsBeginRequestKey))
                 {
                     startTimestamp = Stopwatch.GetTimestamp();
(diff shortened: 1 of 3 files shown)


benaadams:
`BeginRequest` is looking very chunky for `AggressiveInlining`; any scope to shift some of the conditional branches into method calls?

davidfowl:
We're not taking this for 2.1, anything we do here will be in 2.2. We can try to remove the correlation id that was added in 2.1 so we're at feature parity with 2.0. 

Also, as @benaadams says this code is messy already and extremely performance sensitive so we can't add to it lightly (Like always starting an activity).

That said, we do want to fix this distributed tracing logging scope activity ID situation we're currently in by designing something that will properly work with structured logging by default. What we have right now is too tied to DiagnosticListeners and the whole AI experience.

davidfowl:
@ngbrown Also, could you show what the experience looks like in a tool like seq or kibana, AI analytics? That will help visualize the changes you've made here.

ngbrown:
On the `BeginRequest` looking chunky, is there a tool metric or benchmark to measure that?  Does the performance sensitive case mostly apply to when logging is disabled?

With logging disabled the heaviest work is probably just checking if logging/diagnostic are each enabled, and that was added in #1138.  The logging/diagnostic listener code could be refactored to pull that branch code into a separate function to reduce the cyclomatic complexity.

If the added correlation id is going to be pulled back out, is someone specifically responsible for reverting pull #1138?

I've used Seq, so I can put a scenario together to illustrate the experience with with that.

benaadams:
> On the BeginRequest looking chunky, is there a tool metric or benchmark to measure that?

Its mostly the about amount of code jumped over in the fast-path as its just taking up space in the CPU's instruction cache for no purpose. 

As its cache lines are the same as the data cache; cache lines are 64 byte chunks of asm. So "unused code" in the fast-path can use up a lot of CPU cache lines that the CPU still has to pull from main memory; which can even evict the fast-path from cache.

> Does the performance sensitive case mostly apply to when logging is disabled?

Yes

However... there is an issue around scopes as they aren't necessarily enabled; so allocating lots for them when its just thrown away can be problematic, but that might be more an ASP.NET issue to solve https://github.com/aspnet/Logging/issues/714

benaadams:
Inlining looks good with the change (even better than before 😄)

src/Microsoft.AspNetCore.Hosting/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -167,6 +131,53 @@ public void ContextDisposed(HostingApplication.Context context)
...
+                {
+                    startTimestamp = Stopwatch.GetTimestamp();
+                    RecordBeginRequestDiagnostics(httpContext, startTimestamp);
+                }
+            }
+
+            // To avoid allocation, return a null scope if the logger is not on at least to some degree.
+            if (loggingEnabled)
+            {
+                // If the baggage includes the Id, then than is more correct as the root Id than the
```
Tratcher: then than

davidfowl:
@ngbrown can you setup the Seq thing and paste a screen shot. I'll try to get this into 2.1 and running a perf test

ngbrown:
Does a separate pull request need to be opened to remove the correlation id from 2.1 before it goes RTM?

davidfowl:
OK I'm going to take another look at this.

natemcmaster:
Heads up @ngbrown, we're going to be merging this repo tomorrow into https://github.com/aspnet/AspNetCore as a part of https://github.com/aspnet/AspNetCore/issues/3598. Unless @davidfowl or @Tratcher are available to merge this in the next day, this PR will be closed and you'll have to re-open this against the new repo.

davidfowl:
Feel free to close this. We're not ready to merge it as is. 

@ngbrown There's a bunch of work going into distributed tracing in 3.0. I'll bring you up to speed once we move repos.

PR #1395, opened 2018-04-22
There should only be a single IHostLifetime in the service container.

polewskm:
### Problem Statement
Currently both `HostBuilder.CreateServiceProvider` and `HostingHostBuilderExtensions.UseConsoleLifetime` add to the service collection which causes multiple registrations for the `IHostLifetime` singleton. When using the default service container, this is not a problem because only the last registration is used when creating singleton instances. But this may be a problem for other DI containers.

### Proposed Solution
When configuring the `IHostLifetime` service registration, use the `Replace` extension method for `IServiceCollection` instead of calling `Add`. I also added unit tests and a `UseHostLifetime` extension method for `IHostBuilder`.

### DIFF
--- a/src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
+++ b/src/Microsoft.Extensions.Hosting/HostingHostBuilderExtensions.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting.Internal;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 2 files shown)


davidfowl:
> But this may be a problem for other DI containers.

Do you have examples of ones that support the DI spec but break with the generic host? I believe this is one of the requirements.

polewskm:
So I just tested the following containers:

- Autofac
- DryIoc
- Grace
- LightInject
- StructureMap
- Stashbox
- Unity

and they all seem to behave identically by returning the last registration when multiple registrations are added.

Since there is no known current defect with the current generic host implementation adding multiple registrations for `IHostLifetime`, I will understand if you reject this pull request. I was merely hoping to remove any potential problems or confusion with multiple registrations for a singleton.

davidfowl:
If that’s didn’t work nothing would work as nothing in our stack uses replace.

PR #1397, opened 2018-04-24
Check existence of ANCMV2 Path

jkotalik:
By creating a new global module for ANCM (aspnetcorev2.dll), we need the IISExpressDeployer to be able to change the ANCMPath. The issue is, the ANCMPath is hard coded to aspnetcore.dll, meaning we cannot use it for aspnetcorev2.dll. Also, it is hard to modify it outside of the deployer because the temporary file and folder location is done on [DotnetPublish()](https://github.com/aspnet/Hosting/blob/78904e70c7ee04d36a2de9182e66f14880e808df/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs#L52), which is called on [DeployAsync()](https://github.com/aspnet/Hosting/blob/78904e70c7ee04d36a2de9182e66f14880e808df/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs#L58).

There are a few other ways I can make this fix (ex: creating a property that is the aspnetcore dll to replace ANCMPath), but this seemed like the simplest for now.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
@@ -124,6 +124,20 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                                 serverConfig.Replace("[ANCMPath]", ancmFile);
                         }
 
+                        if (serverConfig.Contains("[ANCMV2Path]"))
+                        {
+                            var ancmFile = Path.Combine(contentRoot, Is64BitHost ? @"x64\aspnetcorev2.dll" : @"x86\aspnetcorev2.dll");
+
+                            if (!File.Exists(Environment.ExpandEnvironmentVariables(ancmFile)))
+                            {
+                                throw new FileNotFoundException("AspNetCoreModuleV2 could not be found.", ancmFile);
+                            }
+
+                            Logger.LogDebug("Writing ANCMPath '{ancmPath}' to config", ancmFile);
+                            serverConfig =
+                                serverConfig.Replace("[ANCMV2Path]", ancmFile);
+                        }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -124,6 +124,20 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
...
+                        if (serverConfig.Contains("[ANCMV2Path]"))
+                        {
+                            var ancmFile = Path.Combine(contentRoot, Is64BitHost ? @"x64\aspnetcorev2.dll" : @"x86\aspnetcorev2.dll");
+
+                            if (!File.Exists(Environment.ExpandEnvironmentVariables(ancmFile)))
+                            {
+                                throw new FileNotFoundException("AspNetCoreModuleV2 could not be found.", ancmFile);
+                            }
+
+                            Logger.LogDebug("Writing ANCMPath '{ancmPath}' to config", ancmFile);
```
Tratcher: Update the log to ANCMV2Path

Tratcher: Why is the format parameter `ancmPath`, but the field is `ancmFile`?

pakrym:
Which one are we going to put into web.configs of test apps?

jkotalik:
@pakrym here was my original intention with this change. 
For ANCMV1, the web.config file would contain the attribute ```handler="AspNetCoreModule"``` and for ANCMV2 it would be ```handler="AspNetCoreModuleV2"```. ANCM V1 will only use aspnetcore.dll and ANCM V2 will only use aspnetcorev2.dll. 
The drawback with this approach is we need to duplicate sites for V1 and V2. However, I'd rather test V2 with the changed dll names as it will be a likely breaking point.

Issue #1399, opened 2018-04-24
Self-host functional tests broken by nuget warning


Tratcher:
Functional tests using the SelfHostDelpoyer are broken due to a new nuget warning about mismatched Microsoft.NETCore.App versions.

This affects ServerTests, WebSockets, and MusicStore

```
| [2018-04-24T02:56:08] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: Deploying [Variation] :: ServerType=Kestrel, Runtime=CoreClr, Arch=x64, BaseUrlHint=, Publish=False
| [2018-04-24T02:56:08] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: Executing dotnet run --no-restore --framework netcoreapp2.1 /p:VerifyMatchingImplicitPackageVersion=false -- --server.urls http://127.0.0.1:0/  --server Microsoft.AspNetCore.Server.Kestrel
| [2018-04-24T02:56:08] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: SET ASPNETCORE_ENVIRONMENT=ManagedSockets
| [2018-04-24T02:56:08] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: SET ASPNETCORE_DETAILEDERRORS=true
| [2018-04-24T02:56:08] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: Started dotnet. Process Id : 14032
| [2018-04-24T02:56:09] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: dotnet stdout: C:\b\w\33bdfc1cae7b2a38\modules\WebSockets\test\AutobahnTestApp\AutobahnTestApp.csproj : error : The project was restored using Microsoft.NETCore.App version 2.2.0-preview1-26423-02, but with current settings, version 2.1.0-preview3-26413-05 would be used instead.  To resolve this issue, make sure the same settings are used for restore and for subsequent operations such as build or publish.  Typically this issue can occur if the RuntimeIdentifier property is set during build or publish but not during restore.
| [2018-04-24T02:56:09] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Warning: dotnet stderr: The build failed. Please fix the build errors and run again.
[...]

PR #1400, opened 2018-04-24
Run self-host tests with no-build

Tratcher:
#1399 The CI test passes build the whole repo before running tests, so we shouldn't need to rebuild the test sites before deploying them. I tested this privately with ServerTests without issue, but it's hard to replicate the CI testing environment to make sure it fixes the nuget warning we only see there.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Reflection;
 
 namespace Microsoft.AspNetCore.Server.IntegrationTesting
 {
@@ -45,6 +46,12 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
             ServerType = serverType;
             RuntimeFlavor = runtimeFlavor;
             EnvironmentVariables["ASPNETCORE_DETAILEDERRORS"] = "true";
+
+            var configAttribute = Assembly.GetCallingAssembly().GetCustomAttribute<AssemblyConfigurationAttribute>();
+            if (configAttribute != null && !string.IsNullOrEmpty(configAttribute.Configuration))
+            {
+                Configuration = configAttribute.Configuration;
+            }
         }
 
         public ServerType ServerType { get; }
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
@@ -97,9 +97,7 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                     var targetFramework = DeploymentParameters.TargetFramework ?? (DeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr ? "net461" : "netcoreapp2.0");
 
                     executableName = DotnetCommandName;
-                    // Set VerifyMatchingImplicitPackageVersion to disable errors when Microsoft.NETCore.App's version is overridden externally
-                    // This verification doesn't matter if we are skipping restore during tests.
-                    executableArgs = $"run --no-restore --framework {targetFramework} /p:VerifyMatchingImplicitPackageVersion=false {DotnetArgumentSeparator}";
+                    executableArgs = $"run --no-build -c {DeploymentParameters.Configuration} --framework {targetFramework} {DotnetArgumentSeparator}";
                 }
 
                 executableArgs += $" --server.urls {hintUrl} "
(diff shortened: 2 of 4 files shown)


Tratcher:
Hmm, one test in Hosting is consistently failing, buto only on Mac and Windows, not Linux. And it doesn't fail locally on Windows. It is using the modified code path.
InjectedStartup_DefaultApplicationNameIsEntryAssembly

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -97,9 +97,7 @@ protected async Task<(Uri url, CancellationToken hostExitToken)> StartSelfHostAs
                     var targetFramework = DeploymentParameters.TargetFramework ?? (DeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr ? "net461" : "netcoreapp2.0");
 
                     executableName = DotnetCommandName;
-                    // Set VerifyMatchingImplicitPackageVersion to disable errors when Microsoft.NETCore.App's version is overridden externally
-                    // This verification doesn't matter if we are skipping restore during tests.
-                    executableArgs = $"run --no-restore --framework {targetFramework} /p:VerifyMatchingImplicitPackageVersion=false {DotnetArgumentSeparator}";
+                    executableArgs = $"run --no-build --no-restore --framework {targetFramework} {DotnetArgumentSeparator}";
```
natemcmaster: So, is it safe to assume the application has already been built?

Tratcher: That's the theory. I don't think even a local universe build would be enough to test it though, the CI runs tests differently.

natemcmaster: To test this, try doing this:
```
dotnet restore /p:MicrosoftNETCoreApp21PackageVersion=(something newer than what's in build/dependencies.props)
dotnet run --no-build
```
I want to know if the target that issues the VerifyMatchingImplicitPackageVersion error still runs if you pass `--no-build`.

natemcmaster: By the way, --no-build implies --no-restore

Tratcher:
Ah, figured it out, I needed to set the configuration. We pre-build everything as release, but `run` defaults to debug. This should also be a lot faster.

test/Microsoft.AspNetCore.Hosting.FunctionalTests/WebHostBuilderTests.cs
```diff
@@ -43,7 +43,12 @@ public class WebHostBuilderTests : LoggedTest
                 {
                     TargetFramework = runtimeFlavor == RuntimeFlavor.Clr ? "net461" : "netcoreapp2.0",
                     ApplicationType = ApplicationType.Portable,
-                    StatusMessagesEnabled = false
+                    StatusMessagesEnabled = false,
+#if DEBUG
```
Tratcher: @pranavkm @natemcmaster any ideas how we can avoid #if DEBUG in every test?  I found https://github.com/aspnet/BuildTools/blob/c390edb6f73e3ab992402835c004414c1fadfd4a/files/KoreBuild/KoreBuild.Common.props#L33 for how we enable Release builds for appveyor and travis, and the CI driectly sets Configuration via env var.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -45,6 +45,18 @@ public class DeploymentParameters
             ServerType = serverType;
             RuntimeFlavor = runtimeFlavor;
             EnvironmentVariables["ASPNETCORE_DETAILEDERRORS"] = "true";
+
+            var config = Environment.GetEnvironmentVariable("CI");
+            if (!string.IsNullOrEmpty(config) && bool.TryParse(config, out var isCi) && isCi)
+            {
+                Configuration = "Release";
```
natemcmaster: This isn't always going to be a safe assumption. We are working on adding Debug CI configurations too.

Tratcher: On travis & appveyor? That's where this variable is coming from.

natemcmaster:
> any ideas how we can avoid #if DEBUG in every test?

Add targets to the IntegrationTesting package that generate something to capture the build context. AssemblyAttributes, a plain text file, etc.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
natemcmaster: TeamCity also sets `CI=true`. We default to Configuration=Release when CI == true, but it's not guaranteed that CI==true implies Configuration == Release.

natemcmaster: Example: https://github.com/aspnet/KestrelHttpServer/pull/2523

Tratcher:
The state is already captured in the AssemblyConfiguration attribute, but knowing which assembly to check for it is the trick. You'd have a similar issue for any other attribute you added. GetCallingAssembly looks like it will work for all of our use cases, and I've already verified it with ServerTests.

Issue #1402, opened 2018-04-26
Add a AddHostedService extension


Tratcher:
In the generic host an IHostedService is the primary unit of work.  Add an `AddHostedService<THostedService>` convenience extension.

This can also help mitigate the fact that several hosting types exists in both Extensions.Hosting and AspNetCore.Hosting namespaces and may be confusing.

Tratcher:
@DamianEdwards has requested this for 2.1 RTM

Tratcher:
@davidfowl This extension actually makes the problem worse. It's in the DI namespace in a transitive dependency so it starts showing up in AspNetCore projects. It also hides which IHostedService is being registered. This is going to confuse and break AspNet users.

```
using Microsoft.Extensions.Hosting;

namespace Microsoft.Extensions.DependencyInjection
{
    public static class ServiceCollectionHostedServiceExtensions
    {
        /// <summary>
        /// Add an <see cref="IHostedService"/> registration for the given type.
        /// </summary>
        /// <typeparam name="THostedService">An <see cref="IHostedService"/> to register.</typeparam>
        /// <param name="services">The <see cref="IServiceCollection"/> to register with.</param>
        /// <returns>The original <see cref="IServiceCollection"/>.</returns>
        public static IServiceCollection AddHostedService<THostedService>(this IServiceCollection services)
            where THostedService : class, IHostedService
            => services.AddTransient<IHostedService, THostedService>();
    }
}
```

Tratcher:
Ah, nevermind. I forgot that we consolidated IHostedService

PR #1405, opened 2018-05-01
Add a AddHostedService extension 

Tratcher:
#1402 This helps avoid the ambiguity between types in Extensions.Hosting and AspNetCore.Hosting.

This requires shiproom approval.


talanc:
it looks like it wasnt changed to singleton on release -- why is that?

PR #1407, opened 2018-05-01
Make deployer handle ANCM V1 and V2

jkotalik:
Right now, in IISIntegration, we handle modifying the web.config file from modules="AspNetCoreModule" to "AspNetCoreModuleV2". Instead, we can use a property to modify it in the deployer.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ANCMVersion.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ANCMVersion.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting
+{
+    public enum ANCMVersion
+    {
+        AspNetCoreModule,
+        AspNetCoreModuleV2
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ANCMVersion.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting
+{
+    public enum ANCMVersion
+    {
+        AspNetCoreModule,
```
Tratcher: V1

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -120,6 +120,12 @@ public class DeploymentParameters
 
         public HostingModel HostingModel { get; set; }
 
+        /// <summary>
+        /// When using the IISExpressDeployer, determines whether to use the older or newer version
+        /// of ANCM.
+        /// </summary>
+        public ANCMVersion ANCMVersion { get; set; } = ANCMVersion.AspNetCoreModuleV2;
```
Tratcher: Default to v1 or you'll break existing tests.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ANCMVersion.cs
jkotalik: I want to use the name as the value for the ```modules="..."``` section, hence the value doesn't include V1.

PR #1409, opened 2018-05-03
Fixes incorrect descendant for handlers section.

jkotalik:
For https://github.com/aspnet/IISIntegration/issues/758


jkotalik:
Yeah, I reverted it before so now it should be good to go.

Issue #1371, opened 2018-03-31
Move IIS deployers to IISIntegration.


jkotalik:
Though logically it makes sense to have the deployers together, modifying/using the IIS deployer cross repo is painful. There are no Deployer tests in this repo, which means validation for them takes either hacking up a dependency, adding deployer as a project reference, or blindly adding a dependency. 

cc/ @Tratcher @pakrym

Tratcher:
I expect there's an existing issue for this in the backlog. It can't be done without a complete refactor of the deployers to an extensible model. Right now everything is statically registered.

jkotalik:
Why can't we just move the IISExpressDeployer class to IISIntegration? That's all this is really asking for.

Tratcher:
It's not quite as bad as I remembered. You need to get rid of this method that all of the tests call:
https://github.com/aspnet/Hosting/blob/e93aa56b588bee1e9ef0d25fd0176545f91d0c40/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployerFactory.cs#L20-L49

But you'd have to duplicate that code in the tests, or write them quite differently.
https://github.com/aspnet/ServerTests/blob/042b0ff46a758941766c011cf210e27164d3e2fb/test/ServerComparison.FunctionalTests/ResponseTests.cs#L157-L173

pakrym:
@muratg consider for MQ work

Tratcher:
A related issue about refactoring this code: https://github.com/aspnet/Hosting/issues/241

jkotalik:
Now that we have added the [IIS deployer](https://github.com/aspnet/IISIntegration/commit/dfd75e939d49733fd8dcb0104a7b121671c467f6) to IISIntegration, I think it is a good time to move the IISExpressDeployer.

Tratcher:
Figure out how to get the IIS deployer working in ServerTests first.

jkotalik:
@pakrym can this be considered done?

PR #1412, opened 2018-05-08
Refactor integration testing for test matrix generation.

Tratcher:
#1370 - Usability improvements via a new TestMatrix generator, and making it easier to flow common parameters via TestVariant. I'll send out reaction PRs up stack to demonstrate the changes. I already have it working with ServerTests and some of IISIntigration.

Also renamed WebListener -> HttpSys, ANCMVersion -> AncmVersion, etc..


src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -13,6 +14,35 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
     /// </summary>
     public class DeploymentParameters
     {
+        public DeploymentParameters()
```
pakrym: We are having to many entities now, why do we need both this and variations with the same information?

pakrym:
Can you link to a sample of how the test code would look after this change?

Tratcher:
https://github.com/aspnet/ServerTests/pull/120

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
Tratcher: The TestVariant has pretty strict requirements around serialization. DeploymentParameters has a lot more state including things like `IDictionary<string, string>` and `Action<DeploymentParameters>`. I'll see if I can collapse DeploymentParameters into the ApplicationDeployer base class.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -13,6 +14,35 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
...
+            var configAttribute = Assembly.GetCallingAssembly().GetCustomAttribute<AssemblyConfigurationAttribute>();
+            if (configAttribute != null && !string.IsNullOrEmpty(configAttribute.Configuration))
+            {
+                Configuration = configAttribute.Configuration;
+            }
+        }
+
+        public DeploymentParameters(TestVariant variant)
+        {
+            EnvironmentVariables["ASPNETCORE_DETAILEDERRORS"] = "true";
```
jkotalik: Call other constructor and then set variant information.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -124,7 +154,7 @@ public class DeploymentParameters
         /// When using the IISExpressDeployer, determines whether to use the older or newer version
         /// of ANCM.
         /// </summary>
-        public ANCMVersion ANCMVersion { get; set; } = ANCMVersion.AspNetCoreModule;
+        public AncmVersion AncmVersion { get; set; } = AncmVersion.AspNetCoreModule;
```
jkotalik: Make sure this namechange changed across all repos before merging.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/Tfm.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting
+{
+    public static class Tfm
+    {
+        public const string Net461 = "net461";
```
jkotalik: Enum instead with ToString?

src/Microsoft.AspNetCore.Server.IntegrationTesting/TestMatrix.cs
```diff
@@ -0,0 +1,266 @@
...
+{
+    public class TestMatrix
+    {
+        public IList<ServerType> Servers { get; set; } = new List<ServerType>();
+        public IList<string> Tfms { get; set; } = new List<string>(); // Can derive from running TFM
+        public IList<ApplicationType> ApplicationTypes { get; set; } = new List<ApplicationType>(); // net461/CLR is always portable, derive from TFM
+        public IList<RuntimeArchitecture> Architectures { get; set; } = new List<RuntimeArchitecture>();
+
+        // ANCM specific...
+        public IList<HostingModel> HostingModels { get; set; } = new List<HostingModel>();
```
jkotalik: Subclass for ANCM specific stuff?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
Tratcher: I'm a little worried about that messing up GetCallingAssembly(). I plan to get rid of most of these overloads. I primarily left them to ease the transition.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/Tfm.cs
Tratcher: No, it's an open set and it may even be case sensitive in some places.

src/Microsoft.AspNetCore.Server.IntegrationTesting/TestMatrix.cs
Tratcher: I'll look closer at that seperation as part of https://github.com/aspnet/Hosting/issues/1371

src/Microsoft.AspNetCore.Server.IntegrationTesting/TestMatrix.cs
```diff
@@ -0,0 +1,266 @@
...
+                foreach (var tfm in Tfms)
+                {
+                    if (!TfmIsSupportedOnThisOS(tfm))
+                    {
+                        continue;
+                    }
+
+                    foreach (var t in ApplicationTypes)
+                    {
+                        var type = t;
```
jkotalik: Why?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/Tfm.cs
pakrym: How often do we add  Tfms? And it shouldn't be case sensitive

src/Microsoft.AspNetCore.Server.IntegrationTesting/TestMatrix.cs
```diff
@@ -0,0 +1,266 @@
...
+                        if (Tfm.Matches(Tfm.Net461, tfm) && type == ApplicationType.Portable)
+                        {
+                            if (ApplicationTypes.Count == 1)
+                            {
+                                // Override the default
+                                type = ApplicationType.Standalone;
```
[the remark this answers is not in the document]
pakrym: No, there is no portable net461.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
Tratcher: Consolidating DeploymentParameters is going to cause way too much churn for this pass with no functional benifit. MVC has a whole test infrastructure built around it.

src/Microsoft.AspNetCore.Server.IntegrationTesting/TestMatrix.cs
Tratcher: Otherwise it won't let me overwrite t inside the loop. foreach params are readonly.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/Tfm.cs
Tratcher: Every release. And you don't want to prevent one-off testing of other TFMs. E.g. doing a manual test against net471.

Tratcher: This is only a helper to avoid having strings everywhere and prevent typoes.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployerFactory.cs
```diff
@@ -35,7 +35,7 @@ public static IApplicationDeployer Create(DeploymentParameters deploymentParamet
                     return new IISExpressDeployer(deploymentParameters, loggerFactory);
                 case ServerType.IIS:
                     throw new NotSupportedException("The IIS deployer is no longer supported");
-                case ServerType.WebListener:
+                case ServerType.HttpSys:
```
jkotalik: You'll need to make a PR to music store due to this namechange.

Tratcher: And about a dozen other repos. I'll work through those once IIS and ServerTests are happy.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/RuntimeFlavor.cs
```diff
@@ -5,7 +5,8 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
 {
     public enum RuntimeFlavor
     {
-        Clr,
-        CoreClr
+        None,
```
pakrym: none?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ServerType.cs
```diff
@@ -5,9 +5,10 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
 {
     public enum ServerType
     {
+        None,
```
pakrym: What the use case for None?

Tratcher: So we can tell if a field was accidentally not set. I want to avoid defaulting to an incorrect value.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/RuntimeFlavor.cs
Tratcher: So we can tell if nothing was specified and provide a default.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/ServerType.cs
pakrym: Nullable?

Tratcher:
Merging now because Ryan has some issues in IIS that he needs me to unblock with this.

PR #1420, opened 2018-05-14
 Always check if ANCM is in the base output folder.

jkotalik:
Fix for https://github.com/aspnet/IISIntegration/issues/831.
For standalone coreclr applications, the aspnetcorerh.dll will be put in the app root folder rather than the (Bitness)/aspnetcorerh.dll folder. Instead of special casing for deployment type, we will check the app root too.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
@@ -106,36 +106,9 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
 
                         // Pass on the applicationhost.config to iis express. With this don't need to pass in the /path /port switches as they are in the applicationHost.config
                         // We take a copy of the original specified applicationHost.Config to prevent modifying the one in the repo.
-                        if (serverConfig.Contains("[ANCMPath]"))
-                        {
-                            // We need to pick the bitness based the OS / IIS Express, not the application.
-                            // We'll eventually add support for choosing which IIS Express bitness to run: https://github.com/aspnet/Hosting/issues/880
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -265,6 +238,26 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
             }
         }
 
+        private string ModifyANCMPathInConfig(string replaceFlag, string dllName, string serverConfig, string contentRoot)
+        {
+            if (serverConfig.Contains(replaceFlag))
+            {
+                var ancmFile = Path.Combine(contentRoot, Is64BitHost ? $@"x64\{dllName}" : $@"x86\{dllName}");
+                if (!File.Exists(Environment.ExpandEnvironmentVariables(ancmFile)))
```
pakrym: var p = Environment.ExpandEnvironmentVariables(ancmFile)

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -265,6 +238,26 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
...
+                {
+                    ancmFile = Path.Combine(contentRoot, Is64BitHost ? $@"{dllName}" : $@"{dllName}");
```
pakrym: `$@"{dllName}".ToString()`

pakrym: Remove ternary.

jkotalik:
I'm assuming this is good to merge?

PR #1426, opened 2018-05-18
Enable x86 testing

Tratcher:
#949 The biggest task is to get the right dotnet.exe in all the right places.

Verified with ServerTests. Minimal changes are required in the consuming repos.

TODO: The one big item left is to figure out how to modify the build scripts to install the x86 sdk and runtimes for select repos. We already have the commands to do it manually, I just need to automate it.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
@@ -65,11 +65,6 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                 throw new DirectoryNotFoundException(string.Format("Application path {0} does not exist.", applicationPath));
             }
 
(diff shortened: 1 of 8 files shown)


pakrym:
> TODO: The one big item left is to figure out how to modify the build scripts to install the x86 sdk and runtimes for select repos. We already have the commands to do it manually, I just need to automate it.

https://github.com/aspnet/IISIntegration/blob/dev/build/repo.props#L30-L31

Tratcher:
@pakrym thanks, that got me the right runtimes. How about the SDK?

Tratcher:
For `dotnet run`.

pakrym:
Can we change to dotnet build and then run directly?

pakrym:
You can install SDK using similar item group https://github.com/aspnet/BuildTools/blob/599e691c41f502ed9e062b1822ce13b673fc916e/modules/KoreBuild.Tasks/module.props#L68

Tratcher:
Switching to a feature branch.

Issue #1428, opened 2018-05-21
Enable net461 x86 self-host scenarios 


Tratcher:
A continuation of https://github.com/aspnet/Hosting/issues/949. I was able to get most of the x86 scenarios to work, but `dotnet run --no-build --framework net461` used by the selfhost deployer always launches x64, regardless of the bitness of dotnet.exe used. I suspect --no-build is part of the problem but it's required to avoid a different issue regarding nuget restore version mismatches. 

Recommend backlog since we're getting similar coverage for this in the IIS Express tests that publish the app first.

PR #1429, opened 2018-05-21
Enable x86 testing

Tratcher:
#949 I primarily had to use the right bitnes of dotnet.exe everywhere. ~~The only scenario I didn't work out was selfhost net461 without publishing~~ (https://github.com/aspnet/Hosting/issues/1428) - Fixed by publishing.

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DotNetCommands.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DotNetCommands.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.AspNetCore.Server.IntegrationTesting.Common
+{
+    internal static class DotNetCommands
+    {
+        private const string _dotnetFolderName = ".dotnet";
+
+        internal static string DotNetHome { get; } = GetDotNetHome();
+
+        // Compare to https://github.com/aspnet/BuildTools/blob/314c98e4533217a841ff9767bb38e144eb6c93e4/tools/KoreBuild.Console/Commands/CommandContext.cs#L76
+        private static string GetDotNetHome()
+        {
+            var dotnetHome = Environment.GetEnvironmentVariable("DOTNET_HOME");
+            var userProfile = Environment.GetEnvironmentVariable("USERPROFILE");
+            var home = Environment.GetEnvironmentVariable("HOME");
+
+            var result = Path.Combine(Directory.GetCurrentDirectory(), _dotnetFolderName);
+            if (!string.IsNullOrEmpty(dotnetHome))
+            {
+                result = dotnetHome;
+            }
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DotNetCommands.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+            else if (!string.IsNullOrEmpty(home))
+            {
+                result = home;
+            }
+
+            return result;
+        }
+
+        public static string GetDotNetInstallDir(RuntimeArchitecture arch)
```
pakrym: Does this have to be public?

pakrym:
Looks fine, not a fan how our code is structured but it's not a point of this PR.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DotNetCommands.cs
Tratcher: The whole class could be internal.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -246,8 +240,14 @@ private string ModifyANCMPathInConfig(string replaceFlag, string dllName, string
...
         {
+            var programFiles = "Program Files";
+            if (DotNetCommands.IsRunningX86OnX64(DeploymentParameters.RuntimeArchitecture))
+            {
+                programFiles = "Program Files (x86)";
+            }
+
             // Get path to program files
-            var iisExpressPath = Path.Combine(Environment.GetEnvironmentVariable("SystemDrive") + "\\", "Program Files", "IIS Express", "iisexpress.exe");
+            var iisExpressPath = Path.Combine(Environment.GetEnvironmentVariable("SystemDrive") + "\\", programFiles, "IIS Express", "iisexpress.exe");
```
pakrym: Can we use environment variable here?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -297,8 +297,24 @@ public override void Dispose()
             }
         }
 
-        // Transforms the web.config file to include the hostingModel="inprocess" element
-        // and adds the server type = Microsoft.AspNetServer.IIS such that Kestrel isn't added again in ServerTests
+        private void ModifyDotNetExePathInWebConfig()
+        {
+            // We assume the x64 dotnet.exe is on the path so we need to provide an absolute path for x86 scenarios.
```
pakrym: Let's not assume, let's always set the full path to dotnet we think is correct to run.

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
Tratcher: No, the environment variable changes based on the current process architecture, but we're trying to launch a process with a different arch.

Tratcher:
(offline approval)

PR #1432, opened 2018-05-24
Add try..finally around RunAsync and WaitForTokenShutdownAsync

### LINKED ISSUE #1194: Processus never stops on error at startup
Hello,

With the new Core 2.0 SDK, I'm using the extension method "IWebHost.Run" to start the host in a console application. But when an exception is thrown during startup (wrong url by exemple), the processus never stops even with "Ctrl-C" keys combination.

In the method RunAsync of the Microsoft.AspNetCore.Hosting.WebHostExtensions class, the code line "done.Set();" is never called so in the "AttachCtrlcSigtermShutdown" method, thread waits on the following code line "resetEvent.Wait();".

Is it possible to do a try - finally on "host.RunAsync" with "done.Set();" in the finally block or any other solution to not block the thread execution ?

Thanks in advance

Rutix:
The try..finally ensures that the ManualResetEventSlim which AttachCtrlcSigtermShutdown uses is set even when an exception occurs in these two methods.

Fixes #1194 

I also looked at if I could any tests to test this behavior (besides manually testing it with a repo sample project) and looked at https://github.com/aspnet/Hosting/blob/dev/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs but saw in the code that it was disabled because of #1214 and so I didn't add upon it. If anyone has any idea how to add tests for this behavior I am all ears and will add it to this PR.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostExtensions.cs
@@ -45,8 +45,14 @@ namespace Microsoft.AspNetCore.Hosting
             {
                 AttachCtrlcSigtermShutdown(cts, done, shutdownMessage: string.Empty);
 
-                await host.WaitForTokenShutdownAsync(cts.Token);
-                done.Set();
+                try
+                {
+                    await host.WaitForTokenShutdownAsync(cts.Token);
+                }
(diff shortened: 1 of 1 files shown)


Tratcher:
Thanks

PR #1434, opened 2018-05-24
Always use .exe extension for desktop targeting projects

pranavkm:


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs
```diff
@@ -82,8 +82,17 @@ protected async Task<(Uri url, CancellationToken hostExitToken)> StartSelfHostAs
                 var executableName = string.Empty;
                 var executableArgs = string.Empty;
                 var workingDirectory = string.Empty;
-                var executableExtension = DeploymentParameters.ApplicationType == ApplicationType.Portable ? ".dll"
-                    : (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ".exe" : "");
+                string executableExtension;
+
+                if (DeploymentParameters.RuntimeFlavor == RuntimeFlavor.Clr ||
```
pranavkm: This lets you publish desktop projects as portable (i.e. without specifying a rid). Right now, we assume desktop projects must be standalone

Tratcher: (offline discussion) We'd rather throw if you try to run net461 as Portable.

Issue #1431, opened 2018-05-23
Allow running IIS Express tests without publishing


Tratcher:
IIS Express functional tests always publish first. Publish is slow and it makes the tests take much longer. While publish is required for some scenarios (Core+Standalone), for others it would be faster to run from the pre-built binaries. This can be done via copying the web.config and re-writing it. Self-host now does in place execution where possible, it's just not dependent on the web.config.

Tratcher:
Metrics:
Self-host test that require publish are 3-10 seconds slower than their non-published counterparts.
IIS Express hello world tests take 8-21 seconds today where publish accounts for 5-11 seconds.

Tratcher:
This is working out nicely. Unpublished tests now take about 3 seconds, and most of that time is spent on the first request:
```
| [2018-05-24T09:31:10] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Information: iisexpress stdout: Request started: "GET" http://localhost:3032/
| [2018-05-24T09:31:13] Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer Information: iisexpress stdout: Request ended: http://localhost:3032/ with HTTP status 200.0
```
Half of ServerTest's HelloWorld tests can be run without publishing. All of the other ServerTests can be run without publishing.

PR #1435, opened 2018-05-25
Allow running some IIS Express variants without publishing

Tratcher:
#1431 Some test variants don't require publishing, especially the defaults Core+Portable and Net+Standalone for x64.  x86 net461 and .NET Core Standalone still need to publish.
The average IIS Express test has been reduced from 9s to 3s.

I also changed to providing a config file and site name by default. Most tests do not need to define their own config.

Published tests use their own web.config as before, but unpublished tests embed their settings inside the host.config.


Tratcher:
Moving to a feature branch.

PR #1438, opened 2018-05-29
Add TestPortHelper class for tests

HaoK:
Needed for https://github.com/aspnet/IISIntegration/pull/874/files

@Tratcher


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/IISExpressDeployer.cs
```diff
@@ -71,7 +71,7 @@ private async Task<(Uri url, CancellationToken hostExitToken)> StartIISExpressAs
                 var port = uri.Port;
                 if (port == 0)
                 {
-                    port = TestUriHelper.GetNextPort();
+                    port = TestPortHelper.GetNextPort();
```
Tratcher: Call the other API if the uri scheme is https.

PR #1448, opened 2018-06-06
Move AddCommandLine down in StartupFullControl

guardrex:
@Tratcher Looks like this is the only one over here that doesn't match ...

1. env vars
2. settings file
3. command line


Tratcher:
Can you rebase on dev? I tried moving the PR but it pulled in other changes.

Issue #1451, opened 2018-06-08
Flaky Test: Hang during ApplicationDeployer.DotnetPublish()


mikeharder:
On my Win10 Hyper-V VM, `IISIntegration.FunctionalTests` frequently hangs at the following callstack:

```
System.Private.CoreLib.dll!System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
System.Private.CoreLib.dll!System.Threading.Tasks.Task.SpinThenBlockingWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
System.Private.CoreLib.dll!System.Threading.Tasks.Task.InternalWaitCore(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
System.Private.CoreLib.dll!System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(System.Threading.Tasks.Task task)	Unknown
System.Diagnostics.Process.dll!System.Diagnostics.AsyncStreamReader.WaitUtilEOF()	Unknown
System.Diagnostics.Process.dll!System.Diagnostics.Process.WaitForExitCore(int milliseconds)	Unknown
Microsoft.AspNetCore.Server.IntegrationTesting.dll!Microsoft.AspNetCore.Server.IntegrationTesting.ApplicationDeployer.DotnetPublish(string publishRoot)	Unknown
Microsoft.AspNetCore.Server.IntegrationTesting.dll!Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer.DeployAsync()	Unknown
System.Private.CoreLib.dll!System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start<Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer.<DeployAsync>d__7>(ref Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer.<DeployAsync>d__7 stateMachine)	Unknown
Microsoft.AspNetCore.Server.IntegrationTesting.dll!Microsoft.AspNetCore.Server.IntegrationTesting.IISExpressDeployer.DeployAsync()	Unknown
[...]

PR #1453, opened 2018-06-09
Pass timeout to WaitForExit() on "dotnet publish"

mikeharder:
* Addresses https://github.com/aspnet/Hosting/issues/1451

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
@@ -134,8 +134,8 @@ namespace Microsoft.AspNetCore.Server.IntegrationTesting
                 //    the parent process is killed and ignores child processes (https://stackoverflow.com/a/37983587/102052).
                 // 
                 // 2. If "dotnet publish" does hang indefinitely for some reason, tests should fail fast with an error message.
-                const int _timeoutMinutes = 5;
-                if (hostProcess.WaitForExit(milliseconds: _timeoutMinutes * 60 * 1000))
+                const int timeoutMinutes = 5;
+                if (hostProcess.WaitForExit(milliseconds: timeoutMinutes * 60 * 1000))
                 {
                     if (hostProcess.ExitCode != 0)
                     {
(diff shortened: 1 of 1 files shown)


BrennanConroy:
FYI, you have to call `WaitForExit` without a timeout after using the timeout if you want to guarantee that the output stream has completed
See https://msdn.microsoft.com/en-us/library/ty0d8k56(v=vs.110).aspx#Anchor_2 for reference

mikeharder:
@BrennanConroy: Thanks.  I will try that, but if the call to `WaitForExit()` still blocks on child processes even after `WaitForExit(milliseconds)` has returned, we may not be able to use it.  In practice, I've never had issues with the output stream after `WaitForExit(milliseconds)`.

BrennanConroy:
I did the other day which is why I know :D

src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs
```diff
@@ -127,11 +127,26 @@ protected void DotnetPublish(string publishRoot = null)
...
-
-                if (hostProcess.ExitCode != 0)
+                // A timeout is passed to Process.WaitForExit() for two reasons:
+                // 
+                // 1. Without a timeout, WaitForExit() blocks until child processes are killed, which can cause hangs due to
+                //    MSBuild NodeReuse child processes started by dotnet.exe.  With a timeout, WaitForExit() returns when
+                //    the parent process is killed and ignores child processes (https://stackoverflow.com/a/37983587/102052).
+                // 
+                // 2. If "dotnet publish" does hang indefinitely for some reason, tests should fail fast with an error message.
+                const int _timeoutMinutes = 5;
```
Tratcher: inline const int and named with an underscore? I haven't seen that style used anywhere in our projects. Normally it'd be class level, camel cased, no underscore.

mikeharder: I think method-level is better if the const is only needed in a single method.  Removed the leading underscore to follow convention for local variables.

mikeharder:
I verified that `WaitForExit()` still blocks after `WaitForExit(milliseconds)`:

https://github.com/mikeharder/WaitForExit

So we'll need to just call the latter.  At least on Windows, it seems impossible to guarantee the streams have completed without waiting for child processes to exit.  Relevant code:

https://github.com/dotnet/corefx/blob/e567f83941c7eeba852d024aa00817d00525406c/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Windows.cs#L161-L162

I did find that `WaitForExit()` only blocks when called from an XUnit test, not when called from a non-XUnit console app.  Likely related to XUnit SyncContext.

PR #1461, opened 2018-06-13
Inherit web-specific IHostingEnvironment and IApplicationLifetime from the Generic Host ones

### LINKED ISSUE #1447: Make web specific IHostingEnvironment and IApplicationLifetime inherit from generic host versions
We made the concrete implementations work this way so that you could write types that took in IHostingEnvironment and IApplicationLifetime but we didn't do it for the interfaces themselves. If you have methods that take the generic IHostingEnvironment and IApplicationLifetime, then you can't just pass the implementation you got from the web host into these methods e.g.

Library

```C#
public static class CustomEnvironmentExtensions
{
    public static bool IsCustomDevelopment(this Extensions.Hosting.IHostingEnvironment env) { ... }
}
```

ASP.NET Core app

```C#
public void Configure(IHostingEnvironment env)
{
    // This is a compile error
    if (env.IsCustomDevelopment())
    {
        
    }
}
```

See https://github.com/aspnet/Hosting/issues/1218#issuecomment-394956520 for more info.

henkmollema:
Resolves #1447 and contributes to #1218.

/cc @davidfowl

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingEnvironment.cs
+++ b/src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingEnvironment.cs
@@ -8,20 +8,8 @@ namespace Microsoft.AspNetCore.Hosting
     /// <summary>
     /// Provides information about the web hosting environment an application is running in.
     /// </summary>
-    public interface IHostingEnvironment
+    public interface IHostingEnvironment : Extensions.Hosting.IHostingEnvironment
     {
-        /// <summary>
-        /// Gets or sets the name of the environment. The host automatically sets this property to the value
-        /// of the "ASPNETCORE_ENVIRONMENT" environment variable, or "environment" as specified in any other configuration source.
-        /// </summary>
-        string EnvironmentName { get; set; }
-
-        /// <summary>
-        /// Gets or sets the name of the application. This property is automatically set by the host to the assembly containing
-        /// the application entry point.
-        /// </summary>
-        string ApplicationName { get; set; }
-
         /// <summary>
         /// Gets or sets the absolute path to the directory that contains the web-servable application content files.
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.Hosting.Abstractions/IHostingEnvironment.cs
```diff
@@ -8,20 +8,8 @@ namespace Microsoft.AspNetCore.Hosting
     /// <summary>
     /// Provides information about the web hosting environment an application is running in.
     /// </summary>
-    public interface IHostingEnvironment
+    public interface IHostingEnvironment : Extensions.Hosting.IHostingEnvironment
     {
-        /// <summary>
-        /// Gets or sets the name of the environment. The host automatically sets this property to the value
```
henkmollema: These comments might provide valuable, web-specific information which gets lost this way.

alexk8:
@henkmollema , 
I think, at this point some previous changes should also be undone (see commit
/aspnet/Hosting/commit/a78b9c7490846e60d37e14609921a1ebdc201fdc )
Particularly these:
src/Microsoft.AspNetCore.Hosting/Internal/ApplicationLifetime.cs
 +    public class ApplicationLifetime : IApplicationLifetime, Extensions.Hosting.IApplicationLifetime

src/Microsoft.AspNetCore.Hosting/Internal/HostingEnvironment.cs
 +    public class HostingEnvironment : IHostingEnvironment, Extensions.Hosting.IHostingEnvironment

henkmollema:
@alexk8 thanks for the heads up. I've updated the PR. 😄

henkmollema:
Thanks @davidfowl! 👍

natemcmaster:
@davidfowl I think we should revert this. This is a breaking change for anyone explicitly implementing the interface. This is causing compiler errors in aspnet/Metapackage because `Microsoft.AspNetCore.Hosting.IHostingEnvironment.ContentRootPath` no longer exists.

Tratcher:
It is a binary break, but an explicit interface implementation should still inherit methods from a parent interface. Is this a doc xml specific issue?

henkmollema:
This code will break indeed:

```cs
public class HostingEnvironment : IHostingEnvironment
{
    string IHostingEnvironment.ContentRootPath { get; set; }
}
```

> 'HostingEnvironment.ContentRootPath' in explicit interface declaration is not a member of interface.

It should be

```cs
string Extensions.Hosting.IHostingEnvironment.ContentRootPath { get; set; }
```

XML docs like `<see cref="IHostingEnvironment.ContentRootPath"/>` are broken as well.

davidfowl:
OK, lets postpone this change to 3.0.

Also, API check should explode here /cc @javiercn

henkmollema:
I've created a revert-PR: #1467.

natemcmaster:
I opened https://github.com/aspnet/BuildTools/issues/695 to improve our API check tool to handle this scenario.

davidfowl:
Thanks @natemcmaster

PR #1480, opened 2018-07-01
Change List<Action> to Action for backing fields in WebHostBuilder

### LINKED ISSUE #1479: Change List<Action> to Action for backing fields in WebHostBuilder
https://github.com/aspnet/Hosting/blob/f9d145887773e0c650e66165e0c61886153bcc0b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L28

Every delegate declared with the `delegate` keyword in C# is in fact a `MulticastDelegate`, so unless there's a specific need, constructs such as `List<Action>` are unnecessary and add overhead.  

```csharp
using BenchmarkDotNet.Attributes;
using System;
using System.Collections.Generic;

namespace DelegateDemos
{
    [DisassemblyDiagnoser]
    [MemoryDiagnoser]
    public class ListOfDelegatesVsMulticastDelegates
    {
        private Action<string> _multicastDelegate;
        private List<Action<string>> _listOfAction;
        public string _backingStore;
        public string _anotherBackingStore;

        private void SaveInBackingStore(string value)
        {
            _backingStore = value;
        }

        private void SaveInAnotherBackingStore(string value)
        {
            _anotherBackingStore = value;
        }

        [GlobalSetup]
        public void Setup()
        {
            _multicastDelegate += SaveInBackingStore;
            _multicastDelegate += SaveInAnotherBackingStore;

            _listOfAction = new List<Action<string>> { SaveInBackingStore, SaveInAnotherBackingStore };
        }

        [Benchmark]
        public void InvokeMulticastDelegate()
        {
            _multicastDelegate?.Invoke(string.Empty);
        }

        [Benchmark(Baseline = true)]
        public void InvokeListOfAction()
        {
            foreach (var action in _listOfAction)
            {
                action(string.Empty);
            }
        }
    }
}
```

``` ini

BenchmarkDotNet=v0.10.14, OS=Windows 10.0.15063.1029 (1703/CreatorsUpdate/Redstone2)
Intel Core i7-5600U CPU 2.60GHz (Broadwell), 1 CPU, 4 logical and 2 physical cores
Frequency=2533199 Hz, Resolution=394.7578 ns, Timer=TSC
.NET Core SDK=2.1.301
  [Host]     : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT
  DefaultJob : .NET Core 2.0.7 (CoreCLR 4.6.26328.01, CoreFX 4.6.26403.03), 64bit RyuJIT


```
|                  Method |     Mean |     Error |    StdDev | Scaled | ScaledSD | Allocated |
|------------------------ |---------:|----------:|----------:|-------:|---------:|----------:|
| InvokeMulticastDelegate | 14.28 ns | 0.2591 ns | 0.2297 ns |   0.54 |     0.02 |       0 B |
|      InvokeListOfAction | 26.37 ns | 0.7487 ns | 1.0496 ns |   1.00 |     0.00 |       0 B |

alfredmyers:
Fixes #1479

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
+++ b/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
@@ -25,13 +25,13 @@ namespace Microsoft.AspNetCore.Hosting
     public class WebHostBuilder : IWebHostBuilder
     {
         private readonly HostingEnvironment _hostingEnvironment;
-        private readonly List<Action<WebHostBuilderContext, IServiceCollection>> _configureServicesDelegates;
+        private Action<WebHostBuilderContext, IServiceCollection> _configureServices;
 
         private IConfiguration _config;
         private WebHostOptions _options;
         private WebHostBuilderContext _context;
         private bool _webHostBuilt;
-        private List<Action<WebHostBuilderContext, IConfigurationBuilder>> _configureAppConfigurationBuilderDelegates;
+        private Action<WebHostBuilderContext, IConfigurationBuilder> _configureAppConfigurationBuilder;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WebHostBuilder"/> class.
@@ -39,8 +39,6 @@ namespace Microsoft.AspNetCore.Hosting
         public WebHostBuilder()
         {
             _hostingEnvironment = new HostingEnvironment();
-            _configureServicesDelegates = new List<Action<WebHostBuilderContext, IServiceCollection>>();
-            _configureAppConfigurationBuilderDelegates = new List<Action<WebHostBuilderContext, IConfigurationBuilder>>();
 
             _config = new ConfigurationBuilder()
                 .AddEnvironmentVariables(prefix: "ASPNETCORE_")
@@ -111,12 +109,7 @@ namespace Microsoft.AspNetCore.Hosting
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
         public IWebHostBuilder ConfigureServices(Action<WebHostBuilderContext, IServiceCollection> configureServices)
         {
-            if (configureServices == null)
-            {
-                throw new ArgumentNullException(nameof(configureServices));
-            }
-
-            _configureServicesDelegates.Add(configureServices);
+            _configureServices += configureServices;
             return this;
         }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -111,12 +109,7 @@ public IWebHostBuilder ConfigureServices(Action<IServiceCollection> configureSer
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
         public IWebHostBuilder ConfigureServices(Action<WebHostBuilderContext, IServiceCollection> configureServices)
         {
-            if (configureServices == null)
```
davidfowl: What happens if you pass null?

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
```diff
@@ -131,12 +124,7 @@ public IWebHostBuilder ConfigureServices(Action<WebHostBuilderContext, IServiceC
         /// </remarks>
         public IWebHostBuilder ConfigureAppConfiguration(Action<WebHostBuilderContext, IConfigurationBuilder> configureDelegate)
         {
-            if (configureDelegate == null)
-            {
-                throw new ArgumentNullException(nameof(configureDelegate));
-            }
-
-            _configureAppConfigurationBuilderDelegates.Add(configureDelegate);
+            _configureAppConfigurationBuilder += configureDelegate;
```
Tratcher: Does this guarantee the same ordering as the List?

alfredmyers: [Yes](https://docs.microsoft.com/en-us/dotnet/api/system.multicastdelegate?redirectedfrom=MSDN&view=netframework-4.7.2#remarks)

src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs
alfredmyers: The reason you'd want to check for null in the current version of the code is so you don't have to check for it later when enumerating the list for invoking the delegates when a null entry would throw a NullReferenceException.
But Action<T> is a MulticastDelgate and as such it already embeds the functionality of invoking multiple delegates. It also takes care of null, so you only have to check for null on the "head" of the mulicast delegate.

https://docs.microsoft.com/en-us/dotnet/api/system.delegate.combine?view=netcore-2.1#System_Delegate_Combine_System_Delegate_System_Delegate_

PR #1492, opened 2018-07-16
Added Ability to GenericHost to be run as Windows Service

mesuttalebi:
Added a new Copy of WebHostService and change it to accept IHost to be able to run with new Generic host feature. also updated WebHostWindowsServiceExtensions.cs and added an overload for RunAsService to extend IHost.


Tratcher:
We have a different plan for generic host in windows services. See the prototype here:
https://github.com/aspnet/Hosting/blob/master/samples/GenericHostSample/ServiceBaseControlled.cs
https://github.com/aspnet/Hosting/blob/master/samples/GenericHostSample/ServiceBaseLifetime.cs

davidfowl:
Closing this as a result. We plan to ship this in 3.0

PR #1493, opened 2018-07-16
Remove IISExpress deployer

pakrym:
And make useful things public.


pakrym:
> WebSockets?

Seems like it only uses httpsys.

Issue #1299, opened 2017-12-20
Urls not read from appsettings.json


b-stime:
See https://github.com/aspnet/Hosting/pull/1298 for my quick attempt at a fix via the GitHub web interface.

b-stime:
@muratg
@Tratcher 

> What are you exactly trying to fix? Is there a bug filed for the issue you're targeting here?

> This code doesn't compile. _startup.Configure is a method, not an IConfiguration instance. I'm closing this and we can discuss it issue over at #1299.

I guess this may've already been discussed at some length\* but it continues to burn new users (e.g., me)...

The issue is that you don't appear to be able to configure the webhost port via a config file or command line out of the box when you create a project via:

$ dotnet new razor

The WebHostBuilder already parses appsettings.json and the command line args and hydrates the Startup object IConfigure so my patch attempt seemed like an easy way to reduce astonishment.

\*       https://stackoverflow.com/a/45854704/93345
          // https://github.com/aspnet/Hosting/issues/1148
            // https://github.com/aspnet/KestrelHttpServer/issues/1851

Tratcher:
[Rambling investigation]

The command line issue was fixed here: https://github.com/aspnet/MetaPackages/issues/221. The workaround is here: https://github.com/aspnet/MetaPackages/issues/221#issuecomment-335207431

A similar workaround works for appsettings.json as well. 

The new Kestrel config file support also reads from appsettings.json.
https://github.com/aspnet/KestrelHttpServer/issues/1290#issuecomment-353416087

So the remaining gap is reading Urls from appsettings.json. Right now it's read from the wrong config layer. Hosting builds an initial configuration and services so it has enough information to build the rest of the app, like the Environment. But it also reads Urls from this config.
https://github.com/aspnet/Hosting/blob/cd3f58bed7234ed5035a995c8688dccd6706b584/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L290

The other config is built seperately and added to the app services: https://github.com/aspnet/Hosting/blob/cd3f58bed7234ed5035a995c8688dccd6706b584/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L246-L256

Option A to fix this would be to resolve the IConfiguration from the app services like how it resolves the server except as optional.
https://github.com/aspnet/Hosting/blob/cd3f58bed7234ed5035a995c8688dccd6706b584/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L284
[...]

PR #1495, opened 2018-07-17
Fixed bug where 'ServerUrlsKeys' was not read from user configuration.

flaviofak:
There was no way to read URLs from 'appsettings.json' file because the configuration object passed to the WebHost was not the object the user built.


Tratcher:
This has been discussed previously and we opted not to change this behavior. See https://github.com/aspnet/Hosting/issues/1299

PR #1502, opened 2018-07-25
Shorten shared sources for MAX_PATH issue on VSTS

HaoK:
@natemcmaster 

I believe EFCore is the only other place that uses this package, MVC has a reference to WebHostBuilderFactory but it gets that reference thru TestHost I think


natemcmaster:
To get the EF Core repo to pass, you can push these changes to a  feature/* branch. This should create a build of the package with the new name and upload to myget. In EF, you can change the deps.props file to the version created from your feature branch.

natemcmaster:
Also, this should probably target release/2.2 instead of master. The EF PR targets release/2.2

HaoK:
Replaced by https://github.com/aspnet/Hosting/pull/1505

PR #1508, opened 2018-07-26
More robust username discovery on Linux in NginxDeployer

natemcmaster:
VSTS does not have the LOGNAME env var set. This adds some fallbacks (USER and USERNAME) and `whoami` if they are not present.

@HaoK this unblocks https://github.com/aspnet/ServerTests/pull/126

### DIFF
--- a/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/NginxDeployer.cs
+++ b/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/NginxDeployer.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Net.Http;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/NginxDeployer.cs
```diff
@@ -77,16 +78,48 @@ public NginxDeployer(DeploymentParameters deploymentParameters, ILoggerFactory l
...
+                {
+                    StartInfo =
+                    {
+                        FileName = "whoami",
+                        RedirectStandardOutput = true,
+                    }
+                })
+                {
+                    process.Start();
+                    process.WaitForExit(10_000);
```
HaoK: I had to google what this number syntax was :)

PR #1511, opened 2018-07-30
Add TestApplication to allow publish output caching

pakrym:
Copying files is much faster than publishing


src/Microsoft.AspNetCore.Server.IntegrationTesting/TestApplication.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            foreach (DirectoryInfo directoryInfo in source.GetDirectories())
+            {
+                CopyFiles(directoryInfo, target.CreateSubdirectory(directoryInfo.Name), logger);
+            }
+
+            logger.LogDebug($"Processing {target.FullName}");
+            foreach (FileInfo fileInfo in source.GetFiles())
+            {
+                logger.LogDebug($"  Copying {fileInfo.Name}");
```
jkotalik: This log seem too frequent.

pakrym: I thought so too. But it allowed me to see that we publish "refs" directory every time. So maybe it's not that bad?

src/Microsoft.AspNetCore.Server.IntegrationTesting/Common/DeploymentParameters.cs
```diff
@@ -99,6 +99,8 @@ public DeploymentParameters(DeploymentParameters parameters)
             }
         }
 
+        public TestApplication TestApplication { get; set; }
```
Tratcher: This object model is circular. DeploymentParamenters has a TestApplication, but TestApplication methods take DeploymentParmameters.

src/Microsoft.AspNetCore.Server.IntegrationTesting/TestApplication.cs
jkotalik: Maybe just print directories added?

src/Microsoft.AspNetCore.Server.IntegrationTesting/PublishedApplication.cs
```diff
@@ -0,0 +1,23 @@
...
+        public string Path { get; }
+
+        public PublishedApplication(string path)
+        {
+            Path = path;
+        }
+
+        public void Dispose()
+        {
+            RetryHelper.RetryOperation(() => Directory.Delete(Path, true), _ => { });
```
Tratcher: You lost the failure log

pakrym: Yeah, I was considering adding Dispose overload with Logger

Tratcher: Or passing the logger into the ctor.

pakrym: That's harder in cached publisher case.

pakrym: I lied, it's easy. Done.

Issue #1503, opened 2018-07-25
Flaky test: ShutdownTestRun


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=514188&buildTypeId=XPlat_Linux_UbuntuUniverseCoreCLR) occasionally with the following error:
```
System.Exception : dotnet publish exited with exit code : 1
   at Microsoft.AspNetCore.Server.IntegrationTesting.ApplicationDeployer.DotnetPublish(String publishRoot) in /_/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/ApplicationDeployer.cs:line 150
   at Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer.DeployAsync() in /_/src/Microsoft.AspNetCore.Server.IntegrationTesting/Deployers/SelfHostDeployer.cs:line 55
   at Microsoft.AspNetCore.Hosting.FunctionalTests.ShutdownTests.ExecuteShutdownTest(String testName, String shutdownMechanic) in /_/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs:line 70
   at Microsoft.AspNetCore.Hosting.FunctionalTests.ShutdownTests.ShutdownTestRun() in /_/test/Microsoft.AspNetCore.Hosting.FunctionalTests/ShutdownTests.cs:line 33
--- End of stack trace from previous location where exception was thrown ---
------- Stdout: -------
| [0.025s] TestLifetime Information: Starting test ShutdownTestRun at 2018-07-25T01:54:30
| [0.001s] TestLifetime Information: Starting test ExecuteShutdownTest at 2018-07-25T01:54:31
| [0.014s] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: Deploying [Variation] :: ServerType=Kestrel, Runtime=CoreClr, Arch=x64, BaseUrlHint=, Publish=True
| [0.023s] Microsoft.AspNetCore.Server.IntegrationTesting.SelfHostDeployer Information: SET ASPNETCORE_ENVIRONMENT=Shutdown
[...]

PR #1518, opened 2018-08-10
Add test build targets to ensure version alignment

Tratcher:
#1503 Trying the pattern from scaffolding to force dependencies to stay in sync during a CI build.
https://github.com/aspnet/Scaffolding/blob/master/test/Directory.Build.targets

Unable to repro the failure locally, but this doesn't seem to make anything worse.

### DIFF
--- a/test/Directory.Build.targets
+++ b/test/Directory.Build.targets
@@ -1,0 +1,46 @@
+<Project>
+  <Import Project="..\Directory.Build.targets" />
+
+  <ItemGroup>
+    <None Include="$(TestProjectTemplateFile)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <TestProjectGenInputs Include="$(ProjectAssetsFile)" />
+    <TestProjectGenInputs Include="$(MSBuildAllProjects)" />
+    <TestProjectGenInputs Include="$(TestProjectTemplateFile)" />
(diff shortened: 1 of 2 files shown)


test/Directory.Build.targets
```diff
@@ -0,0 +1,46 @@
...
+    <TestProjectGenInputs Include="$(ProjectAssetsFile)" />
+    <TestProjectGenInputs Include="$(MSBuildAllProjects)" />
+    <TestProjectGenInputs Include="$(TestProjectTemplateFile)" />
+    <TestProjectGenInputs Include="$(MSBuildThisFileDirectory)..\build\dependencies.props" />
+    <TestProjectGenInputs Include="$(MSBuildThisFileDirectory)..\build\sources.props" />
+    <TestProjectGenInputs Include="$(DotNetPackageVersionPropsPath)" Condition=" '$(DotNetPackageVersionPropsPath)' != '' " />
+    <TestProjectGenInputs Include="$(DotNetRestoreSourcePropsPath)" Condition=" '$(DotNetRestoreSourcePropsPath)' != '' " />
+  </ItemGroup>
+
+  <Target Name="GenerateTestProject" BeforeTargets="CoreCompile" Inputs="@(TestProjectGenInputs)" Outputs="$(GeneratedTestProjectFile)">
```
natemcmaster: I don't see where `$(GeneratedTestProjectFile)` is defined.

natemcmaster:
Turns out node reuse is the problem.

PR #1531, opened 2018-08-27
Don't hold onto the HttpContext in the HostingLogScope

davidfowl:
- Multiple things capture the ExecutionContext, reduce the changes of improperly rooting the HttpContext when we only need a few properties from the request.

Intentional memory leak

```C#
using System.Threading;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Http;

namespace WebApplication23
{
    public class Startup
    {
        {
            
        }

        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            var cts = new CancellationTokenSource();

            app.Run(async (context) =>
            {
                cts.Token.Register(() =>
                {

                });

                await context.Response.WriteAsync("Hello World!");
            });
        }
    }
}
```

This code will capture every HttpContext in the logging scope:

image
image

It's rooted by the scope which is being captured in the CancellationToken callback.

cc @benaadams


benaadams:
Getting gummed up everywhere 
image

PR #1548, opened 2018-09-18
Move all request diagnostics into an IStartupFilter

davidfowl:
- This should enable people to plug in a custom request ID, logging scope, or any other diagnostic thing, without adding extra extensibility.

The one potential breaking change is for things like Autofac.Multitenant (https://github.com/autofac/Autofac.AspNetCore.Multitenant/blob/6ab27ab5312159da2f50e4d8f88dfbca670fb09e/src/Autofac.Integration.AspNetCore.Multitenant/AutofacMultitenantWebHostBuilderExtensions.cs#L65). They inject themselves as the first Startup filter which means request diagnostics would run after instead of before. 

cc @tillig

There are probably other subtle changes to when things run but I'm hoping those specific details don't matter too much. This enables extensibility through existing mechanisms (like middleware) instead of adding more specific hooks to hosting.

### DIFF
--- a/src/Microsoft.AspNetCore.Hosting/Internal/RequestDiagnosticsMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Hosting/Internal/RequestDiagnosticsMiddleware.cs
@@ -1,0 +1,43 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    public class RequestDiagnosticsMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private HostingApplicationDiagnostics _diagnostics;
+
+        public RequestDiagnosticsMiddleware(RequestDelegate next, DiagnosticListener diagnosticListener, ILogger logger)
+        {
+            _next = next;
+            _diagnostics = new HostingApplicationDiagnostics(logger, diagnosticListener);
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            var context = new HostingApplication.Context();
+            _diagnostics.BeginRequest(httpContext, ref context);
+
+            try
+            {
+                await _next.Invoke(httpContext);
+
+                _diagnostics.RequestEnd(httpContext, null, context);
+            }
+            catch (Exception exception)
+            {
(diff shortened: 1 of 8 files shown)


src/Microsoft.AspNetCore.Hosting/Internal/RequestDiagnosticsMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            try
+            {
+                await _next.Invoke(httpContext);
+
+                _diagnostics.RequestEnd(httpContext, null, context);
+            }
+            catch (Exception exception)
+            {
+                _diagnostics.RequestEnd(httpContext, exception, context);
```
Tratcher: Moving this to middleware loses quite a bit of fidelity for the error case, and a little fidelity even for normal requests.

When an exception is thrown before the first write:
2.1: the server changes the status code to 500, clears the headers, send the response, and then logs RequestEnd with a 500 and no headers.
New: RequestEnd is logged with a 200 status code and standard response headers, then the server changes the status code to 500, clears the headers, send the response.

This is too breaking, the logs will be much less accurate.

This also makes the timing less accurate as the server used to do more work before logging RequestEnd (flushes headers, send the final response chunk, etc.)

tillig:
Sounds like there will be a race condition for something at the start no matter what we do. I may be able to iterate through the list of startup filters and insert at the point _just before_ I find one that is called `AutoRequestServicesStartupFilter`... but that's also a potential issue if names change or the pipeline changes how request services get populated.

We can probably update the logic to search for `AutoRequestServicesStartupFilter` and if it finds it, great, insert just before; if not, insert at zero.

I wonder if there are other race conditions like that or different/better ways to extend things like the request services setup. Hmmmm.

In any case, I can look at an enhancement to adjust as needed. Thanks for the heads up.

src/Microsoft.AspNetCore.Hosting/Internal/RequestDiagnosticsMiddleware.cs
davidfowl: Would running request end in OnCompleted help here? We don’t flow the exception there but it could be captured in the middleware.

Tratcher: OnCompleted would be better timing wise, that's [immediately before DisposeContext](https://github.com/aspnet/KestrelHttpServer/blob/edc19354751dd43dc48e71cdfe192147e40122c8/src/Kestrel.Core/Internal/Http/HttpProtocol.cs#L602-L607). You will still miss out on logging a few exceptions from places like [OnStarting](https://github.com/aspnet/KestrelHttpServer/blob/edc19354751dd43dc48e71cdfe192147e40122c8/src/Kestrel.Core/Internal/Http/HttpProtocol.cs#L700-L716) and [ContentLength](https://github.com/aspnet/KestrelHttpServer/blob/edc19354751dd43dc48e71cdfe192147e40122c8/src/Kestrel.Core/Internal/Http/HttpProtocol.cs#L894) though.


It would require additional allocations per request to register and pass state though.

PR #1552, opened 2018-09-22
Dereference the HttpContext when it's no longer needed

davidfowl:
- No need to hold onto it


davidfowl:
> Why bother? The HttpContext and the RequestServicesFeature have almost exactly the same lifetime.

I was looking at dumps and noticed this. What's interesting is that when looking at the dump (when using kestrel), it is stored per connection. Even though it *does* get cleared out between, it looks pretty weird. It also doesn't hurt, but you're right we *do* have a bunch of features other features that are in the same boat. 

> Are you going to clear the _scopeFactory while you're at it?

Yes, I should

davidfowl:
Yes thats what I said 😄

PR #1558, opened 2018-10-04
Fix misspelled word in comment

guardrex:


Tratcher:
Hehe. Why master vs release/2.2?

guardrex:
Ryan was just explaining to me the diff between "branches" and "tags" for preview releases.

This will be another concept on engineering repos to understand.

Tratcher:
😁 release/2.2 is where we're finishing up all of the 2.2 work. Master has already moved on to 3.0 work.

guardrex:
Right, but I guessed (wrong, I guess 😄) that the misspelled word would still exist on master if this change only goes to the release/2.2 branch. ........ so you take release/2.2 updates to master? Is that how it goes?

Tratcher:
Yes, make the fix in the 'oldest' branch first and merge forwards.

PR #1580, opened 2018-11-10
Added support for generic host based IWebHostBuilder

### LINKED ISSUE #1544: Feature gaps to make replace WebHostBuilder with HostBuilder
- [ ] - Add support for `IHostingStartup`
- [ ] - Add support for `Startup`

davidfowl:
- This adds an implementation of IWebHostBuilder as a facade over the IHostBuilder.
This removes the 2 container issue by executing the Startup.ConfigureServies and Startup.ConfigureContainer inline as part of building the IHostBuilder.
- The implementation is highly compatible implementation since it exposes the same IWebHostBuilder interface.
Existing extensions mostly work.
- There are some caveats with this approach.
    - Injecting services into Startup is not extremely constrained to the
    services availble on HostBuilderContext. This includes the IHostingEnvironment
    and the IConfiguration.
    - IStartup is broken when using this pattern because it isn't composable.
    - The IStartupConfigureServicesFilter and IStartupConfigureContainer The before
    and after filters added in 2.1 are also broken  because there's a single container (it could maybe be fixed by downcasting and doing something specific on the GenericHostBuilder instance) (@javiercn)
    - Calling into IWebHostBuilder.Build will throw a NotSupportedException since
    this implementation is just a facade over the IHostBuilder.
    - TestServer isn't compatible with this implementation as yet.

Fixes #1544

There are 3 incompatibilities found in running the existing test suite:

- [x] - HostingContextContainsAppConfigurationDuringConfigureLogging => Hosting configuration flows to ConfigureLogging call backs application configuration does not (Won't fix)
- [x] - HostingContextContainsAppConfigurationDuringConfigureServices =>  Hosting configuration flows to ConfigureServices callbacks not application configuration does not (Won't fix)
- [x] - ExternalContainerInstanceCanBeUsedForEverything => Adding an IServiceProviderFactory as a service no longer works. Call IHostBuilder.UseServiceProviderFactory instead. (Won't fix)

This is what we're aiming for with the 3.0 template:

```C#
using System.Threading.Tasks;
using Microsoft.Extensions.Hosting;

namespace SampleApplication
{
    public class Program
    {
        public static async Task Main(string[] args)
        {
            await CreateHostBuilder(args).Build().RunAsync();
        }

        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureWebHostDefaults(builder =>
                {
                    builder.UseStartup<Startup>();
                });
    }
}
```


src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
```diff
@@ -0,0 +1,200 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.StackTrace.Sources;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal class GenericWebHostService : IHostedService
+    {
+        private static readonly string DeprecatedServerUrlsKey = "server.urls";
+
+        public GenericWebHostService(IOptions<GenericWebHostServiceOptions> options,
```
davidfowl: Something interesting here, we should maybe delay resolution of these services until StartAsync. If any of these things fail to resolve, you won't get startup errors in the browser.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -29,15 +29,21 @@ public static IWebHostBuilder Configure(this IWebHostBuilder hostBuilder, Action
 
             var startupAssemblyName = configureApp.GetMethodInfo().DeclaringType.GetTypeInfo().Assembly.GetName().Name;
 
-            return hostBuilder
-                .UseSetting(WebHostDefaults.ApplicationKey, startupAssemblyName)
-                .ConfigureServices(services =>
+            hostBuilder.UseSetting(WebHostDefaults.ApplicationKey, startupAssemblyName);
+
+            // Light up the GenericWebHostBuilder implementation
+            if (hostBuilder is GenericWebHostBuilder genericWebHostBuilder)
```
Tratcher: Rather than sniffing for a specific type can we at least sniff for a new interface like IConfigureStartup with Configure, UseStartup, etc..

davidfowl: Sure but why?

davidfowl: Nah. It’s overkill and can be added at anytime. I would do that if we had something else concrete that would implement. Also I plan to keep this all internal as we’re not trying to expose anything public.

So in the end we’d have a new internal interface implemented and an internal type that would allow us to test these extension methods. I don’t see the value.

davidfowl:
This might be the way we recommend frameworks plug into the generic host. A single top level method with a builder specific to that framework. I'm thinking that Orleans could potentially use the same approach with the SiloBuilder.

cc @ReubenBond @jdom

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
Tratcher: 👃

davidfowl:
cc @tillig 

Just an FYI that this breaks the pattern where you return an IServiceProvider from Startup.ConfigureServices. The preferred pattern will be an IServiceProviderFactory. That makes things compose since the web application configuration isn't the only thing existing. This will require a migration story when moving apps from 2.x to 3.x. Luckily, we're not breaking any of the WebHostBuilder behavior so it can be smooth 😄 .

tillig:
Thanks for the heads-up. I'll make sure our corresponding docs and examples get updated!

davidfowl:
Pinging other DI container authors (feel free to spread the word if I missed people) cc @dadhi @seesharper @ipjohnson @khellang @ENikS

oferns:
@dotnetjunkie might be interested in this conversation

seesharper:
@davidfowl Thanks for letting us know up front about changes like this. 

Correct me if I am wrong, but this is what I can gather from this

*  `IServiceProvider  ConfigureServices` will not be supported anymore.
*  We can now add the `IServiceProviderFactory` to an `IHostBuilder` rather that the `IWebHostBuilder`  https://github.com/seesharper/LightInject.Microsoft.AspNetCore.Hosting/blob/master/src/LightInject.Microsoft.AspNetCore.Hosting/LightInjectWebHostBuilderExtensions.cs

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting.Server;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
```
rynowak: Eww `.Internal` - this is just noise

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
```
rynowak: copyright

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
rynowak: Also folder and namespace haven't even met each other.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
```diff
@@ -0,0 +1,32 @@
...
+            _host = host;
+        }
+
+        public IFeatureCollection ServerFeatures => Services.GetRequiredService<IServer>().Features;
+
+        public IServiceProvider Services => _host.Services;
+
+        public void Dispose() => _host.Dispose();
+
+        public void Start() => _host.Start();
```
rynowak: Is this really right? Is the fact that this class is an implementation of `IWebHost` even significant since users can't call `.Build()`?

rynowak: I suppose this is here so you can resolve it as a service, but if you do - is it right to have these methods do what they do?

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostApplicationLifetime.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Threading;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal class GenericWebHostApplicationLifetime : IApplicationLifetime
+    {
+        private readonly Microsoft.Extensions.Hosting.IApplicationLifetime _applicationLifetime;
+        public GenericWebHostApplicationLifetime(Microsoft.Extensions.Hosting.IApplicationLifetime applicationLifetime)
```
rynowak: You from a year ago is trolling you now 😆

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
rynowak: and everytwhere

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
rynowak: What of these things aren't registered by hosting?

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
```diff
@@ -0,0 +1,197 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Extensions.StackTrace.Sources;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal class GenericWebHostService : IHostedService
+    {
+        private static readonly string DeprecatedServerUrlsKey = "server.urls";
```
rynowak: If this is deprecated do we need it?

src/Microsoft.AspNetCore.Hosting/GenericHost/HostingStartupWebHostBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+        private Action<WebHostBuilderContext, IServiceCollection> _configureServices;
+
+        public HostingStartupWebHostBuilder(GenericWebHostBuilder builder)
+        {
+            _builder = builder;
+        }
+
+        public IWebHost Build()
+        {
+            throw new NotSupportedException();
```
rynowak: Message?

src/Microsoft.AspNetCore.Hosting/GenericHost/HostingStartupWebHostBuilder.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    // We use this type to capture calls to the IWebHostBuilder so the we can properly order calls to 
+    // to GenericHostWebHostBuilder.
+    internal class HostingStartupWebHostBuilder : IWebHostBuilder, ISupportsStartup, ISupportsUseDefaultServiceProvider
+    {
+        private readonly GenericWebHostBuilder _builder;
+        private Action<WebHostBuilderContext, IConfigurationBuilder> _configureConfiguration;
+        private Action<WebHostBuilderContext, IServiceCollection> _configureServices;
```
rynowak: Is it worth clearing these after use? Or will this create more problems?

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
psibernetic: Why wouldn't webhost just call base on build?

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
davidfowl: Yea, I wanted to group the implementation together without sticking .GenericHost in the namespace name.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
davidfowl: Wait, I just realized, if the server can't be resolved we can't capture startup errors and do anything meaningful anyways 😄 . This is fine.

javiercn:
> * The IStartupConfigureServicesFilter and IStartupConfigureContainer The before
>   and after filters added in 2.1 are also broken  because there's a single container (it could maybe be fixed by downcasting and doing something specific on the GenericHostBuilder instance) (@javiercn)

I think I know how we can fix this easily. We can use the new properties dictionary in IHostBuilderContext (should be an abstract class) to tag the filter instances in there and then when ConfigureServices is invoked on the host builder, we simply retrieve the filters from the hostbuildercontext and apply them.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
Tratcher: Yes this should be removed.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
Tratcher: Added them where?

An interface per extension seems overkill as they're all related to Startup/Configure.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
```diff
@@ -0,0 +1,382 @@
...
+                services.TryAddSingleton<IHttpContextFactory, HttpContextFactory>();
+                services.TryAddScoped<IMiddlewareFactory, MiddlewareFactory>();
+                services.TryAddSingleton<IApplicationBuilderFactory, ApplicationBuilderFactory>();
+
+                // Conjure up a RequestServices
+                services.TryAddTransient<IStartupFilter, AutoRequestServicesStartupFilter>();
+                services.TryAddTransient<IServiceProviderFactory<IServiceCollection>, DefaultServiceProviderFactory>();
+
+                // Ensure object pooling is available everywhere.
+                services.TryAddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
```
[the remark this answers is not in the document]
Tratcher: Now is a great time to drop it before anyone takes a dependency on it.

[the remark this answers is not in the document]
davidfowl: We just need routing to add the ObjectPoolProvider https://github.com/aspnet/Routing/blob/7b16053f27ac982eb71317fa8072b55c443aa158/src/Microsoft.AspNetCore.Routing/DependencyInjection/RoutingServiceCollectionExtensions.cs#L36.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
davidfowl: Removed it

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
```diff
@@ -51,8 +57,15 @@ public static IWebHostBuilder UseStartup(this IWebHostBuilder hostBuilder, Type
         {
```
Tratcher: What about the overload here:
https://github.com/aspnet/Hosting/blob/2de091e154cc33a5b3d413572826c523d98462a4/src/Microsoft.AspNetCore.Hosting.Abstractions/HostingAbstractionsWebHostBuilderExtensions.cs#L52-L63

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        public IWebHostBuilder UseSetting(string key, string value)
+        {
+            _config[key] = value;
+            return this;
+        }
+
+        // This exists just so that we can use ActivatorUtilities.CreateInstance on the Startup class
+        private class ServiceProvider : IServiceProvider
```
Tratcher: HostServiceProvider?

Tratcher (changes requested):

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
davidfowl: What about it? This works fine.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
```diff
@@ -0,0 +1,385 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.ObjectPool;
+
+namespace Microsoft.AspNetCore.Hosting.Internal
+{
+    internal class GenericWebHostBuilder : IWebHostBuilder, ISupportsStartup, ISupportsUseDefaultServiceProvider
```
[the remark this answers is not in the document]
davidfowl: Do you have any actionable feedback? Do you see a leak?

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
```diff
@@ -0,0 +1,385 @@
...
+    {
+        private readonly IHostBuilder _builder;
+        private readonly IConfiguration _config;
+        private readonly object _startupKey = new object();
+        private readonly bool _allowBuild;
+
+        private AggregateException _hostingStartupErrors;
+        private HostingStartupWebHostBuilder _hostingStartupWebHostBuilder;
+
+        public GenericWebHostBuilder(IHostBuilder builder, bool allowBuild = true)
```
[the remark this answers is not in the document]
davidfowl: Why does that matter? Nobody can create this class as it's internal.

src/Microsoft.AspNetCore.Hosting/WebHostBuilderExtensions.cs
davidfowl: https://github.com/aspnet/Hosting/pull/1580/files#diff-575646ff9c6b4006e3d4eb641c7c8f01R105

davidfowl: https://github.com/aspnet/Hosting/blob/111d6932a07cb732b304e04025348deba4d2ed75/src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs#L105

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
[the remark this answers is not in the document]
Tratcher: Trying to limit the amount of test code mixed into production code, and to prevent mistakes like defaulting to a test configuration.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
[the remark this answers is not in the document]
Tratcher: It does leak in that it lives beyond the obvious lifespan from the extension method, but since there's only one that shouldn't cause too much trouble. I don't have a specific suggestion here, I was only calling this out as an unusual side-effect of this design approach.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
```diff
@@ -0,0 +1,200 @@
...
+                    }
+                }
+            }
+
+            RequestDelegate application = null;
+
+            try
+            {
+                var builder = ApplicationBuilderFactory.CreateBuilder(Server.Features);
+                Action<IApplicationBuilder> configure = Options.ConfigureApplication ?? _defaultApplication;
```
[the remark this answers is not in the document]
Tratcher: Nevermind, I misread the signature of _defaultApplication, I thought it was throwing a per-request exception.  It's a little strange you wouldn't just throw the exception here rather than running the filters first. Do we have an examples of terminal filters that would would replace the app, or error handling filters that would catch the exception?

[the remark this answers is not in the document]
Tratcher: This scenario wasn't supported by the old WebHost. If you didn't specify startup then the filters never ran. https://github.com/aspnet/Hosting/blob/d7b9fd480765bdc01f06441f308fb288e6001049/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L191-L196

tillig:
> Just an FYI that this breaks the pattern where you return an IServiceProvider from Startup.ConfigureServices. The preferred pattern will be an IServiceProviderFactory.

Currently the only way to build the Autofac container _and keep a reference to it_ if you need to do that is to return the `IServiceProvider` from `Startup.ConfigureServices`. We've seen some interesting use cases where people want to use a single container and host multiple application types (like a self-hosted ASP.NET Core app alongside... uh... something else? Not sure what). That's led to requests for things like the ability to create a pseudo-container in a child lifetime scope.

```
+-------------------------------------------------------+
| root container: shared registrations                  |
| and singletons                                        |
|                                                       |
| +------------------------+ +------------------------+ |
| | child scope: root      | | child scope: root      | |
| | for ASP.NET Core self- | | for background service | |
| | hosted app             | | or something (?)       | |
| +------------------------+ +------------------------+ |
+-------------------------------------------------------+
```

Having the ability to return the `IServiceProvider` means the dev could provide a pre-created scope to the `ConfigureServices` method and return a custom-built `IServiceProvider`.

I gather that in order to do something like this in the new system folks will need a custom `IServiceProviderFactory` or figure out a workaround.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
```diff
@@ -0,0 +1,385 @@
...
+                .AddEnvironmentVariables(prefix: "ASPNETCORE_")
+                .Build();
+
+            _builder.ConfigureHostConfiguration(config =>
+            {
+                config.AddConfiguration(_config);
+
+                // We do this super early but still late enough that we can process the configuration
+                // wired up by calls to UseSetting
+                ExecuteHostingStartups();
```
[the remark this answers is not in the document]
Tratcher: In the WebHostBuilder world you could configure HostingStartup via the host config APIs (UseSetting, UseConfiguration). Those still work for generic web host, but only if you call them directly inside the web host context. As implemented, HostingStartup will not honor host level configuration set via generic hosts's ConfigureHostConfiguration. This is the only feature that ends up using an isolated config.

Mitigation: Before calling ExecuteHostingStartups, call `_config = config.Build()` to at least capture the host config added so far.

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostedService.cs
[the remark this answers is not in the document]
davidfowl: So this behavior is important enough that a test should fail right?

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHostBuilder.cs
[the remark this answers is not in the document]
Tratcher: https://github.com/aspnet/Hosting/pull/1586

src/Microsoft.AspNetCore.Hosting/GenericHost/GenericWebHost.cs
rynowak: put them in one file and use `#region`

poke:
Is there any way we can bring back the functionality that allows us to inject an logger into the `Startup`? It’s very common for applications to log certain things during startup, at a time when the DI container has not been built.

I understand that with this implementation, there is only hard-coded support for `IHostingEnvironment` and `IConfiguration` but I think `ILogger<T>` or `ILoggerFactory` would be another of the few types that actually make sense to get injected there.

I’ve already looked into the `HostBuilder` source for this but unfortunately, logging is just set up [through DI](https://github.com/aspnet/Hosting/blob/05d1a6eb0e051a0cbadc64246f77fe9b836d33a6/src/Microsoft.Extensions.Hosting/HostBuilder.cs#L184) so it will be difficult to get a proper logger factory out of it. But I think logging would be a really valuable addition to those explicit types that are supported in the startup.

Tratcher:
@poke I don't see how it's possible to bring logging back, it fundamentally depends on DI and config.  It's worth noting that logging was already broken (multiple singletons) and one of the reasons we dropped the second container.

poke:
Yeah, I understand that it’s very difficult and not possible in the current design, but I really think that it’s something that is expected to work. So if we don’t end up having a good approach for that, I am sure people will make up a way to achieve this, for example by creating a new (second) logger factory from within their startup. Just like people built the service provider within the `ConfigureServices` method when they needed some random service to configure stuff before it was properly communicated (and supported by utility methods) how to do that properly.

Tratcher:
At least logging can still be injected into Configure, it's only ConfigureServices that's missing it.

davidfowl:
Logging can't really be done because logger providers are themselves in the DI container.

PureKrome:
So if we are logging when we `ConfigureServices` which takes an `ILogger`, then we would need to create a new (Second) logger factory (as @poke said) ?

davidfowl:
> So if we are logging when we ConfigureServices, where a method (in here) takes an ILogger, then we would need to create a new (Second) logger factory (as @poke said) ?

If you're logging today in ConfigureServices then you're in this torn state where you're using another instance of the ILoggerFactory that's configured. What that can lead to is several background threads running for each of the logger factories (for example the console logger and blob storage logger).

Sure. you can create another logger factory that is separate in there but where would you be logging to?

davidfowl:
@tillig Yes, you'd need a custom factory to do that, but you can capture the instance by using the Configure method.

Tratcher (approved):
Update https://github.com/aspnet/Hosting/blob/master/samples/GenericWebHost/ and then I think you're good for now.

davidfowl:
> Update https://github.com/aspnet/Hosting/blob/master/samples/GenericWebHost/ and then I think you're good for 

😍

PureKrome:
> > So if we are logging when we ConfigureServices, where a method (in here) takes an ILogger, then we would need to create a new (Second) logger factory (as @poke said) ?
> 
> If you're logging today in ConfigureServices then you're in this torn state where you're using another instance of the ILoggerFactory that's configured. What that can lead to is several background threads running for each of the logger factories (for example the console logger and blob storage logger).
> 

yep. but can we dispose of this factory (and therefore any bg threads) once we leave the scope? Then there could be issues of flushing, etc ?

> Sure. you can create another logger factory that is separate in there but where would you be logging to?

Console / some 3rd party service 

in the example above i have: 

`services.AddLogging(configure => configure.AddSerilog(dispose: true));`

so the logging sinks are read-in/defined -prior- to the `HostBuilder` getting created. So logging has been added/configured for serilog ... and then I was hoping to use this seril-logger as a MEL `ILogger`. Happy that there's no DI in this bit .. but possible to expose/use currently existing one?

poke:
> Logging can't really be done because logger providers are themselves in the DI container.

I know but I think there still _has_ to be some way to do it. Even if you don’t necessarily need to log stuff out from the `ConfigureServices` method, people will still want to do it (there’s a relatively popular [SO question](https://stackoverflow.com/q/41287648/216074) on this that proves the desire).

And I really don’t want people to end up doing things like this:
```
{
    var provider = services.BuildServiceProvider();
    var logger = provider.GetService<ILogger<Startup>>();

    logger.LogInformation("Adding MVC");
    services.AddMvc();
}
```
But that’s totally going to happen if we don’t have an actual recommendation on what to do instead.

You mentioned on Slack that you are considering a root logger as a singleton, and I think that would be totally fine. Just allow _some_ sane way to get a logger inside of there, so we can direct people to that solution.

davidfowl:
That will be the way to do it regardless, it *will* require another container. It should not be the same service provider that you use the build up the application though. We'll discuss it more.

poke:
Yeah, but since there already is a separate service provider at web host level by default, it just works. It even uses the same configuration and setup even if it’s technically a separate instance. And that would be my expectation with 3.0 too: That there would be a default and easy way that just works, even if this means that the framework will create a separate logger for me.

davidfowl:
"Just works" There's been numerous unsolvable bugs around the fact that we have 2 containers (sometimes 3!). Most of our workarounds have been to not log anything until after the application container is wired up. We want to be out of that world.

davidfowl:
cc @ENikS I can't remember if I mentioned you on this.

Issue #1583, opened 2018-11-12
Add TestHost support for ConfigureWebHost


davidfowl:
Today the TestHost builds and starts the IWebHostBuilder directly in it's constructor https://github.com/aspnet/Hosting/blob/d7b9fd480765bdc01f06441f308fb288e6001049/src/Microsoft.AspNetCore.TestHost/TestServer.cs#L41-L42. This won't work when using ConfigureWebHost. We may need to revisit the design to be something like this https://github.com/aspnet/Hosting/issues/973.

PR #1595, opened 2018-11-16
Add TestServer support for generic WebHost

Tratcher:
#1583 Create new test patterns for anyone using generic host.

### DIFF
--- a/src/Microsoft.AspNetCore.TestHost/HostBuilderTestServerExtensions.cs
+++ b/src/Microsoft.AspNetCore.TestHost/HostBuilderTestServerExtensions.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting.Server;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.AspNetCore.TestHost
+{
+    public static class HostBuilderTestServerExtensions
+    {
+        /// <summary>
+        /// Retrieves the TestServer from the host services.
+        /// </summary>
+        /// <param name="host"></param>
+        /// <returns></returns>
+        public static TestServer GetTestServer(this IHost host)
+        {
+            return (TestServer)host.Services.GetRequiredService<IServer>();
+        }
+
+        /// <summary>
+        /// Retrieves the test client from the TestServer in the host services.
+        /// </summary>
+        /// <param name="host"></param>
+        /// <returns></returns>
+        public static HttpClient GetTestClient(this IHost host)
+        {
+            return host.GetTestServer().CreateClient();
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNetCore.TestHost/HostBuilderTestServerExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public static class HostBuilderTestServerExtensions
+    {
+        /// <summary>
+        /// Builds and Starts the Host using a TestServer.
+        /// </summary>
+        /// <param name="hostBuilder"></param>
+        /// <param name="featureCollection"></param>
+        /// <returns></returns>
+        public static async Task<TestServer> StartTestServerAsync(this IHostBuilder hostBuilder, IFeatureCollection featureCollection = null)
```
davidfowl: I think we should stop making extension methods on the builder that starts. It's been a pattern that I've hardly seen anyone use and it breaks the builder flow and the host flow. How do you dispose the host in this case? It's lost...

src/Microsoft.AspNetCore.TestHost/HostBuilderTestServerExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+            await host.StartAsync();
+            return testServer;
+        }
+
+        /// <summary>
+        /// Builds and Starts the Host using a TestServer.
+        /// </summary>
+        /// <param name="hostBuilder"></param>
+        /// <returns></returns>
+        public static async Task<IHost> StartTestHostAsync(this IHostBuilder hostBuilder)
```
davidfowl: Same comment as above. I don't think we should do this automatically.

davidfowl (changes requested):
I think if we go this route we need to let the TestServer own the IHost instance just as it owns the IWebHost instance today. That also means making the IWebHost property work with a wrapper implementation.

src/Microsoft.AspNetCore.TestHost/HostBuilderTestServerExtensions.cs
jkotalik: Agreed. I think it's better to just have TestServer.StartAsync().

src/Microsoft.AspNetCore.TestHost/TestServer.cs
```diff
@@ -49,16 +70,26 @@ public IWebHost Host
         {
             get
             {
+                if (_hostInstance == null)
```
jkotalik: nit: stay consistent with getter syntax. This one checks _hostInstance == null while other uses ??.

src/Microsoft.AspNetCore.TestHost/HostBuilderTestServerExtensions.cs
Tratcher: The test code is very verbose without this kind of helper. We don't want this busywork in our tests.
```
            var server = new TestServer();
            var host = new HostBuilder()
                .ConfigureWebHost(webBuilder =>
                {
                    webBuilder
                        .UseServer(server)
                        .Configure(app => { });
                })
                .Build();
            await host.StartAsync();
             var response = await server.CreateClient().GetAsync("/");
```
compared to:
```

            var host = await new HostBuilder()
                .ConfigureWebHost(webBuilder => webBuilder.Configure(app => { }))
                .StartTestHostAsync();
             var response = await host.GetTestClient().GetAsync("/");
```

Tratcher:
Note to self:
```
            var host = await new HostBuilder()
                .ConfigureWebHost(webBuilder =>
                {
                    webBuilder
                        .UseTestServer()
                        .Configure(app => { });
                })
                .StartAsync();

             var response = await host.GetTestClient().GetAsync("/");
```

jkotalik (approved):

Tratcher:
Updated with more distinct types of sugar.

davidfowl (commented):
Much better!

Issue #1594, opened 2018-11-16
TestServer hangs when an exception is thrown on response starting for big responses


javiercn:
[buildhang.zip](https://github.com/aspnet/Hosting/files/2590510/buildhang.zip)

PR #1596, opened 2018-11-17
Make TestServer handle exceptions from OnStarting 

Tratcher:
#1594 Large exceptions would cause back-pressure and prevent the app from exiting. Abort the stream so no more writes are allowed and report the exception to the client.

### DIFF
--- a/test/Microsoft.AspNetCore.TestHost.Tests/ClientHandlerTests.cs
+++ b/test/Microsoft.AspNetCore.TestHost.Tests/ClientHandlerTests.cs
@@ -272,6 +272,63 @@ namespace Microsoft.AspNetCore.TestHost
             Assert.IsType<InvalidOperationException>(ex.GetBaseException());
         }
 
+        [Fact]
+        public Task ExceptionFromOnStartingFirstWriteIsReported()
+        {
+            var handler = new ClientHandler(PathString.Empty, new DummyApplication(context =>
+            {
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.TestHost.Tests/ClientHandlerTests.cs
```diff
@@ -272,6 +272,63 @@ public async Task ExceptionAfterFirstWriteIsReported()
...
+        [Fact]
+        public Task ExceptionFromOnStartingWithNoWriteIsReported()
+        {
+            var handler = new ClientHandler(PathString.Empty, new DummyApplication(context =>
+            {
+                context.Response.OnStarting(() =>
+                {
+                    throw new InvalidOperationException(new string('a', 1024 * 32));
+                });
+                return Task.CompletedTask;
```
jkotalik: Good amount of shared code between these methods. Consider making the at least the last 2 lines a shared methods between the 3 tests.

jkotalik (approved):
Minor nit, otherwise LGTM.

test/Microsoft.AspNetCore.TestHost.Tests/ClientHandlerTests.cs
javiercn: IMO it’s better to have tests be isolated than DRY

